PR #15, opened 2017-06-07
Update timestamp according to device timestamp

SchumyHao:
Hi @rytilahti 
I fixed time-stamp issue when I want to control some other Xiaomi devices.
As you meet the same issue after you update Vacuum FW.
May you try this patch?


rytilahti:
Thank you @SchumyHao for pointing me to the correct direction with this!

PR #31, opened 2017-07-23
Device support for the xiaomi air purifier added.

syssi:


rytilahti:
Hi Sebastian! Patches for new devices are very welcome and I think people with those devices will truly appreciate that :-) I think it is time to start considering how to deal with the issue of the separate binaries soon enough, but before that I think it makes sense not to populate the containers file with everything, but just have those in the same file as the rest of the (device-specific) implementation. What do you think?

mirobo/containers.py
```diff
@@ -39,6 +39,115 @@ def pretty_time(x: int) -> datetime:
...
+        # ['power', 'aqi', 'humidity', 'temp_dec',
+        #  'mode', 'led', 'led_b', 'buzzer', 'child_lock',
+        #  'limit_hum', 'trans_level', 'bright',
+        #  'favorite_level', 'filter1_life', 'act_det',
+        #  'f1_hour_used', 'use_time', 'motor1_speed']
+        self.data = data
+
+    @property
+    def power(self) -> str:
+        return self.data["power"]
```
rytilahti: Do you have an example what type of response is expected here? A numerical or a string?

mirobo/containers.py
```diff
@@ -39,6 +39,115 @@ def pretty_time(x: int) -> datetime:
...
+
+    @property
+    def is_on(self) -> bool:
+        return self.power == "on"
+
+    @property
+    def aqi(self) -> int:
+        return self.data["aqi"]
```
rytilahti: aqi = air quality index?

mirobo/containers.py
```diff
@@ -39,6 +39,115 @@ def pretty_time(x: int) -> datetime:
...
+        }
+        return brightness_levels[self.data["led_b"]]
+
+    @property
+    def buzzer(self) -> bool:
+        return self.data["buzzer"] == "on"
+
+    @property
+    def child_lock(self) -> str:
+        return self.data["child_lock"]
```
rytilahti: Is this a boolean or a string?

syssi:
This is fine for me. I will just move it.

mirobo/airpurifier.py
```diff
@@ -0,0 +1,63 @@
...
+        properties = ['power', 'aqi', 'humidity', 'temp_dec',
+                      'mode', 'led', 'led_b', 'buzzer', 'child_lock',
+                      'limit_hum', 'trans_level', 'bright',
+                      'favorite_level', 'filter1_life', 'act_det',
+                      'f1_hour_used', 'use_time', 'motor1_speed']
+
+        values = self.send(
+            "get_prop",
+            properties
+        )
```
rytilahti: Considering that the get_prop is pretty generic (it's also used for the strip it seems, as well as for the yeelights), maybe it makes sense to have a list (or a mapping?) of properties and their corresponding types listed and move the status() implementation to the parent class at some point? See also the comments below regarding to property names.

mirobo/airpurifier.py
```diff
@@ -0,0 +1,63 @@
...
+        return AirPurifierStatus(dict(zip(properties, values)))
+
+    def set_mode(self, mode: str):
+        """Set mode."""
+
+        # auto, silent, favorite, medium, high, strong, idle
+        return self.send("set_mode", [mode])
```
rytilahti: Enumize setting and getting of the mode?

mirobo/containers.py
```diff
@@ -39,6 +39,115 @@ def pretty_time(x: int) -> datetime:
...
 
+class AirPurifierStatus:
+    """Container for status reports from the air purifier."""
+    def __init__(self, data: Dict[str, Any]) -> None:
+        # ['power', 'aqi', 'humidity', 'temp_dec',
+        #  'mode', 'led', 'led_b', 'buzzer', 'child_lock',
+        #  'limit_hum', 'trans_level', 'bright',
+        #  'favorite_level', 'filter1_life', 'act_det',
+        #  'f1_hour_used', 'use_time', 'motor1_speed']
+        self.data = data
```
rytilahti: Do you have an example of a response? It would be nice to have included also.

mirobo/containers.py
```diff
@@ -39,6 +39,115 @@ def pretty_time(x: int) -> datetime:
...
+        return self.data["trans_level"]
+
+    @property
+    def bright(self) -> int:
+        return self.data["bright"]
+
+    @property
+    def favorite_level(self) -> str:
+        # Favorite level used when the mode is `favorite`. Between 0 and 16.
+        return self.data["favorite_level"]
```
rytilahti: Is this an integer or str?

mirobo/containers.py
```diff
@@ -39,6 +39,115 @@ def pretty_time(x: int) -> datetime:
...
+
+    @property
+    def filter_life_remaining(self) -> int:
+        return self.data["filter1_life"]
+
+    @property
+    def act_det(self) -> bool:
+        return self.data["act_det"] == "on"
```
rytilahti: Any description for this , limit_hum and trans_level? The names are not really descriptive.

mirobo/airpurifier.py
syssi: Yes, the properties are somehow generic but they cannot be applied to all devices. For example there are three versions of the air purifier out there all of them with different supported OperationModes. Some devices does respond with "on" and "off"... others with "true" and "false". I suggest to refactor the code if the support is stable and the knowledge about different responses stronger.

mirobo/airpurifier.py
```diff
@@ -0,0 +1,182 @@
...
+                      'favorite_level', 'filter1_life', 'act_det',
+                      'f1_hour_used', 'use_time', 'motor1_speed']
+
+        values = self.send(
+            "get_prop",
+            properties
+        )
+        return AirPurifierStatus(dict(zip(properties, values)))
+
+    def set_mode(self, mode: OperationMode):
```
rytilahti: It's better to move the class on top of the file, another possibility would be to use a forward declaration ('OperationMode'), which does not look so nice IMO and should be avoided where not really necessary (like when using it inside its own defining class).

mirobo/airpurifier.py
rytilahti: Ok, sounds sane! :+1:

mirobo/airpurifier.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def set_led_brightness(self, brightness: int):
+        """Set led brightness."""
+
+        # bright: 0, dim: 1, off: 2
+        return self.send("set_led_b", [brightness])
+
+    def set_led(self, led: bool):
+        """Turn led on/off."""
+
```
rytilahti: No need for extra white-spaces after the docstring, odd that hound did not yell about that...

mirobo/airpurifier.py
rytilahti: To add, do you happen to have an example also for this case? It would be nice to have example answers for all calls supported by devices.

mirobo/airpurifier.py
```diff
@@ -0,0 +1,182 @@
...
+    def buzzer(self) -> bool:
+        return self.data["buzzer"] == "on"
+
+    @property
+    def child_lock(self) -> str:
+        return self.data["child_lock"]
+
+    @property
+    def limit_hum(self) -> str:
+        return self.data["limit_hum"]
```
rytilahti: I think my comment on these got lost somewhere, can we have better, more descriptive names for this, trans_level, act_det and led_b? led_brightness probably for led_b, but I have no idea what those two others could even be..

syssi: Done. I don't know the meaning of trans_level and act_det, too. ;-)

mirobo/airpurifier.py
syssi: I must confess I don't own the device. I just want to see it supported. I will ask around. :-)

mirobo/airpurifier.py
```diff
@@ -0,0 +1,183 @@
...
+    def led_brightness(self) -> LedBrightness:
+        return LedBrightness(self.data["led_b"])
+
+    @property
+    def buzzer(self) -> bool:
+        return self.data["buzzer"] == "on"
+
+    @property
+    def child_lock(self) -> str:
+        return self.data["child_lock"]
```
rytilahti: Is this a boolean or "on" / "off"?

mirobo/airpurifier.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @property
+    def humidity_limit(self) -> str:
+        return self.data["limit_hum"]
```
rytilahti: str or float/int?

mirobo/airpurifier.py
```diff
@@ -0,0 +1,183 @@
...
+        return self.data["trans_level"]
+
+    @property
+    def bright(self) -> int:
+        return self.data["bright"]
+
+    @property
+    def favorite_level(self) -> str:
+        # Favorite level used when the mode is `favorite`. Between 0 and 16.
+        return self.data["favorite_level"]
```
rytilahti: Considering the setter is using int, this is probably integer too, right?

mirobo/airpurifier.py
```diff
@@ -0,0 +1,183 @@
...
+    def child_lock(self) -> str:
+        return self.data["child_lock"]
+
+    @property
+    def humidity_limit(self) -> str:
+        return self.data["limit_hum"]
+
+    @property
+    def trans_level(self) -> str:
+        return self.data["trans_level"]
```
rytilahti: Level indicates that it's probably an integer, but this is ok and can be tuned when the payload is known..

mirobo/airpurifier.py
rytilahti: Thanks for helping to get it supported! :-) Potential tester available here: https://community.home-assistant.io/t/xiaomi-mi-air-purifier-support/14228/12 but for that we need to have a simple test script.

syssi:
I will fix/clarify the open questions if some payload of a response is known. It's fine for me if the PR won't get merged some time.

mirobo/airpurifier.py
```diff
@@ -0,0 +1,185 @@
...
+
+class AirPurifierStatus:
+    """Container for status reports from the air purifier."""
+    def __init__(self, data: Dict[str, Any]) -> None:
+        # Response of a Air Purifier Pro:
+        # ['power': 'off', 'aqi': 41, 'humidity': 62, 'temp_dec': 293,
+        #  'mode': 'auto', 'led': 'on', 'led_b': null, 'buzzer': null,
+        #  'child_lock': 'off', 'limit_hum': null, 'trans_level': null,
+        #  'bright': 71, 'favorite_level': 17, 'filter1_life': 77,
+        #  'act_det': null, 'f1_hour_used': 771, 'use_time': 2776200,
```
rytilahti: Favorite level is 17 here, but in the comment of it says: # Favorite level used when the mode is `favorite`. Between 0 and 16. Those which return null should probably be marked in typing with eg. `Optional[bool]` (buzzer for example). Hopefully we can get some more test examples to see if those are always null..

rytilahti: Oh, to add. Those are probably encoded as Nones, trying to construct an enum out of them will cause ValueError, so trying to access `led_brightness` would cause a crash at least. I think 
```if self.data["led_b"] is not None:
```
is probably the best solution here.

rytilahti:
We are also going to need `mirobo -d info` output from this device to add support to the (upcoming) new cli tool.

syssi:
Known models: 'zhimi.airpurifier.m1', 'zhimi.airpurifier.v1', 'zhimi.airpurifier.v2', 'zhimi.airpurifier.v3', 'zhimi.airpurifier.v6', 'zhimi.humidifier.v1'.

rytilahti:
Ok, thanks! Let's get this merged, the client tool can be added / updated later on, maybe having this in encourages someone  to do that :-)

PR #32, opened 2017-07-23
Device support for the xiaomi power strip added.

syssi:


mirobo/strip.py
```diff
@@ -0,0 +1,62 @@
...
+        """Retrieve properties."""
+        properties = ['power', 'temperature', 'current', 'mode']
+        values = self.send(
+            "get_prop",
+            properties
+        )
+        return StripStatus(dict(zip(properties, values)))
+
+    def set_power_mode(self, mode: str):
+        """Set mode."""
```
rytilahti: What kind of modes are accepted?

mirobo/strip.py
```diff
@@ -0,0 +1,62 @@
...
+        """Power on."""
+        return self.send("set_power", ["on"])
+
+    def off(self):
+        """Power off."""
+        return self.send("set_power", ["off"])
+
+    def status(self):
+        """Retrieve properties."""
+        properties = ['power', 'temperature', 'current', 'mode']
```
rytilahti: Do you have an example response? It would be nice to be added either here or into the Status class.

mirobo/strip.py
syssi: "green" (aka eco) and "normal".

rytilahti: Ok. Would it be worth expose this as an enum too?

syssi: Could you provide an example? I'm confused.

mirobo/strip.py
syssi: I will care about.

mirobo/strip.py
rytilahti: For example something like this:
```
import enum
class PowerMode(enum.Enum):
    Eco = "green"
    Normal = "normal"
```
and calling `set_power_mode(PowerMode.Eco)`

syssi: Got it! This is nice. I will update the PRs asap.

rytilahti: Btw, you can then simply use mode.value to get the value on the right hand side insie `set_mode()`, and other way around when reading it, PowerMode('green') will give you back the PowerMode.Eco :-)

mirobo/strip.py
```diff
@@ -0,0 +1,69 @@
...
+        values = self.send(
+            "get_prop",
+            properties
+        )
+        return StripStatus(dict(zip(properties, values)))
+
+    def set_power_mode(self, mode: PowerMode):
```
syssi: I don't understand the complain. :-(

rytilahti: The PowerMode is being used before it has been defined (see my comment in another PR :))

mirobo/strip.py
rytilahti: Great! (I saw this was added, but the "todo" is still open.

mirobo/strip.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # green, normal
+        return self.send("set_power_mode", [mode.value])
+
+
+class StripStatus:
+    """Container for status reports from the strip."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        # {'power': 'on', 'temperature': 48.11, 'current': 0.06, 'mode': 'green'}
```
rytilahti: :+1:

rytilahti:
Looks good, let's merge it!

syssi:
Known models: 'qmi.powerstrip.v1', 'zimi.powerstrip.v2'

PR #33, opened 2017-07-23
Device support for the chuangmi plug v1 added

syssi:


rytilahti:
Do you happen to have the devtype (assuming it's usable at all..) / other identifying information we could use for discovering these new devices? The PR looks fine otherwise.

syssi:
I requested the devtype and hope for feedback: https://community.home-assistant.io/t/xiaomi-mi-wifi-plug-air-purifier/16189/99

mirobo/plug_v1.py
```diff
@@ -0,0 +1,54 @@
...
+        values = self.send(
+            "get_prop",
+            properties
+        )
+        return PlugV1Status(dict(zip(properties, values)))
+
+
+class PlugV1Status:
+    """Container for status reports from the plug."""
+    def __init__(self, data: Dict[str, Any]) -> None:
```
rytilahti: An example would also be nice here :-) Are those values already booleans or should they be converted below?

syssi: I asked for the response at the same post: https://community.home-assistant.io/t/xiaomi-mi-wifi-plug-air-purifier/16189/99

rytilahti: Oops, sorry for that, too many PRs got me confused.

rytilahti:
Btw, for the future, it's usually better to rebase than merge on PRs to keep the commit log cleaner :-)

syssi:
Got it. Thanks! I'm learning a lot.

rytilahti:
We are also going to need `mirobo -d info` output from this device to add support to the (upcoming) new cli tool.

syssi:
The output is of the correct device if it contains
```
'model': 'chuangmi.plug.v1'
```

rytilahti:
Thanks again :) I think we are going to need to update the README to have information about other supported devices & their models.

PR #39, opened 2017-08-15
Supported devices added to README.md and version bumped

syssi:


rytilahti:
Thanks! I'll have some local changes still to push for this release (including better discovery of devices), which I think should go into the new release. I will handle this at some point this week.

syssi:
All right!

rytilahti:
I just commited the revamped discovery, so I think this can be merged now too! (we are really going to need some unit tests soon enough to be sure nothing gets broken between these refactorings...). My next step would be to work on merging the cli tools.

rytilahti:
Made a new release, thanks a lot @syssi and @kuduka :-)

PR #45, opened 2017-08-29
Xiaomi Ceiling Lamp: Some refactoring and fault tolerance if a philips light ball is used

### LINKED ISSUE #43: Some element about Xiaomi Philips Bulb
Thx to add support for this.

I test all fonction present for ceiling.

Some of them doesn't work, because this option aren't present on the app and on the device it self : 

- acct_on/acct_off doesn't work
- sml_on / sml_off doesn't work

status return an error : 

```
nico@debianpcnico:~$ miceil --ip 192.168.0.208 --token 324b4147dc6c1a9867f6e19ec9070ca0 status
Power: on
Brightness: 10
CCT: 1
Scene Number: 4
dv: 0
  File "/usr/local/bin/miceil", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 722, in __call__
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 1066, in invoke
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 895, in invoke
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 535, in invoke
  File "/usr/local/lib/python3.4/dist-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 535, in invoke
  File "/usr/local/lib/python3.4/dist-packages/mirobo/ceil_cli.py", line 104, in status
    click.echo("Smart Midnight Light: %s" % res.bl)
  File "/usr/local/lib/python3.4/dist-packages/mirobo/ceil.py", line 40, in bl
    return self.data["bl"]
KeyError: 'bl'
```

Hopping this can help !to provied a "mibulb" section

syssi:
Fixes #43.

### DIFF
--- a/mirobo/ceil.py
+++ b/mirobo/ceil.py
@@ -126,5 +126,4 @@ class Ceil(Device):
                     "count (%s) of received values.",
                     properties_count, values_count)
 
-        print(values)
         return CeilStatus(dict(zip(properties, values)))
--- a/mirobo/ceil.py
+++ b/mirobo/ceil.py
@@ -48,7 +48,7 @@ class CeilStatus:
 
     def __str__(self) -> str:
         s = "<CeilStatus power=%s, brightness=%s, " \
-            "correlated_color_temperature=%s, scene=%s, dv=%s, " \
+            "color_temperature=%s, scene=%s, dv=%s, " \
             "bl=%s, ac=%, >" % \
             (self.power, self.brightness,
              self.color_temperature, self.scene, self.dv,
--- a/mirobo/ceil.py
+++ b/mirobo/ceil.py
@@ -44,16 +44,16 @@ class CeilStatus:
         return self.data["bl"]
 
     @property
-    def ac(self) -> int:
+    def automatic_color_temperature(self) -> int:
         return self.data["ac"]
 
     def __str__(self) -> str:
         s = "<CeilStatus power=%s, brightness=%s, " \
             "color_temperature=%s, scene=%s, dv=%s, " \
-            "bl=%s, ac=%, >" % \
+            "smart_night_light=%s, automatic_color_temperature=%, >" % \
             (self.power, self.brightness,
              self.color_temperature, self.scene, self.dv,
-             self.bl, self.ac)
+             self.smart_night_light, self.automatic_color_temperature)
         return s
 
 
--- a/mirobo/ceil.py
+++ b/mirobo/ceil.py
@@ -111,19 +111,19 @@ class Ceil(Device):
         )
         properties_count = len(properties)
         values_count = len(values)
-
-
         if properties_count != values_count:
             if properties_count == values_count+2:
-                _LOGGER.info("The values of two properties are missing. "
-                    "Assumption: A Xiaomi Philips Light LED Ball is "
-                    "used which doesn't provide the property bl and ac.",
+                _LOGGER.info(
+                    "The values of two properties are missing. Assumption: A "
+                    "Xiaomi Philips Light LED Ball is used which doesn't "
+                    "provide the property bl and ac.",
                     properties_count, values_count)
 
                 values.extend((None, None))
             else:
-                _LOGGER.error("Count (%s) of requested properties does not "
-                    "match the count (%s) of received values.",
+                _LOGGER.error(
+                    "Count (%s) of requested properties does not match the "
+                    "count (%s) of received values.",
                     properties_count, values_count)
 
         print(values)
--- a/mirobo/ceil.py
+++ b/mirobo/ceil.py
@@ -1,5 +1,8 @@
+import logging
 from .device import Device
 from typing import Any, Dict
+
+_LOGGER = logging.getLogger(__name__)
 
 class CeilStatus:
     """Container for status reports from Xiaomi Philips LED Ceiling Lamp"""
@@ -20,11 +23,11 @@ class CeilStatus:
         return self.power == "on"
 
     @property
-    def bright(self) -> int:
+    def brightness(self) -> int:
         return self.data["bright"]
 
     @property
-    def snm(self) -> int:
+    def scene(self) -> int:
         return self.data["snm"]
 
     @property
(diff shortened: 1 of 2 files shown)


mirobo/ceil.py
```diff
@@ -104,4 +109,22 @@ def status(self) -> CeilStatus:
...
+        values_count = len(values)
+        if properties_count != values_count:
+            if properties_count == values_count+2:
+                _LOGGER.info(
+                    "The values of two properties are missing. Assumption: A "
+                    "Xiaomi Philips Light LED Ball is used which doesn't "
+                    "provide the property bl and ac.",
+                    properties_count, values_count)
+
+                values.extend((None, None))
```
syssi: I admit inheritance of two devices (led ball and ceiling lamp) would be the better approach.

mirobo/ceil.py
```diff
@@ -20,19 +23,19 @@ def is_on(self) -> bool:
...
         return self.data["bright"]
 
     @property
-    def snm(self) -> int:
+    def scene(self) -> int:
         return self.data["snm"]
 
     @property
     def dv(self) -> int:
         return self.data["dv"]
```
rytilahti: what could dv mean? It would be nice to have a real name for it too.

mirobo/ceil.py
rytilahti: This feels way too complicated and unpythonic. How about using `defaultdict` just below
```
defaultdict(lambda: None, zip(properties, values))
```
and having a warning() or debug() here if the amount of returned values is different from what was asked?

mirobo/ceil_cli.py
```diff
@@ -97,9 +97,9 @@ def status(dev: mirobo.Ceil):
         return  # bail out
 
     click.echo(click.style("Power: %s" % res.power, bold=True))
-    click.echo("Brightness: %s" % res.bright)
-    click.echo("CCT: %s" % res.cct)
-    click.echo("Scene Number: %s" % res.snm)
+    click.echo("Brightness: %s" % res.brightness)
+    click.echo("Correlated color temperatur: %s" % res.color_temperature)
```
rytilahti: temperature :-)

mirobo/ceil_cli.py
```diff
@@ -97,9 +97,9 @@ def status(dev: mirobo.Ceil):
...
+    click.echo("Scene: %s" % res.scene)
     click.echo("dv: %s" % res.dv)
     click.echo("Smart Midnight Light: %s" % res.bl)
     click.echo("Auto CCT: %s" % res.ac)
```
rytilahti: Could you also rename `ac` and `bl`, maybe `auto(matic)_color_temperature` and `smart_midnight_mode` or similar.

mirobo/ceil_cli.py
```diff
@@ -120,19 +120,20 @@ def off(dev: mirobo.Ceil):
...
 
 @cli.command()
-@click.argument('level', callback=validate_bright, required=True,)
+@click.argument('level', callback=validate_percentage, required=True,)
 @pass_dev
-def set_cct(dev: mirobo.Ceil, level):
+def set_color_temperature(dev: mirobo.Ceil, level):
     """Set CCT level."""
-    click.echo("CCT level: %s" % dev.set_cct(level))
+    click.echo("Correlated color temperatur level: %s" %
```
rytilahti: temperature. Any ideas what "correlated" means and does it bring anything for the user what simply calling it "color temperature" would not?

mirobo/ceil.py
syssi: Unfortunately I don't know the meaning. Google didn't help as well.

mirobo/ceil_cli.py
syssi: I removed "correlated". It won't help anybody and isn't correct because a percentage is delivered.

mirobo/ceil.py
rytilahti: Fair enough, it can be corrected later on. Btw, looking at the list of properties (with the comment of limitation of 8), is any of those left out properties interesting to include considering that currently 7 properties is being fetched?

syssi: I just own the the Philips LED Ball. This device just provides the first 5 properties. Maybe @kuduka can provide some more informations.

kuduka: dv property represents seconds remaining until turn off is activated.

At least in ceiling light maximum value is 6 hours.

```
### 192.168.1.219 => 192.168.1.227 (b8:27:eb:34:b0:bb => 34:ce:00:86:db:44)
{"id":9056,"method":"delay_off","params":[21599]}
### 192.168.1.219 => 192.168.1.227 (b8:27:eb:34:b0:bb => 34:ce:00:86:db:44)
{"id":9057,"method":"get_prop","params":["power","bright","snm","dv","cct","sw","bl","mb","ac","ms"]}
### 192.168.1.227 => 192.168.1.219 (34:ce:00:86:db:44 => b8:27:eb:34:b0:bb)
{"result":["on",35,0,**21597**,30,[[0,3],[0,2],[0,1]],1,1,1,1],"id":9057}
```

rytilahti: @kuduka do you also have descriptions for the rest of the undocumented properties?

kuduka: Which ones are left undocumented? I think that was the last one on this code.

From product side there are quite a few that are not included on this code related to miband/wall switch/remote controller pairing and its functionality.

rytilahti: 'mb', 'ms', 'cctsw' and 'sw' are shown in the example query, but are not currently fetched and there are no property getters for them.

kuduka: yes, those are related to miband/wall switch integration and they are not supported yet :)

I need to spend some time deciphering them and with current limitation of 8 props I don't know how to do it, see this  --> https://github.com/rytilahti/python-mirobo/pull/35#issuecomment-322198864

Although some times using more than 8 props as shown on this previous PR works fine, so it's kind of a mystery this behavior.

Any ideas how I could query more than 8 props? Doing 2 calls and merging its results? That would be acceptable? Any other ideas out there?

PR #46, opened 2017-09-01
Device support for the Chuangmi IR Remote Controller

syssi:
The command set is based on an [unofficial home assistant component](https://github.com/syssi/chuangmi_ir/blob/master/custom_components/switch/chuangmi_ir.py) was discussed [here](https://community.home-assistant.io/t/xiaomi-chuangmi-ir-universal-ir-remote-controller/25821) and applies to this [product](https://www.gearbest.com/smart-home/pp_229556.html).


rytilahti:
Looks simple and straightforward, thanks!

PR #47, opened 2017-09-01
Device support for the Xiaomi Mi Smart Fan

syssi:


mirobo/fan.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def set_natural_level(self, level: int):
+        """Set natural level."""
+        return self.send("set_natural_level", [level])
+
+    def set_speed_level(self, level: int):
+        """Set speed level."""
+        return self.send("set_speed_level", [level])
+
+    def set_move(self, direction: MoveDirection):
```
rytilahti: set_direction maybe?

mirobo/fan.py
```diff
@@ -0,0 +1,188 @@
...
+        """Power off."""
+        return self.send("set_power", ["off"])
+
+    def set_natural_level(self, level: int):
+        """Set natural level."""
+        return self.send("set_natural_level", [level])
+
+    def set_speed_level(self, level: int):
+        """Set speed level."""
+        return self.send("set_speed_level", [level])
```
rytilahti: What are natural and speed levels?

mirobo/fan.py
```diff
@@ -0,0 +1,188 @@
...
+        return self.send("set_angle_enable", ["on"])
+
+    def oscillate_off(self):
+        """Disable oscillate."""
+        return self.send("set_angle_enable", ["off"])
+
+    def set_led_brightness(self, brightness: LedBrightness):
+        """Set led brightness."""
+
+        # bright: 0, dim: 1, off: 2
```
rytilahti: This is already encoded in the enum, so please remove the comment.

syssi:
Hmm... it looks like I'm still unable to do a proper rebase. :-(

rytilahti:
The following works for me usually, I don't do pull but use fetch and rebase instead of it like so:
```
git fetch upstream
git rebase upstream/master
```

syssi:
Hmm..  I did the same, fixed the conflict, was forced to "add" and "git rebase --continue". To push the result I was forced to do a merge commit. I'll be damned!

rytilahti:
Maybe it would make just sense to create a new branch on top of the current master, and cherry-pick the commits you want from this one there? That would keep the history clean.

PR #50, opened 2017-09-03
Power strip: Fixes calculation of the instantaneous current

syssi:
The value of the property "current" is misleading for countries with 230V mains because it was calculated with a wrong reference voltage (110V). As workaround a calculated property "load_power" is provided now. It's a firmware issue of the Xiaomi strip.


rytilahti:
I trust you on this one too, thanks!

PR #52, opened 2017-09-03
Device support for the Xiaomi Mi Smart Fan

syssi:
Retry.

### DIFF
--- a/mirobo/fan.py
+++ b/mirobo/fan.py
@@ -1,6 +1,6 @@
 import logging
 from .device import Device
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 import enum
 
 _LOGGER = logging.getLogger(__name__)
@@ -42,22 +42,20 @@ class FanStatus:
         return self.data["humidity"]
 
     @property
-    def temperature(self) -> float:
+    def temperature(self) -> Optional[float]:
         if self.data["temp_dec"] is not None:
             return self.data["temp_dec"] / 10.0
-        else:
-            return None
+        return None
 
     @property
     def led(self) -> bool:
         return self.data["led"] == "on"
 
     @property
-    def led_brightness(self) -> LedBrightness:
+    def led_brightness(self) -> Optional[LedBrightness]:
         if self.data["led_b"] is not None:
             return LedBrightness(self.data["led_b"])
-        else:
-            return None
+        return None
 
     @property
     def buzzer(self) -> bool:
--- a/mirobo/fan.py
+++ b/mirobo/fan.py
@@ -45,6 +45,8 @@ class FanStatus:
     def temperature(self) -> float:
         if self.data["temp_dec"] is not None:
             return self.data["temp_dec"] / 10.0
+        else:
+            return None
 
     @property
     def led(self) -> bool:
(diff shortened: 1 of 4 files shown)


mirobo/fan.py
```diff
@@ -0,0 +1,188 @@
...
+        s = "<FanStatus power=%s, temperature=%s, humidity=%s, led=%s, " \
+            "led_brightness=%s buzzer=%s, child_lock=%s, natural_level=%s, " \
+            "speed_level=%s, oscillate=%s, battery=%s, ac_power=%s, " \
+            "poweroff_time=%s, speed=%s, angle=%s" % \
+            (self.power, self.temperature, self.humidity, self.led,
+             self.led_brightness, self.buzzer, self.child_lock,
+             self.natural_level, self.speed_level, self.oscillate,
+             self.battery, self.ac_power, self.poweroff_time,
+             self.speed_level, self.angle)
+        return s
```
rytilahti: I'm unsure if it makes sense to expose everything in `__str__` for all devices, but that can be cleaned up for all supported devices at some point if needed.

mirobo/fan.py
```diff
@@ -0,0 +1,188 @@
...
+        return self.power == "on"
+
+    @property
+    def humidity(self) -> int:
+        return self.data["humidity"]
+
+    @property
+    def temperature(self) -> float:
+        if self.data["temp_dec"] is not None:
+            return self.data["temp_dec"] / 10.0
```
rytilahti: This requires return None in the end to avoid missing "return" (typing, flake8 etc. can be run by simply calling `tox` in the source dir btw).

mirobo/fan.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @property
+    def led(self) -> bool:
+        return self.data["led"] == "on"
+
+    @property
+    def led_brightness(self) -> LedBrightness:
+        if self.data["led_b"] is not None:
+            return LedBrightness(self.data["led_b"])
```
rytilahti: Ditto here, just add `return None` to the end.

rytilahti:
In general looks much better, thanks! Just a couple of nits to be done before merging. I'll probably have to look into how the output of mypy/flake8 can be integrated here directly to checks.

mirobo/fan.py
syssi: My IDE is unhappy now because None isn't "float". ;-)

syssi:
Could you just tell me the commands I can execute locally to check?

mirobo/fan.py
syssi: Alright.

mirobo/fan.py
rytilahti: Ah, it should be defined as `Optional[float]` then :-)

rytilahti:
> Could you just tell me the commands I can execute locally to check?

Executing `tox` on the source directory will run py.test, flake8 and mypy accordingly. Here's how to run those manually (outside of tox's env):
mypy: `mypy --ignore-missing-imports mirobo`
flake8: `flake8 mirobo`
py.test: `pytest mirobo`

mirobo/fan.py
```diff
@@ -0,0 +1,192 @@
...
+    @property
+    def humidity(self) -> int:
+        return self.data["humidity"]
+
+    @property
+    def temperature(self) -> float:
+        if self.data["temp_dec"] is not None:
+            return self.data["temp_dec"] / 10.0
+        else:
+            return None
```
rytilahti: I think it's better to skip the `else` block completely and just return False if the `if` doesn't get hit.

PR #56, opened 2017-09-04
Fixes the verify token method

### LINKED ISSUE #54: Error in new mirobo/protocol.py
Hello rytilahti,
the new protocol.py caused an error:
>  File "/mirobo/protocol.py", line 42, in verify_token
>     raise ValueError("Token must be of length 32")
> ValueError: Token must be of length 32

This is the code block:
```
    @staticmethod
    def verify_token(token: bytes):
        if not isinstance(token, bytes):
            raise TypeError("Token must be bytes")
        if len(token) != 32:
            raise ValueError("Token must be of length 32")
```
I think
> if len(token) != 32:

must be

> if len(token) != 16:

because of "bytes".

---
Short question:
Why is
    # click.echo("DND enabled: %s" % res.dnd)
    # click.echo("Map present: %s" % res.map)
    # click.echo("in_cleaning: %s" % res.in_cleaning)
comment out in vacuum_cli.py?

---

Greetings

syssi:
Closes #54. Kudos to @SelmaUrban.

### DIFF
--- a/mirobo/protocol.py
+++ b/mirobo/protocol.py
@@ -38,7 +38,7 @@ class Utils:
     def verify_token(token: bytes):
         if not isinstance(token, bytes):
             raise TypeError("Token must be bytes")
-        if len(token) != 32:
+        if len(token) != 16:
             raise ValueError("Token must be of length 32")
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


rytilahti:
Ah, didn't see this early enough. I fixed it a bit differently to make it comply with the `bytes.fromhex()` like it's done elsewhere.

PR #58, opened 2017-09-05
Fixes various lint issues

syssi:


syssi:
Are you aware about this issue:

mirobo/vacuum_cli.py:198:1: F811 redefinition of unused 'start' from line 148
mirobo/vacuum_cli.py:206:1: F811 redefinition of unused 'stop' from line 169

rytilahti:
Yes I'm aware, but I'm not sure what causes it unfortunately.. Regarding to your patch, I don't think the pep8 line length limitation should be strictly followed and I prefer actually keep method signatures in one line if possible, so I may change those at some point in the future.

For now let's get this merged and make a new release!

syssi:
The cause is L150 and L171. vac.manual_start and vac.start has the same method name.

rytilahti:
Ohh, true. Mmh, I'm unsure if it would be better just to ignore that specific error for those two cases; `mirobo manual start_manual` instead of `mirobo manual start` would not be a nice solution in my eyes. As this doesn't affect the functionality, I think we can just merge this PR now.

PR #63, opened 2017-09-06
Adjust property names of some devices 

syssi:
...as preparation for refactoring (inheritance of properties).

### DIFF
--- a/mirobo/airpurifier.py
+++ b/mirobo/airpurifier.py
@@ -158,7 +158,7 @@ class AirPurifierStatus:
         return self.data["child_lock"] == "on"
 
     @property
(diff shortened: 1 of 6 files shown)


syssi:
@kuduka Could you check something again? The unit of the delay_off_countdown (property: dv) of the ceiling lamp is in seconds and at the eyecare 2 (property: dvalue) is in minutes, right?

syssi:
I would be happy about a merge. If needed I will hands in changes later.

kuduka:
Yes, on this one it is in minutes, don't ask me why )

syssi:
Alright!

PR #64, opened 2017-09-06
Abstract device model exteded by model name (identifier)

syssi:


mirobo/device.py
```diff
@@ -2,7 +2,7 @@
 import datetime
 import socket
 import logging
-from typing import Any, List
+from typing import Any, List, Optional
```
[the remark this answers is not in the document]
syssi: It's used a annotation. This is fine.

mirobo/device.py
```diff
@@ -184,6 +190,9 @@ def raw_command(self, cmd, params):
     def info(self):
         return DeviceInfo(self.send("miIO.info", []))
 
+    def model(self):
+        return self.info().model
```
rytilahti: Although I think it's fine to expose the model information directly in `DeviceInfo`, I'm not sure if the `Device` interface should be extended to expose this too. What is the usecase you have in your mind and would it be too bad to access the model directly there on the call-site?

mirobo/device.py
```diff
@@ -25,14 +25,20 @@ def __repr__(self):
                                                  self.data["token"])
 
     @property
-    def netif(self):
+    def netif(self) -> str:
```
rytilahti: IIRC these are not strings but rather dicts, maybe it's a good idea to add an example output into the class' docstring.

syssi: My fault. I wil add a proper docstring for the future.

mirobo/device.py
syssi: I want to provide the model of the device as sensor attribute at home assistant. It's good to know as long there is no discovery process and the user has to decide which component to use. Especially if various generations of a device exists I would like to handle the firmware differences properly.

mirobo/device.py
rytilahti: Not a problem. Looks like I lost my connectivity to my raspberry pi running as a controller, so I can't do much testing on this for now. If you can, could you please add the output of the info() to the doctstring of DeviceInfo class (that will also help when creating unit tests at some point).

mirobo/device.py
rytilahti: Makes sense, but would it be too bad to fetch info only once and use the DeviceInfo to get extract the information out from it, making the fetch explicit and avoid doing multiple I/O calls when accessing other information from the object.

mirobo/device.py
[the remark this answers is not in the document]
rytilahti: Indeed, I added a `noqa` for those cases for now.

rytilahti:
Mmh, looks like coveralls got stuck for some reason. Anyway, this all looks fine so I'm merging it.

PR #65, opened 2017-09-07
Device info extended by two additional properties

syssi:
Access the raw data is a bit dangerous.

cp. https://github.com/home-assistant/home-assistant/pull/9325/files


rytilahti:
Agreed, it's not a good idea to rely on raw values in the long run!

PR #66, opened 2017-09-08
Device support of the Xiaomi Air Humidifier

syssi:


rytilahti:
Thanks, looks like there is quite a bit of shared features between the devices, maybe this can be refactored later on somehow..

PR #67, opened 2017-09-08
First draft of the water purifier support

syssi:
The exact feature set is unkown at the moment. A list of supported properties / some network traffic would be nice.


rytilahti:
We can add it, but maybe it makes sense to add a call for help to the README.md regarding to missing functionality?

syssi:
You are right. I will extend the README.md!

rytilahti:
Great!

PR #72, opened 2017-09-11
WIP: Avoid discovery flooding

syssi:
Just added the first of probably two needed changes to fix the discovery of some devices (water purifier, vacuum(?)).


rytilahti:
I added the loop in f35910e20cbe8579a462268ac6e74f27d5db29cc as in some point the robot was not always responsive. However, maybe that has been fixed since then or was just a problem with my network (hard to say), so I don't think there's much harm if we merge this for now and see if it works just fine.

I have been running it with this patch for some time now and tested that the everything seems to still function just fine.

rytilahti:
Unfortunately it looks like this breaks the connectivity with the vacuum, it was a bit too early to get it merged.. :(

PR #85, opened 2017-10-01
Extend yeelight support

rytilahti:
Adds more functionality and allows yeelights to be discovered.
* set_developer_mode(bool) allows enabling and disabling the developer mode (necessary for python-yeelight etc.)
* set_name(str) to set the internal name, note that this is not the same as shown in mobile apps
* set_save_state_on_change(bool) to enable saving the current bulb state when doing modifications

Note: initialization still causes a warning to recommend python-yeelight, which offers a much nicer API. However, for some users the ability to activate the developer mode with this can be very useful.


rytilahti:
Added and fixed conflicts, merging!

Issue #69, opened 2017-09-08
Xiaomi WiFi Speaker support


syssi:
```
-> 192.168.2.44 data= {"id":2893,"method":"get_prop","params":["umi"]}
<- 192.168.2.103 data= {"result":["{ \"DeviceName\": \"Mi Internet Speaker\", \"channel_title\": \"音箱下载\", \"current_state\": \"PLAYING\", \"hardware_version\": \"S602\", \"play_mode\": \"REPEAT_ALL\", \"track_artist\": \"Kygo&M83\", \"track_duration\": \"00:04:58\", \"track_title\": \"Wait(Remix)\", \"transport_channel\": \"PLAYLIST\" }"],"id":2893}
-> 192.168.2.44 data= {"id":2894,"method":"get_prop","params":["rel_time"]}
<- 192.168.2.103 data= {"result":["00:04:00"],"id":2894}
-> 192.168.2.44 data= {"id":2895,"method":"power","params":[]}
<- 192.168.2.103 data= {"result":"ok","id":2895}
-> 192.168.2.44 data= {"id":2896,"method":"get_prop","params":["rel_time"]}
<- 192.168.2.103 data= {"result":["00:04:02"],"id":2896}
-> 192.168.2.44 data= {"id":2897,"method":"get_prop","params":["rel_time"]}
<- 192.168.2.103 data= {"result":["00:04:03"],"id":2897}
-> 192.168.2.44 data= {"id":2898,"method":"vol_up","params":[5]}
<- 192.168.2.103 data= {"result":"ok","id":2898}
-> 192.168.2.44 data= {"id":2899,"method":"vol_down","params":[5]}
<- 192.168.2.103 data= {"result":"ok","id":2899}
-> 192.168.2.44 data= {"id":2900,"method":"previous_track","params":[]}
<- 192.168.2.103 data= {"result":"ok","id":2900}
-> 192.168.2.44 data= {"id":2901,"method":"next_track","params":[]}
<- 192.168.2.103 data= {"result":"ok","id":2901}
-> 192.168.2.44 data= {"id":2902,"method":"get_prop","params":["rel_time"]}
<- 192.168.2.103 data= {"result":["00:00:00"],"id":2902}
-> 192.168.2.44 data= {"id":2903,"method":"get_prop","params":["umi"]}
[...]

PR #86, opened 2017-10-01
Initial support for wifi speakers

rytilahti:
First steps towards fixing #69, we require feedback
from someone who owns the device to complete it.

### DIFF
--- a/mirobo/wifispeaker.py
+++ b/mirobo/wifispeaker.py
@@ -1,0 +1,79 @@
+import warnings
+import logging
+from .device import Device
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class WifiSpeakerStatus:
+    def __init__(self, data):
+        self.data = data
+
+    @property
+    def device_name(self) -> str:
+        return self.data["DeviceName"]
+
+    @property
+    def channel(self) -> str:
+        return self.data["channel_title"]
+
+    @property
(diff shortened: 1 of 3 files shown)


mirobo/wifispeaker.py
```diff
@@ -0,0 +1,79 @@
...
+    def volume_down(self, amount: int = 5):
+        return self.send("vol_down", [amount])
+
+    def track_previous(self):
+        return self.send("previous_track")
+
+    def track_next(self):
+        return self.send("next_track")
+
+    def track_location(self):
```
syssi: I would prefer track_position here.

mirobo/wifispeaker.py
```diff
@@ -0,0 +1,79 @@
+import warnings
+import logging
+from .device import Device
+
+_LOGGER = logging.getLogger(__name__)
```
syssi: The logger isn't used?

rytilahti: It is not currently, but I think it's ok to leave it like that for now. This will probably need some revisions to be remotely useful, and for that there'll be debug outputs I hope.

mirobo/wifispeaker.py
rytilahti: Changed + added some apidocs & fixed the broken `status`.

PR #94, opened 2017-10-21
Device support of the Xioami Philips Smart LED Ball Lamp

syssi:
I like to see this device supported instead of using a subset of the Philips Celing Lamp.


rytilahti:
Looks good to me!

PR #96, opened 2017-10-21
Some typing hints added and the code order aligned

syssi:


rytilahti:
Looks good, I removed the unused optimal.

PR #99, opened 2017-10-21
Add a base to allow easier testing of devices

rytilahti:
To demonstrate its functionality unittests for yeelight and plug
are included in this commit. It also allowed to spot a couple of bugs
in yeelight already..


syssi:
Cool! Could be merged. I will try to adopt the test for a air purifier soon!

PR #101, opened 2017-10-24
use upstream android_backup

rytilahti:
Considering bluec0re has merged the required functionality back to his repository, we should use it instead of having a local copy.


syssi:
Straightforward! :+1:

PR #105, opened 2017-10-28
Make vacuum robot wifi settings configurable via CLI

infinitydev:
These changes allow you to configure the wifi of the Xiaomi vacuum cleaner without needing the Mi Home smartphone app.

It is now effectively possible to use python-miio with the robot in your home wifi network without a Xiaomi account and their app.

Reference:
https://github.com/OpenMiHome/mihome-binary-protocol/blob/master/doc/PROTOCOL.md#payloads

The document mentions that the 'id' field must be a Unix timestamp. This seems to be nonsense. Works perfectly fine with the sequential id. I assumed that it might be used to set the clock of the rebot, but it is not. (It uses NTP.)

I am not sure what the 'uid' field is. Probably the Xiaomi account id that allows them to connect your robot to your account. This should be pretty useless if you don't set the robot up with their app. Hence, I made it an optional parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,8 @@ wifi network, especially if you block its internet access.
 The token does not change by configuring wifi. All subsequent commands must be sent to
 the IP that the robot obtains via DHCP from your router.
 
+*Note, that this only works on devices from which you have obtained a token beforehand and therefore rules out e.g. vacuums with firmware 3.3.9_003077 or higher.* Any ideas for extracting the token on newer devices without using the official app and extracting them from the database are appreciated!
+
 ### Setting the fanspeed
 
 ```
--- a/README.md
+++ b/README.md
@@ -222,6 +222,9 @@ OK
 It may take a few minutes before the robot responds to commands after it connected to the
 wifi network, especially if you block its internet access.
 
+The token does not change by configuring wifi. All subsequent commands must be sent to
+the IP that the robot obtains via DHCP from your router.
+
 ### Setting the fanspeed
 
 ```
--- a/README.md
+++ b/README.md
@@ -211,6 +211,17 @@ $ mirobo home
 Requesting return to home: 0
 ```
 
+### Configure wifi settings (WPA2)
+
+```
+$ mirobo configure_wifi "My SSID" "My Wifi password"
+Configuring wifi to SSID: My SSID
+OK
+```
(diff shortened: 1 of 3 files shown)


marcelrv:
Did your vacuum indeed function proper.
I have users that indicated in the latest fw you need time info

miIO.config_router{“tz”:“Europe/Brussels”,“gmt_offset”:0,“ssid”:“WInternal”,“passwd”:“xxx”}

Also, would be good to understand if you can disconnect the Xiaomi cloud (meaning the vacuum no longer needs internet)

infinitydev:
Yes, my vaccum works properly. I never got it to work with the original app. Hence, my robot is running on stock firmware. So, I cannot test on recent firmware unless there is a way to upload a new one without the app.

I have completely blocked the robot's internet access in my router to prevent it from talking home. I think that is why it takes a couple of minutes before it responds to commands on the configured wifi network.

After connecting my laptop to the robots default wifi, changing wifi settings through this patch and waiting a couple of minutes, I can control the robot with miio and an alternative smartphone app within my LAN.

rytilahti:
They probably added the timezone setting for the `config_router` at the same time they introduced the ability to change it with `set_timezone` (as before that all the schedules run on time in China.

What remains unclear to me what will be used a a token when `config_router` is used, could you elaborate (and add that also to README)?

@infinitydev could you fix those issues as reported by hound?

infinitydev:
The token does not change when configuring wifi. I added a paragraph to the README for clarification.

The workflow for configuring my stock robot boiled down to:

1. Connect to robot's default wifi network
2. Obtain token<br>`mirobo discover --handshake 1`
3. Configure wifi<br>`mirobo --ip=192.168.8.1 --token=<my_token> configure_wifi "SSID" "password"`
4. Wait some minutes
5. Check connection<br>`mirobo --ip=<LAN_IP_of_robot> --token=<my_token>`

rytilahti:
Ok, thanks! I read at some point that the newer firmware releases will disable obtaining the token with a handshake, leaving this method usable only for earlier versions. I adjusted the README to inform potential users about this caveat.

rytilahti:
I'll merge this now, and add the ability to define a timezone in another PR. Thanks for the PR! 👍

PR #107, opened 2017-10-29
Add the ability to define a timezone for configure_wifi

rytilahti:
This will allow passing the timezone to vacuums running
newer firmware versions, as mentioned in PR #105

Thanks to @marcelrv  for pointing this out.


syssi:
Good work! Please merge.

PR #113, opened 2017-11-06
Model name of the Xiaomi Philips Ceiling Lamp updated

syssi:


rytilahti:
Considering that these were not detectable over mdns, I think this is not so important. But let's merge it for the sake of consistence, thanks!

PR #115, opened 2017-11-06
Reset consumable by name

mrin:
Available consumable names:
- main_brush_work_time
- side_brush_work_time
- sensor_dirty_time
- filter_work_time

CLI example:
`miio reset_consumable sensor_dirty_time`

Tested with my robot.

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -21,11 +21,13 @@ class TimerState(enum.Enum):
     On = "on"
     Off = "off"
 
+
 class Consumable(enum.Enum):
     MainBrush = "main_brush_work_time"
     SideBrush = "side_brush_work_time"
     Filter = "filter_work_time"
     SensorDirty = "sensor_dirty_time"
+
 
 class Vacuum(Device):
     """Main class representing the vacuum."""
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -110,10 +110,10 @@ class Vacuum(Device):
         """Return information about consumables."""
         return ConsumableStatus(self.send("get_consumable")[0])
 
-    def consumable_reset(self):
+    def consumable_reset(self, name):
         """Reset consumable information."""
-        raise NotImplementedError("unknown parameters")
-        # self.send("reset_consumable", ["unknown"])
+        # name = ["main_brush_work_time", "side_brush_work_time", "sensor_dirty_time" or "filter_work_time"]
+        return self.send("reset_consumable", [name])
 
     def map(self):
         """Return map token."""
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -21,6 +21,12 @@ class TimerState(enum.Enum):
     On = "on"
     Off = "off"
 
+class Consumable(enum.Enum):
+    MainBrush = "main_brush_work_time"
+    SideBrush = "side_brush_work_time"
+    Filter = "filter_work_time"
+    SensorDirty = "sensor_dirty_time"
+
 class Vacuum(Device):
     """Main class representing the vacuum."""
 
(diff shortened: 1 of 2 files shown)


miio/vacuum_cli.py
```diff
@@ -145,6 +145,15 @@ def consumables(vac: miio.Vacuum):
 
 
 @cli.command()
+@click.argument('name', type=str, required=True)
+@pass_dev
+def reset_consumable(vac: miio.Vacuum, name):
+    """Query and reset consumable."""
+    click.echo("Reset consumable %s" % name)
```
rytilahti: "Reseting consumable '%s'" would look here nicer.

miio/vacuum.py
```diff
@@ -110,10 +110,10 @@ def consumable_status(self) -> ConsumableStatus:
         """Return information about consumables."""
         return ConsumableStatus(self.send("get_consumable")[0])
 
-    def consumable_reset(self):
+    def consumable_reset(self, name):
         """Reset consumable information."""
-        raise NotImplementedError("unknown parameters")
-        # self.send("reset_consumable", ["unknown"])
+        # name = ["main_brush_work_time", "side_brush_work_time", "sensor_dirty_time" or "filter_work_time"]
+        return self.send("reset_consumable", [name])
```
rytilahti: Do you happen to know if multiple consumables can be reseted at once?

miio/vacuum.py
```diff
@@ -110,10 +110,10 @@ def consumable_status(self) -> ConsumableStatus:
         """Return information about consumables."""
         return ConsumableStatus(self.send("get_consumable")[0])
 
-    def consumable_reset(self):
+    def consumable_reset(self, name):
         """Reset consumable information."""
-        raise NotImplementedError("unknown parameters")
-        # self.send("reset_consumable", ["unknown"])
+        # name = ["main_brush_work_time", "side_brush_work_time", "sensor_dirty_time" or "filter_work_time"]
```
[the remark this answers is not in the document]
rytilahti: Could you create a `Consumable` enum, and use it instead of passing strings around. See https://github.com/rytilahti/python-miio/blob/master/miio/vacuum.py#L20 (and its use in https://github.com/rytilahti/python-miio/blob/master/miio/vacuum.py#L164) for example.

We do not need to follow their naming scheme either, I think it makes sense here to go for the same naming as in `ConsumableStatus` (https://github.com/rytilahti/python-miio/blob/master/miio/vacuumcontainers.py#L284).

miio/vacuum.py
mrin: I not tried this case on my robo.

miio/vacuum_cli.py
```diff
@@ -145,6 +145,29 @@ def consumables(vac: miio.Vacuum):
...
+    if name == 'main_brush':
+        consumable = Consumable.MainBrush
+    elif name == 'side_brush':
+        consumable = Consumable.SideBrush
+    elif name == 'filter':
+        consumable = Consumable.Filter
+    elif name == 'sensor_dirty':
+        consumable = Consumable.SensorDirty
+    else:
+        return
```
rytilahti: I think it makes sense to error out here before returning, if not just to avoid bug reports caused by potential misspellings / name changes.

miio/vacuum_cli.py
```diff
@@ -145,6 +145,29 @@ def consumables(vac: miio.Vacuum):
 
 
 @cli.command()
+@click.argument('name', type=str, required=True)
+@pass_dev
+def reset_consumable(vac: miio.Vacuum, name):
+    """Reset consumable."""
```
rytilahti: Could you also add a list information about the available values? Adding an empty line after the "title line" will make it only visible when help on the specific command is requested, IIRC, so I suppose something like this would work:
```
"""Reset consumable state.

Allowed values:
main_brush - explanation
another one - x
```

rytilahti:
Thanks a lot! I adjusted the help text a bit.

PR #117, opened 2017-11-08
Load power of the PowerStrip fixed and removed from the Plug

syssi:
- Property load_power removed from the plug. It's a feature of the PowerStrip. The Xiaomi plug doesn't provide load. 
- Load power workaround of the PowerStrip replaced by a property called "power_consume_rate" which provides the same values as the android app.


rytilahti:
Yay for coverage increase too, thanks :-)

PR #121, opened 2017-11-18
Powerstrip fixes

rytilahti:
This allows powerstrip to work when it has no cloud connectivity + adds a safeguard against unnavailable mode information.


syssi:
Looks good!

PR #122, opened 2017-11-18
Replace PyCrypto with PyCryptoDome

pluehne:
First of all, thanks for this great package!

I’m using Arch Linux, and after some tinkering, I now got `python-miio` to work properly 😃.

One issue that I stumbled upon is that Arch Linux doesn’t provide PyCrypto packages anymore. The background is that PyCrypto is [unsafe (it has security vulnerabilities and is no longer actively maintained)](https://blog.sqreen.io/stop-using-pycrypto-use-pycryptodome/), [outdated (the last release was in 2013)](https://github.com/dlitz/pycrypto/releases), and now superseded by [PyCryptoDome](https://www.pycryptodome.org/en/latest/src/introduction.html).

For this reason, I believe it would be best to replace PyCrypto with PyCryptoDome, which is what this pull request does.

### DIFF
--- a/miio/extract_tokens.py
+++ b/miio/extract_tokens.py
@@ -2,7 +2,7 @@ import logging
 import click
 import tempfile
 import sqlite3
-from Crypto.Cipher import AES
+from Cryptodome.Cipher import AES
 from pprint import pformat as pf
 import attr
 from android_backup import AndroidBackup
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ cryptography
 pretty_cron
 construct
 zeroconf
-pycrypto #  for miio-extract-tokens
+pycryptodomex #  for miio-extract-tokens
 attrs
 typing  # for py3.4 support
 pytz  # for tz offset in vacuum
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
                       'pretty_cron',
                       'typing',
                       'zeroconf',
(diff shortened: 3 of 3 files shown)


rytilahti:
Thanks for the patch! I looked a bit into it, and I believe it makes more sense to avoid depending on two crypto libraries (pycrypto is just used by extract-tokens, the protocol implementation uses cryptography) and will push a that change in in a sec.

pluehne:
@rytilahti: That’s a very good proposition. Thanks for caring about this 😃!

rytilahti:
Not a thing, thanks for your PRs and analyze on the topic :-)

pluehne:
Admittedly, I much prefer handling all packages through the operating system’s very own package manager *only* instead of dealing with additional systems such as `pip`, `gem`, and `npm`. This has the advantage that packages exist in one version only and the neither the user space is cluttered with extra files nor the system space, which should only be maintained by the operating system’s package manager to my mind.

However, I believe that there are reasons for using virtualenvs. This is, perhaps, a matter of philosophy that we need not elaborate here 😃.

rytilahti:
That's a fair point, and besides of that, it also helps when packagers do quality control of some sort to improve the packaging per se. For some usecases it is however necessary to do version pinning, and those cases it's worth also to have a package which is installable outside the global scope. 

Besides, it is easier for regular users to install applications from the official repository of their chosen operating system, so thanks for working to get this shippable :-)

PR #123, opened 2017-11-18
Don’t require typing with Python 3.5 and newer

pluehne:
Python 3.5 and newer ship with [built-in typing support](https://lwn.net/Articles/640359/). For this reason, there is no need to require the `typing` package starting from Python 3.5.

On Arch Linux, this requirement caused trouble, because Arch Linux doesn’t provide the `typing` package anymore, as it is deemed unnecessary now. For this reason, `python-miio` complains about the missing `typing` dependency without this adjustment.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
                       'click',
                       'cryptography',
                       'pretty_cron',
-                      'typing',
+                      'typing; python_version < "3.5"',
                       'zeroconf',
                       'pycrypto',
                       'attrs',


rytilahti:
Looks good and reasonable, adding that requirement so hastily (and keeping support for py < 3.5) was not properly motivated choice. I think it would actually be fair enough to drop support for py < 3.5 considering Debian stable is also already on it, but let's leave it like that for the time being and merge this. Thanks!

PR #124, opened 2017-11-18
Add GPLv3 license

pluehne:
I just packaged [`python-miio` for Arch Linux in the Arch User Repository](https://aur.archlinux.org/packages/python-miio/).

In doing so, I noticed that, while the GLPv3 license is [stated in `setup.py`](https://github.com/rytilahti/python-miio/blob/5f4e8b139aa302375a6e870ca201644bb92d91c2/setup.py#L15), there is no explicit license file. I know that dealing with licenses is somewhat annoying, but proper packages should install license files to `/usr/share/licenses/`. Additionally, this is useful information for users, and the GLPv3 license will be displayed nicely by GitHub for clarity.

I took the license template from [GLPv3 in Markdown](https://github.com/TheFox/GPLv3.md) and filled in the necessary information in the second commit, 18fdce146e51f53c7abdde8fb9697d80d98c583e, to make reviewing the actual changes more easy.

Feel free to make adjustments as necessary 😃.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,637 @@
+# GNU GENERAL PUBLIC LICENSE
+Version 3, 29 June 2007
+
+Copyright (C) 2007 [Free Software Foundation, Inc.](http://fsf.org/)
+
+Everyone is permitted to copy and distribute verbatim copies of this license
+document, but changing it is not allowed.
+
+## Preamble
+
+The GNU General Public License is a free, copyleft license for software and
+other kinds of works.
+
+The licenses for most software and other practical works are designed to take
+away your freedom to share and change the works. By contrast, the GNU General
+Public License is intended to guarantee your freedom to share and change all
+versions of a program--to make sure it remains free software for all its users.
+We, the Free Software Foundation, use the GNU General Public License for most
+of our software; it applies also to any other work released this way by its
+authors. You can apply it to your programs, too.
+
+When we speak of free software, we are referring to freedom, not price. Our
+General Public Licenses are designed to make sure that you have the freedom to
+distribute copies of free software (and charge for them if you wish), that you
+receive source code or can get it if you want it, that you can change the
+software or use pieces of it in new free programs, and that you know you can do
+these things.
+
+To protect your rights, we need to prevent others from denying you these rights
+or asking you to surrender the rights. Therefore, you have certain
+responsibilities if you distribute copies of the software, or if you modify it:
+responsibilities to respect the freedom of others.
+
+For example, if you distribute copies of such a program, whether gratis or for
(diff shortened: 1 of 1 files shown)


rytilahti:
Well spotted, and it is indeed a good idea to have an explicit LICENSE file delivered with the package (needs adjusting also to setup.py? I'm unsure..). However, I think it makes more sense to have a verbatim copy of original, text-based (non-markdownized) version of the license to make it plain enough. What do you think?

pluehne:
That’s up to you, after all this project is mostly your effort 😃.

I usually prefer Markdown license files, because they are easier on the eye should you really want to read them. And even if nobody looks at the license file, why not make it look nicer?

pluehne:
Aside from my earlier comment, I think that naming it `LICENSE.md` is fine. My understanding is that most systems will look for any variations like `LICENSE`, `COPYING`, `LICENSE.txt`, `LICENSE.md`, etc.

And I don’t know what the convention for handling licenses in `setup.py` is, as I’m haven’t build Python install scripts so far.

rytilahti:
https://stackoverflow.com/questions/9977889/how-to-include-license-file-in-setup-py-script makes a remark that using a metadata for pointing out the license is fine, so I don't think it's necessary to ship the license file with it. Let's ship the LICENSE.md for those who want to have it available, the verbatim copy of the license is available anyway e.g. in https://www.gnu.org/licenses/gpl-3.0.en.html . Thanks again!

rytilahti:
I added also a verbatim copy of the license in text as provided by github's "add license file" feature. It doesn't seem to show the license on the front page as of yet, maybe it's something that gets added at some point after some sort of analysis procedure is run on the repo..

pluehne:
@rytilahti: I’m curious about the new release 😃. Is there anything left to do for 0.3.2 that I could help with?

rytilahti:
@pluehne I think the only thing to do for a new release would be updating the README to contain a link to the new documentation (and updating the documentation where necessary!). If you want to contribute, a read-through & updates to the docs would be more than welcome :-)

The preliminary docs are now available in https://python-miio.readthedocs.io/en/latest/ - a local copy can be compiled just by calling `make` inside docs directory. My plan was to have it in such a state that it could be e.g. linked inside homeassistant's documentation instead of duplicating the information (see https://home-assistant.io/components/vacuum.xiaomi_miio/#retrieving-the-access-token).

pluehne:
@rytilahti: Cool 🙂! Unfortunately, I’m very busy until the weekend, so maybe, I’ll have a look at the docs then. Would that be fine?

rytilahti:
That'd we awesome, thanks :-)

PR #125, opened 2017-11-18
add pure text LICENSE

rytilahti:


pluehne:
@rytilahti: It looks like you forgot to fill in the author, project information, and year (see the [placeholders `<year>` etc.](https://github.com/rytilahti/python-miio/blob/5f569cf6b350bd75ccf3a26656242e9a472187e0/LICENSE#L634-L655)). You can simply copy that information from the Markdown version if you like.

rytilahti:
I deliberately left it out, as different files may have different authors :-)

pluehne:
Good point. Not sure why these placeholders exist in the first place 😃.

PR #127, opened 2017-11-22
Fix manual_control error message typo

skorokithakis:
I love this project.


skorokithakis:
I have also removed the equals from the checks, as they would print an error message if velocity was 180, say.

rytilahti:
Looks good, thanks!

PR #128, opened 2017-11-22
Add ability to read IP and token from config file

skorokithakis:
I don't like env vars, so I added a simple config file using ConfigParser. The script will prefer the vars (or command line options) as usual, but it will read the config file as a last resort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,14 @@ export MIROBO_IP=192.168.1.2
 export MIROBO_TOKEN=476e6b70343055483230644c53707a12
 ```
 
+Otherwise, you can create a config file at `~/.config/mirobo.cfg` with the ip and token as contents:
+
+```
+[mirobo]
+ip = 192.168.1.2
+token = 476e6b70343055483230644c53707a12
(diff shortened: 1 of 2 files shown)


skorokithakis:
Ah, right. Feel free to close, I just quickly hacked it together because I don't like env vars and figured I'd issue a PR. No problem!

skorokithakis:
By the way, you might want to add a link to the docs somewhere, currently they're not linked anywhere.

syssi:
Thanks for your appreciation!

rytilahti:
I added a link to README.rst and removed README.md, so now there's a link! @skorokithakis if you want, we can merge the config parser part for now and get rid of it later on, when I get some time to look into making it properly.

skorokithakis:
Up to you man, you're the maintainer. The code can be removed easily, but changing the configuration file is harder. If you think you'll keep the same config file in the future, I would merge, otherwise I would either hold off or try to get the config format right at this time so I didn't have to tell everyone to change later on.

Issue #76, opened 2017-09-12
Xiaomi Mi Home Air Conditioner Companion support


syssi:
https://github.com/mac-zhou/homeassistant-mi-acpartner

First step: Implement the AC partner device. 
Second step: Provide a PR to migrate from python-miio to python-mirobo.

mouth4war:
Is this for the Aqara AC companion?

syssi:
No. It's a full featured air condition.

cxlwill:
Hi, @syssi I'm one of contributors of this component. It keep updating and works very well in HA 0.58.1 and together with present `miio` version now. I have finished HA docs and ready for PR to HA main stream. But I think it is reasonable to change the platform even core to `xiao_miio` for synchronicity. Could I ask for your help here?

cxlwill:
Also latest firmware for ac_companion has already opened LAN functions on it. Now this device get gateway port and function. We can add this info to docs as well?

syssi:
You are right. Please rename the component. I will support your contribution and review your  component as soon as possible. We should implement a miio device class to reduce the complexity of the HA component.

cxlwill:
I have PR it. See [here](https://github.com/home-assistant/home-assistant/pull/10761). Thanks for your help~

syssi:
@cxlwill Could you provide a link to the product (f.e. at gearbest)?

cxlwill:
Here is [Aqara](https://www.gearbest.com/living-appliances/pp_416094.html) one and [Xiaomi](https://www.gearbest.com/power-strips/pp_640724.html) one.

syssi:
> Is this for the Aqara AC companion?

@mouth4war I must revoke my answer. The component is for the Xiaomi AC Companion. :-)

cxlwill:
Yes, this one is AC companion. Xiaomi did produce the air conditioner under Smartmi Brand. See [here](https://www.gearbest.com/air-purifier/pp_773532.html?vip=3250116&gclid=Cj0KCQiAgNrQBRC0ARIsAE-m-1xWhqqmZdwB0mVwq7MqfQIDaOtoWpaaW9ssGNLQ9Y4K8WxPAz6AZiAaAptkEALw_wcB). But, I think it is not worth to buy it outboard.

syssi:
Do you like to provide write permission for https://github.com/cxlwill/home-assistant? I think it's easier as talking about the needed changes. :-)

syssi:
[...]

PR #129, opened 2017-11-23
Xiaomi Air Conditioner Companion support

syssi:


syssi:
@rytilahti Do you like to take a look into the unit test? The response format of the AC companion is unusual, so the device dummy implementation isn't sufficient.

rytilahti:
There's a WIP PR on homeassistant for this, which contains additional features if I'm not mistaken: https://github.com/home-assistant/home-assistant/pull/10761 . @syssi do you think it'd make sense to merge this now, and do incremental changes assuming the authors of that PR are interested doing that?

miio/airconditioningcompanion.py
```diff
@@ -0,0 +1,99 @@
...
+    """Main class representing Xiaomi Air Conditioning Companion."""
+
+    def status(self) -> AirConditioningCompanionStatus:
+        """Return device status."""
+        status = self.send("get_model_and_state", [])
+        return AirConditioningCompanionStatus(status)
+
+    def learn(self):
+        """Learn an infrared command."""
+        return self.send("start_ir_learn", [30])
```
rytilahti: The magic 30 should be extracted into a constant and/or documented some other way.

miio/airconditioningcompanion.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class AirConditioningCompanionStatus:
+    """Container for status reports of the Xiaomi AC Companion."""
+
+    def __init__(self, data):
+        # Device model: lumi.acpartner.v2
+        #
+        # Response of "get_model_and_state":
+        # ['010500978022222102', '010201190280222221', '2']
```
rytilahti: It would be nice to have some sort of description for specific byte indexes (and include a short, informal description of the protocol if possible) already here for readability.

rytilahti:
@syssi considering that the protocol is that much different, it may make sense to avoid using the dummydevice and make the tests specific for this device? Although if it works like it is currently, that's fine too :-)

syssi:
@rytilahti I'm aware of https://github.com/home-assistant/home-assistant/pull/10761 and we are in contact already cp. https://github.com/rytilahti/python-miio/issues/76

syssi:
@rytilahti Could you check the sendcmd method of https://github.com/home-assistant/home-assistant/pull/10761? There is a proper state needed to assemble a full featured command. Do you think the method can be migrated to python-miio? Where should be maintained the state?

syssi:
I would be happy about an early merge.

rytilahti:
Yes, the part definitely should be merged (after cleaning it up!). I'm unsure what type of state maintaining is required here, if it's just the state to do a new request, I think that detail can be hidden / a cached state message can be used? For now I think we can merge this.

PR #131, opened 2017-11-23
Fix Yeelight RGB parsing

Sduniii:


rytilahti:
Thanks for the patch!

PR #132, opened 2017-11-26
Additional properties of the Xiaomi Air Purifier 2 introduced

syssi:
Some comments improved.
Some code clean-up.

### DIFF
--- a/miio/airpurifier.py
+++ b/miio/airpurifier.py
@@ -1,5 +1,6 @@
 import logging
 import enum
+import math
 from typing import Any, Dict, Optional
 from collections import defaultdict
 from .device import Device
@@ -36,14 +37,13 @@ class AirPurifierStatus:
 
         Response of a Air Purifier 2:
 
-        ['power': 'off', 'aqi': 141, 'humidity': 64, 'temp_dec': 236,
-         'mode': 'auto', 'led': 'on', 'led_b': 1, 'buzzer': 'on',
-         'child_lock': 'off', 'limit_hum': null, 'trans_level': null,
-         'bright': null, 'favorite_level': 10, 'filter1_life': 80,
-         'act_det': null, 'f1_hour_used': 680 ]
+        ['power': 'on', 'aqi': 22, 'humidity': 61, 'temp_dec': 197,
+         'mode': 'auto', 'led': 2, 'led_b': 'off', 'buzzer': 'off',
+         'child_lock': 'off', 'bright': None, 'favorite_level': 10,
+         'filter1_life': 80, 'f1_hour_used': 678, 'use_time': 2440800,
+         'motor1_speed': 351, 'purify_volume': 25127]
 
-        use_time and motor1_speed is missing because a request is limitted
-        to 16 properties. We request 15 properties at the moment.
+        A request is limitted to 16 properties.
         """
 
         self.data = data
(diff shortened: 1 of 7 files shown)


miio/airpurifier.py
```diff
@@ -153,16 +185,22 @@ class AirPurifier(Device):
...
+        properties = ['power', 'aqi', 'average_aqi', 'humidity', 'temp_dec',
+                      'mode', 'favorite_level', 'filter1_life', 'f1_hour_used',
+                      'use_time', 'motor1_speed', 'purify_volume', 'f1_hour',
+                      # Second request
+                      'led', 'led_b', 'bright', 'buzzer', 'child_lock', ]
 
         values = self.send(
             "get_prop",
-            properties
+            properties[0:13]
```
rytilahti: Mind leaving a comment behind to explain why using only a slice here? Same below too.

miio/airpurifier.py
```diff
@@ -90,6 +95,11 @@ def led_brightness(self) -> Optional[LedBrightness]:
         """Brightness of the LED."""
         if self.data["led_b"] is not None:
             return LedBrightness(self.data["led_b"])
+
+        # This is the property name of the Air Purifier Pro
+        if self.data["bright"] is not None:
+            return LedBrightness(self.data["bright"])
```
rytilahti: Does this mean that when the device is not air purifier pro, the "led_b" will be empty? If yes, then it's fine.

miio/tests/test_airpurifier.py
```diff
@@ -0,0 +1,186 @@
...
+
+        self.device.set_mode(OperationMode.Idle)
+        assert mode() == OperationMode.Idle
+
+    def test_set_favorite_level(self):
+        def favorite_level():
+            return self.device.status().favorite_level
+
+        self.device.set_favorite_level(1)
+        assert favorite_level() == 1
```
rytilahti: Is it known what values are allowed for favorite level? If yes, it makes probably sense to add an exception if out-of-range values are inputed, and to add a test for that also here (https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions).

miio/airpurifier.py
syssi: Yep. The air purifier provides a "bright" property XOR a "led_b" property.

miio/tests/test_airpurifier.py
```diff
@@ -179,8 +179,3 @@ def test_status_without_temperature(self):
         self.device.state["temp_dec"] = None
 
         assert self.state().temperature is None
-
-    def test_status_string(self):
-        self.device._reset_state()
-
-        assert self.state().__str__
```
rytilahti: The status string should be `__repr__` instead of `__str__`, right? https://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python

miio/airpurifier.py
```diff
@@ -36,14 +40,14 @@ def __init__(self, data: Dict[str, Any]) -> None:
...
+        {'power': 'on, 'aqi': 10, 'average_aqi': 8, 'humidity': 62,
+         'temp_dec': 186, 'mode': 'auto', 'favorite_level': 10,
+        'filter1_life': 80, 'f1_hour_used': 682, 'use_time': 2457000,
+        'motor1_speed': 354, 'purify_volume': 25262, 'f1_hour': 3500,
+        'led': 'off', 'led_b': 2, 'bright': None, 'buzzer': 'off',
+        'child_lock': 'off'}
 
-        use_time and motor1_speed is missing because a request is limitted
-        to 16 properties. We request 15 properties at the moment.
+        A request is limitted to 16 properties.
```
rytilahti: limited

PR #140, opened 2017-12-11
Verification token is invalid

roiff:
fix construct==2.8.17 Verification token is invalid

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     packages=["miio", "mirobo"],
 
     python_requires='>=3.4',
-    install_requires=['construct',
+    install_requires=['construct==2.8.16',
                       'click',
                       'cryptography',
                       'pretty_cron',


rytilahti:
Thanks for finding out where the problem lies, however, I'm not for merging this patch as pinning requirements is not the way to do it properly. If this breakage is caused by our invalid use of construct, we have to adapt the implementation. If the problem is caused by a bug in the newest version of construct, it should be fixed there. I'll leave this open until we find out what is the proper way to fix it!

rytilahti:
For me it works with the vacuum still with 2.8.17, but fails on the plug. I bisected the commit breaking it to https://github.com/construct/construct/commit/3c6714d22a55656e265a8d4cd43e6aaeab4c7641 . Now it will require to find out where is the best place to fix it.

roiff:
@rytilahti
It doesn't matter, I just provide a way to the fastest possible repair, because time is limited, can't further wrong, just as a reminder, after all, a lot of people already can't use, thank you for your effort

PR #141, opened 2017-12-12
update fan speed list for HK

roiff:
In the HA-HB plug, the impact speed_list operating experience, this plug-in, the generation of HA to homekit FAN apparatus, the slider homekit, The slider automatically corresponds to speed_list ,speed_list order of 0% is 'off' or 'idle' and 33% slient, 66% is the favorite , 100% is auto,
Modification here is the least amount of change and impact, can be more perfect
https://github.com/home-assistant/homebridge-homeassistant/blob/master/accessories/fan.js

### DIFF
--- a/miio/airpurifier.py
+++ b/miio/airpurifier.py
@@ -18,7 +18,6 @@ class OperationMode(enum.Enum):
     Auto = 'auto'
 
 
-
 class LedBrightness(enum.Enum):
     Bright = 0
     Dim = 1
--- a/miio/airpurifier.py
+++ b/miio/airpurifier.py
@@ -12,10 +12,11 @@ class AirPurifierException(Exception):
 
 
 class OperationMode(enum.Enum):
-    Auto = 'auto'
+    Idle = 'idle'
     Silent = 'silent'
     Favorite = 'favorite'
-    Idle = 'idle'
+    Auto = 'auto'
+
 
 
 class LedBrightness(enum.Enum):


rytilahti:
Can anyone confirm that this is expected change? @syssi homeassistant component requires an update too, I suppose?

rytilahti:
@yawor do you have insight on this?

yawor:
Well, the Favorite mode is configurable, so it's hard to pinpoint it's position in relation to other speeds, as it can be configured to be anything from really slow to really fast.
My experience with Air Purifier Pro is that the Auto mode is even slower than Night mode, at least in my case - maybe it's a matter of low concentration of PM2.5 in my apartment, but certainly is not the fastest (like 100% in the proposed change). My setup uses totally different approach. I have the purifier running always on Favorite mode and my automation script (running in Appdaemon) controls the favorite_level and turns it on or off depending on the current AQI.

If I somehow misunderstood the reason behind this change, please let me know. I have a hard time fully understanding the author's description of this PR.

roiff:
@yawor 
Because it involves multiple projects,
The official version of the current HA-HB plugin is so logical that the logic of the HA-HB plugin needs to be modified to implement your functionality (in fact, I customized the plugin myself, but I still did not really want to submit it To the official, because this device is a more personalized example), modify a sequence here, the program has no effect

yawor:
Changing the order of that Enum shouldn't affect anything than the order of modes in UI (for example in HA component). I don't know how many projects are using this library and how are they using it so I don't know if something depends on that order.
I'm looking at the HA-HB code but it is very generic and it only differentiates between component types (fan, switch etc) without taking specific platforms into account. It makes it hard to adapt the logic of HA-HB to specific situations like this.
I think that the issue is with the XiaomiAirPurifier component in the HA, which treats operation modes as speeds, which in my personal opinion is not a good choice, because (as I've explained in my previous comment) the speeds of both Auto and Favorite modes are going to vary a lot and depending on their order to control level of speed is really bad idea. It may be correct for one user and totally incorrect for another one. The speed selector in the XiaomiAirPurifier component should (as before, that's my personal opinion) directly control the favorite_level property, maybe with automatically switching to Favorite mode on speed change if not selected before (this could be a configuration option in the entity). Maybe the whole fan support in HA could be enhanced by adding a MODE capability in addition to SET_SPEED, OSCILLATE and DIRECTION ones and both mode and speed could then be presented to the user.

yawor:
BTW seems that the Air Purifier Pro doesn't even support Idle mode.
```python
from miio.airpurifier import AirPurifier, OperationMode
a = AirPurifier(devip, devtoken)
a.set_mode(OperationMode.Idle)
```
returns
```python
{'error': {'code': -5001, 'message': 'invaild_arg'}, 'id': 3}
```

roiff:
@yawor 
Thank you for your comments, your question, I have to consider before, I am here to submit a simple change, because the smallest changes, if you want to do a good experience, you may still need to HA plug-in or HA- HB plug-in to start, if you modify these plug-ins, the problem is HA or HA-HB are not a separate air purifier class, the fan operation, and HA or HA-HB native logic has a small conflict, I do not know After the submission will be how to deal with, I will next improve my own plug-in, try to submit to the HA.
Glad to talk to you

syssi:
The Home Assistent component doesn't use the list of Operation Modes anymore.

PR #142, opened 2017-12-13
Unbreak the communication when using construct v2.8.17

rytilahti:
* Removes usage of Enum (construct internals were changed, and this is the simplest way to fix it)
* devtype and devserial are now just "device_id", as the values of those fields are not really known

Fixes checksum issue seen in https://github.com/home-assistant/home-assistant/issues/11048 - related to https://github.com/construct/construct/issues/416 .

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -112,8 +112,7 @@ class Device:
         self._timeout = 5
         self._device_ts = None  # type: datetime.datetime
         self.__id = start_id
-        self._devtype = None
-        self._serial = None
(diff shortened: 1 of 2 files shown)


fanthos:
I had the same problem when test my new device with Home Assistant.
Simply change `"devtype"` in `protocol.py` from Enum to Default fix the issue.
Combining devtype and serial should be a good idea.

rytilahti:
Ok, thanks for testing, let's get this merged!

PR #143, opened 2017-12-14
Update for Rock Robot (Mi Robot gen 2)

fanthos:
Add version string
Tested on my device and it should work as excepted.

ps. The new version using string `roborock` instead of `rockrobo`.


rytilahti:
Thanks :+1:

PR #153, opened 2018-01-03
Implement firmware update functionality

rytilahti:
Add ability to perform firmware updates on the device.
Works either by passing an URL with md5sum for the firmware image,
or alternatively by passing a local filename.
In case of the input is a local filename, its md5sum will be calculated automatically,
and the update will be delivered by the built-in web server.

```
mirobo update_firmware http://10.10.10.150/v11_003094.pkg <md5sum>
```
or
```
mirobo update_firmware v11_003094.pkg
```

The `install_sound` functionality is also extended to use the builtin web-server class, so installing sound packages is as simple as calling `mirobo install_sound <soundfile.pkg>`.

For more details & archive of firmware updates see https://github.com/dgiese/dustcloud .

Thanks for exploring the device in-depth goes to the dustcloud project,
which also describes how to build own, custom firmwares.

### DIFF
--- a/miio/updater.py
+++ b/miio/updater.py
@@ -1,0 +1,66 @@
+from http.server import HTTPServer, BaseHTTPRequestHandler
+import hashlib
+import logging
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class UpdateServer(BaseHTTPRequestHandler):
+    """A simplified handler just returning the contents of a single file."""
+    def __init__(self, request, client_address, server):
+        self.payload = server.payload
+        self.server = server
+
+        super().__init__(request, client_address, server)
+
+    def handle_one_request(self):
+        self.server.got_request = True
+        self.raw_requestline = self.rfile.readline()
+
+        if not self.parse_request():
+            print("unable to parse request")
+            return
+
+        print("got %s for %s" % (self.command, self.path))
+
+        self.send_response(200)
+        self.send_header('Content-type', 'application/octet-stream')
+        self.end_headers()
+        self.wfile.write(self.payload)
+
+
+class Updater:
+    """A simple HTTP server for serving an update file.
+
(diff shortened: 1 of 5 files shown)


miio/updater.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self, file, interface=None):
+        addr = ('', 0)
+        self.server = HTTPServer(addr, UpdateServer)
+        self.server.got_request = False
+        _, self.port = self.server.server_address
+        _LOGGER.info("Serving on port %s" % self.port)
+        self.server.timeout = 10
+
+        with open(file, 'rb') as f:
+            self.payload = f.read()
```
CODeRUS: Should be memory inefficient

rytilahti: I think it doesn't matter here, as the file needs to be read at some point for sending it to the client. The code is however incomplete, there is currently no way to tell interface/IP to use for the url (e.g. when one has multiple network interfaces as I do). I will have to look more into it in a week when I'll be able to test it on the device.

CODeRUS: The point i mean is you should read file by chunks and feed data to hash, reading hundreds megabytes to memory at once is a bad idea.

rytilahti: I understand what you mean, but at some the file has to be read (again) for sending over the wire to the device. I'll look into that though, thanks :-)

CODeRUS:
btw, why own server? why handle_request so bad?

rytilahti:
@CODeRUS I wanted to have it so that one can simply point it to any file (and not to expose the contents over HTTP), although I think changing to the source directory and launching the server there would be fine too.

rytilahti:
I think this is pretty much done and ready to be tested now. The code could be made cleaner, but that can be done later.

PR #154, opened 2018-01-05
Add preliminary support for managing sound files

rytilahti:
* Adds install_sound which expects url, md5sum and id
* Adds printing of current sound volume to sound output.


rytilahti:
Requires testing & updating the docs accordingly, an open question is what should be used as a "id" (the current one can be queried with 'sound') for newly installed ones. Another open question is how to switch between languages without doing a "full" installation.

rytilahti:
I think this can be merged now, I just tested out generating an audio pack and then installing it afterwards. To simplify the process, the server functionality worked in #153 will be used by this functionality later on.

PR #157, opened 2018-01-07
Air Purifier Pro: support for sound volume level and fix for bright propery

yawor:
The Air Purifier Pro has a property controlling the volume level of sound notifications, which are used when any command is sent to it. I've added support for the property and a method to set it's value. The range is from 0 (mute) to 100 (full volume). I don't know if Air Purifier 2 also has this property or not.

There's a photosensitive element on top of the Air Purifier Pro (right next to the power/mode selection button). It's value is held in the bright property, which has been improperly used as a LED brightness for this model. Air Purifier Pro display probably doesn't support changing brightness. The value for the bright property changes from 0 (no light detected) up to 200. I don't know what's the unit it is given in. I've got 200 when I've used my phone's LED flashlight directly on the sensor.

### DIFF
--- a/miio/airpurifier.py
+++ b/miio/airpurifier.py
@@ -100,11 +100,12 @@ class AirPurifierStatus:
         if self.data["led_b"] is not None:
             return LedBrightness(self.data["led_b"])
 
-        # This is the property name of the Air Purifier Pro
-        if self.data["bright"] is not None:
-            return LedBrightness(self.data["bright"])
-
         return None
+
+    @property
+    def brightness(self) -> Optional[int]:
+        """Environment brightness level"""
+        return self.data["bright"]
 
     @property
     def buzzer(self) -> bool:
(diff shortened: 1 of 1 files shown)


yawor:
OK, I've tested the bright value against another illuminance sensor and the value seems to be in lux. The value is not equal to the other sensor, but there's a constant difference of about 3 lx - probably because of where the sensor is located and gets slightly less light than the other one.

yawor:
I've changed the name of the brightness property to illuminance. It's more appropriate regarding it's function. For example in the Home Assistant, illuminance is a type of sensor providing information about the amount of light.

rytilahti:
This looks fine for me to be merged, however I cannot test it myself. @syssi what's your opinion on this?

syssi:
Looks good! :+1:

yawor:
@rytilahti can you point me to where's the best place to add that information in the code?

rytilahti:
@yawor I'd say directly into the docstring of the method, simply by adding "reported in lux, values between [0-200]" or similar would do fine. The point is just to make it somewhat visible for anyone stumbling upon that code / trying to use it in their own code.

rytilahti:
Looks great, thanks!

PR #160, opened 2018-01-19
Modified to support zoned cleaning mode of Roborock S50.

hastarin:
I was having trouble getting my new vacuum working with this and home-assistant.

The above changes have allowed me to get up and running at least.  If I notice any other issues I'll do my best to fix them and submit pull requests.


rytilahti:
Thanks! (let's ignore that noise from houndci and merge)

PR #161, opened 2018-01-19
Added Air Purifier 2S to supported devices

harnash:
I'm not 100% familiar with this library but I found that it can't properly discover my Air Purifier 2S. I hope this will help.


rytilahti:
Great, thanks for the report!

PR #169, opened 2018-01-21
Extended Air Conditioning Companion support

syssi:


miio/airconditioningcompanion.py
```diff
@@ -109,3 +160,40 @@ def send_command(self, command: str):
...
+                (POWER_OFF in DEVICE_COMMAND_TEMPLATES[model]):
+            return self.send_command(
+                model + DEVICE_COMMAND_TEMPLATES[model][POWER_OFF])
+
+        if model in DEVICE_COMMAND_TEMPLATES:
+            configuration = model + DEVICE_COMMAND_TEMPLATES[model]['base']
+        else:
+            configuration = model + DEVICE_COMMAND_TEMPLATES['fallback']['base']
+
+        configuration = configuration.replace('po', power.value)
```
rytilahti: Just a suggestion, would it be better to replace placeholders with something like `[PO]` to make it more clear that they are placeholders? Just a thought though, otherwise this seems fine to me so please feel to merge when you want.

syssi: Good point!

PR #173, opened 2018-01-22
New properties of the Xiaomi Air Humidifier added

syssi:


miio/airhumidifier.py
```diff
@@ -7,6 +7,10 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+class AirHumidifierException(Exception):
```
rytilahti: This should derive from `DeviceException` (looks like we have some other exceptions too, which do not do this currently) to allow easy catching of all exception from this library no matter which device is being used.

miio/airhumidifier.py
```diff
@@ -126,3 +161,27 @@ def set_buzzer(self, buzzer: bool):
...
+        if humidity not in [40, 50, 60, 70, 80]:
+            raise AirHumidifierException(
+                "Invalid target humidity: %s" % humidity)
+
+        return self.send("set_limit_hum", [humidity])
+
+    def set_favorite_level(self, level: int):
+        """Set favorite level."""
+        if level < 30 or level > 85:
+            raise AirHumidifierException("Invalid favorite level: %s" % level)
```
rytilahti: Is there a link between this level and target humidity (which allows only specific values)?

miio/tests/test_airhumidifier.py
```diff
@@ -1,6 +1,6 @@
 from unittest import TestCase
 from miio import AirHumidifier
-from miio.airhumidifier import OperationMode, LedBrightness
+from miio.airhumidifier import OperationMode, LedBrightness, AirHumidifierException
```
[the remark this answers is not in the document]
rytilahti: I think houndci should be configured to allow lines up to 100 lines, even when <80 is mostly preferred. This is not a biggie (esp. as it is in the tests), but can be fixed at some point.

miio/tests/test_airhumidifier.py
```diff
@@ -112,3 +121,56 @@ def test_status_without_led_brightness(self):
...
+
+    def test_set_child_lock(self):
+        def child_lock():
+            return self.device.status().child_lock
+
+        self.device.set_child_lock(True)
+        assert child_lock() is True
+
+        self.device.set_child_lock(False)
+        assert child_lock() is False
```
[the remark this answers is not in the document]
rytilahti: It's a good idea to have newlines at the end (for grepping etc.): https://stackoverflow.com/a/729795

miio/airhumidifier.py
syssi: No. It's more like a "fan speed".

miio/tests/test_airhumidifier.py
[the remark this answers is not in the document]
syssi: I swear there is a newline. I don't know what's going on here.

miio/airhumidifier.py
syssi: I'm waiting for some details currently: https://community.home-assistant.io/t/xiaomi-humidifier-support/35672/26

syssi:
My last commit doesn't work. Right?

miio/airhumidifier.py
rytilahti: Ok, feel free to merge when you are ready :+1:

rytilahti:
Hmm, I'm not sure if it's working or not, but nevertheless it's a good idea to have those there. Btw, if it's not too much effort, could you cherrypick 000d06e and 02e1e53 off from this PR and just commit them separately? Then those are visible in commit logs making it easier to remember to add a note about them into the changelog.

Otherwise feel free to commit when you think it's ready 👍

rytilahti:
Btw, instead of just reverting you can do a rebase, and simply drop that commit :-)

PR #174, opened 2018-01-23
Tests of the Air Purifier improved

syssi:
Cover the features of the Air Purifier 2 and Air Purifier Pro.
Tests for the new properties (volume, brightness) added.


rytilahti:
Looks good! :+1:

PR #178, opened 2018-01-24
[WIP] zone cleaning

thelittlefireman:
Initial support of zone cleaning for gen2.
ref : [https://github.com/marcelrv/XiaomiRobotVacuumProtocol/blob/master/app_zoned_clean.md](https://github.com/marcelrv/XiaomiRobotVacuumProtocol/blob/master/app_zoned_clean.md)

TODO : 
- Test in real
- Find a way to get parameters easly
- Test on gen 1 ? (will it work ?)
- Get the spécific : "stop zone cleaning function"

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -58,6 +58,10 @@ class Vacuum(Device):
         """Stop cleaning and return home."""
         self.send("app_stop")
         return self.send("app_charge")
+		
+	def zone_cleaning_start(self, params):
+        """Start specific zone cleanning of the map."""
+        return self.send("app_zoned_clean", params)
 
     def manual_start(self):
         """Start manual control mode."""
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -58,8 +58,8 @@ class Vacuum(Device):
         """Stop cleaning and return home."""
         self.send("app_stop")
         return self.send("app_charge")
-		
-	def zone_clean_start(self, params):
+
+    def zone_clean_start(self, params):
         """Start specific zone cleaning of the map."""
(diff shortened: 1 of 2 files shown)


rytilahti:
Hi and thanks for working on this! It will not work on gen1 (without a modified firmware, that is), and I'm not sure how one can get those coordinates (I assume they are extracted from a map which has to be created beforehand).

Do you have a gen2 vacuum by yourself? It'd be nice to test it a bit before merging (as you also stated in your todo).

thelittlefireman:
Hi,
My question for gen 1 was just if Xiaomi use the same firmware for both version: does they plan to implement the function in gen 1 and just add it to gen 2 for marketing benefits?

Yes I've got a gen 2 and a gen 1:) of course I planed to do some wireshark investigation and test by the end of the week I hope.

Sorry for my poor English =s

itelli:
@thelittlefireman Did you investigate further for the zone cleaning? 

Can i provide any help?

thelittlefireman:
Thanks,
I've no time for now to investigate (i hope later) and it seems to need further improvement to work correctly :
[https://github.com/marcelrv/XiaomiRobotVacuumProtocol/issues/1](https://github.com/marcelrv/XiaomiRobotVacuumProtocol/issues/1)

moreover, the map not have always the same orientation and so it cannot be possible to send the same coordinates same twice or more.

I hope xiaomi plan to add periodic zone cleaning in his app (like actual) periodic cleaning)

olskar:
some users report that v1 now has support for zone cleaning @rytilahti :)

https://imgur.com/a/0Cqjz

thelittlefireman:
#310

PR #184, opened 2018-01-27
Tests for the Chuangmi IR controller

syssi:


miio/chuangmi_ir.py
```diff
@@ -13,7 +13,7 @@ class ChuangmiIrException(DeviceException):
 class ChuangmiIr(Device):
     """Main class representing Chuangmi IR Remote Controller."""
 
-    def learn(self, key: int):
+    def learn(self, key: int = 1):
```
rytilahti: You have here whitespaces around `=` while none for frequency in play, looks fine otherwise so ship it!

syssi:
@rytilahti Could you check the travis log for an advice? I'm blind and out of ideas.

miio/tests/test_chuangmi_ir.py
```diff
@@ -0,0 +1,62 @@
...
+        self.state = {}
+        self.return_values = {
+            'miIO.ir_learn': lambda x: True,
+            'miIO.ir_read': lambda x: True,
+            'miIO.ir_play': lambda x: self._ir_play_input_validation(x),
+        }
+        super().__init__(args, kwargs)
+
+    @staticmethod
+    def _ir_play_input_validation(self, payload):
```
rytilahti: I think this definition is wrong for staticmethod, if you remove `self` it should start working :-)

PR #188, opened 2018-01-28
Features of mixed air purifier models added

syssi:


miio/airpurifier.py
```diff
@@ -209,6 +214,27 @@ def learn_mode(self) -> bool:
...
+    def sleep_mode_learn_count(self) -> Optional[int]:
+        return self.data["sleep_data_num"]
+
+    @property
+    def extra_features(self) -> Optional[int]:
+        return self.data["app_extra"]
+
+    @property
+    def turbo_mode_supported(self) -> Optional[bool]:
+        return self.data["app_extra"] == 1
```
rytilahti: This will return a bool for any value of `app_extra`, so no need to mark it Optional here.

miio/airpurifier.py
```diff
@@ -209,6 +214,27 @@ def learn_mode(self) -> bool:
...
+
+    @property
+    def auto_detect(self) -> Optional[bool]:
+        """Return True if auto detect is enabled."""
+        return self.data["act_det"] == "on"
```
rytilahti: Ditto here.

miio/airpurifier.py
```diff
@@ -110,6 +110,11 @@ def mode(self) -> OperationMode:
         """Current operation mode."""
         return OperationMode(self.data["mode"])
 
+    @property
+    def sleep_mode(self) -> Optional[OperationMode]:
+        """Operation mode of the sleep state. (Idle vs. Silent)"""
+        return OperationMode(self.data["sleep_mode"])
```
rytilahti: Here's another optional, I'm unsure what happens if the sleep_mode is None, will it just raise an exception or will it return an OperationMode (which one?)?

syssi: It will return "None" now.

PR #192, opened 2018-01-28
Fix a breaking change of construct 2.9.23

syssi:
Parameter "context" of the parse method was replaced by **kw.


yawor:
Maybe we really should pin the Construct to specific version in `setup.py`, and monitor for new versions from time to time. That way the released versions of the library won't break because of some backward incompatible change in the dependency.

Also you could just change `ctx` to `**ctx` :). But this is also fine.

PR #198, opened 2018-01-30
Make discovery to work with no IP addr and token, courtesy of M0ses

rytilahti:
Obsoletes #194

### DIFF
--- a/miio/click_common.py
+++ b/miio/click_common.py
@@ -17,7 +17,8 @@ _LOGGER = logging.getLogger(__name__)
 
 
 def validate_ip(ctx, param, value):
-    if value is None: return None
+    if value is None:
+        return None
     try:
         ipaddress.ip_address(value)
         return value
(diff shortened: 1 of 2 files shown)


M0ses:
LGTM, but maybe adding my test cases in an adapted version would make sense and could avoid regressions

rytilahti:
Good point! I renamed the file (hopefully there'll be more tests for that part of the code) and changed test to assert with 'is' instead of '=='. Thanks for the patch!

M0ses:
Thanks for fixing.

Just some nitpicking: Wouldn`t it make sense to change https://github.com/rytilahti/python-miio/pull/198/files#diff-57c62bdfa1f30df2fde08081703b47c5R5 to

```
assert validate_token(None, None, None) is None
```

as done below in https://github.com/rytilahti/python-miio/pull/198/files#diff-57c62bdfa1f30df2fde08081703b47c5R9 to be more consistent in coding?

BTW: I built rpm packages for openSUSE https://build.opensuse.org/project/show/home:M0ses:miio for those who want to try it on tumbleweed

rytilahti:
Yeah, I think it would have made sense to compare both to is None, considering that "not" doesn't compare to Noneness but to trueness/falseness. That can probably be fixed at some point when new tests get added to that suite, I think it's not worth to change it for now.

Thanks for the packaging! If you want, the README could be updated to point to available packaging, I think archlinux has also a package available :-)

PR #202, opened 2018-02-03
Preparation of release 0.3.5

syssi:


rytilahti:
If you want to write the release notes and make a release, I can push it to pypi. The only thing besides yours and @yawor's awesome work that I wanted to mention is that the exceptions coming from the device are now wrapped in `DeviceException` for easier handling.

PR #214, opened 2018-02-12
Chuangmi IR: Fix Construct 2.9.28 regresion

### LINKED ISSUE #212: Construct 2.9.28 breaks the Chuangmi IR packet assembly

yawor:
Fixes #212 
Tested against Construct versions 2.9.27, 2.9.28 and 2.9.29.

### DIFF
--- a/miio/chuangmi_ir.py
+++ b/miio/chuangmi_ir.py
@@ -141,7 +141,7 @@ ChuangmiIrSignal = Struct(
     Const(0xa567, Int16ul),
     'edge_count' / Rebuild(Int16ul, len_(this.edge_pairs) * 2 - 1),
     'times_index' / Array(16, Int32ul),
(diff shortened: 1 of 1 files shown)


arekbulski:
This wasnt Construct regression, it was providing a float instead of integer. It was working fine by sheer luck.

arekbulski:
See #222 .

yawor:
Sorry, the wording could be better. It was not my intention to say that it was a regression within the Construct, but rather in my code against newer version of it.

arekbulski:
Thats fine, thanks for explaining. The other PR #220 that fixes adapter path parameter needs some rework, but thats a different matter.

syssi:
I will care about the path parameter.

arekbulski:
#220 has three commits, if you could just undo the 2nd commit, that would work.

PR #215, opened 2018-02-12
Lazy discovery on demand

syssi:
I want to reduce the number of discovery messages of specific devices (Philips Lights).

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -99,7 +99,7 @@ class Device:
     This class should not be initialized directly but a device-specific class inheriting
     it should be used instead of it."""
     def __init__(self, ip: str = None, token: str = None,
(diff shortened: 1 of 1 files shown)


yawor:
@syssi in the async code I've started working on (not much time recently to move it forward) I'm running discovery independently from the requests. It runs on timer every minute or so. There's also an async lock which keeps the command from running if library is waiting for discovery response.
I can push my initial async code to a PR but there's not much there yet.

syssi:
@rytilahti The lazy device discovery works fine for philips lights without drawbacks so far. Are you happy with the order of the init parameters?

rytilahti:
That looks fine for me (the "force discovery" was merely a hack to make it work after some FW update), but let me run this for a day or so on all my devices just to be sure.

PR #220, opened 2018-02-13
Fixes for the API change of construct v2.9.30

syssi:
Pinning (construct==2.9.30) must be the consequence for the present. As long as construct hasn't a stable api no pinning will be brinkmanship. :rage:

Fixes #217.

PR #223, opened 2018-02-14
Use the path parameter of encode/decode again to be future-proof

syssi:
cp. https://github.com/rytilahti/python-miio/pull/220#issuecomment-365515104

### DIFF
--- a/miio/chuangmi_ir.py
+++ b/miio/chuangmi_ir.py
@@ -130,10 +130,10 @@ class ChuangmiIr(Device):
 
 
 class ProntoPulseAdapter(Adapter):
-    def _decode(self, obj, context, *args, **kwargs):
+    def _decode(self, obj, context, path):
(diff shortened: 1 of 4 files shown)


rytilahti:
I suppose the question to ask here is whether we are more afraid of potential silent errors (which we can handle by checking against incoming kwargs) or changes in the interface used by construct. Maybe keeping the path as it is, and adding `**kwargs` in case the interface gets extended at some point, to avoid crashing it all immediately?

yawor:
There's still an open issue #224 when using Construct > 2.9.28.

arekbulski:
Instead of crashing "immediately" due to error explictly related to Adapter code, it will fail "later" due to unknown reasons that dont flag construct lines of code in anyway. You might as well wrap you entire code in try/catch and ship it.

PR #228, opened 2018-02-17
Proper handling of the device_id representation

syssi:


miio/device.py
```diff
@@ -243,7 +243,7 @@ def send(self, command: str, parameters: Any=None, retry_count=3) -> Any:
             data, addr = s.recvfrom(1024)
             m = Message.parse(data, token=self.token)
             self._device_ts = m.header.value.ts
-            self._device_id = m.header.value.device_id
+            self._device_id = binascii.hexlify(m.header.value.device_id)
```
syssi: @rytilahti Does the device_id change over time? I hope not. I assume this line could be removed.

rytilahti:
Wrt. your question about changes in device id over time, I suppose it should not be doing that. The only reason I added `Hex` to that value was to see it pretty-printed out without any extra changes elsewhere in the code, but I suppose that field could simply be left to be bytes and then convert it when printing it out.

PR #236, opened 2018-02-25
Try to fix decrypted payload quirks if it fails to parse as json

jschmer:
It sometimes happens that Xiaomi cloud sends an extra byte after \x00. Need to strip everything after the last \x00 to produce valid json data.

This is an example of decrypted data that Xiaomi cloud sent to my Gen1 vacuum:
`{"id":123456789,"result":["https://awsbj0.fds.api.xiaomi.com/robomap/roboroommap...","https://awsbj0.fds.api.xiaomi.com/robomap/roboroommap...","https://awsbj0.fds.api.xiaomi.com/robomap/roboroommap...","https://awsbj0.fds.api.xiaomi.com/robomap/roboroommap..."]}\x00x
`

The extra last byte will prevent miio from loading this data as this is invalid json.

This will fix dgiese/dustcloud#77

### DIFF
--- a/miio/protocol.py
+++ b/miio/protocol.py
@@ -156,17 +156,32 @@ class EncryptionAdapter(Adapter):
             _LOGGER.debug("Unable to decrypt, returning raw bytes: %s", obj)
             return obj
 
-        decoded = decrypted.decode('utf-8')
-        try:
-            return json.loads(decoded)
-        except:
+        # list of adaption functions for malformed json payload (quirks)
+        decrypted_quirks = [
+            # try without modifications first
+            lambda decrypted_bytes: decrypted_bytes,
+            # powerstrip returns malformed JSON if the device is not
+            # connected to the cloud, so we try to fix it here carefully.
+            lambda decrypted_bytes: decrypted_bytes.replace(b',,"otu_stat"', b',"otu_stat"'),
+            # xiaomi cloud returns malformed json when answering _sync.batch_gen_room_up_url
+            # command so try to sanitize it
+            lambda decrypted_bytes:
+                decrypted_bytes[:decrypted_bytes.rfind(b'\x00')]
+                if b'\x00' in decrypted_bytes
+                else decrypted_bytes
+        ]
+
+        for i, quirk in enumerate(decrypted_quirks):
+            decoded = quirk(decrypted).decode('utf-8')
             try:
-                # powerstrip returns invalid JSON if the device is not
-                # connected to the cloud, so we try to fix it here carefully.
-                decoded = decoded.replace(',,"otu_stat"', ',"otu_stat"')
                 return json.loads(decoded)
             except Exception as ex:
-                _LOGGER.error("unable to parse json '%s': %s", decoded, ex)
+                # log the error when decrypted bytes couldn't be loaded
+                # after trying all quirk adaptions
+                if i == len(decrypted_quirks) - 1:
+                    _LOGGER.error("unable to parse json '%s': %s", decoded, ex)
+
+        return None
 
 
 Message = Struct(
--- a/miio/tests/test_protocol.py
+++ b/miio/tests/test_protocol.py
@@ -1,5 +1,7 @@
 from unittest import TestCase
 from .. import Utils
+from ..protocol import Message
+import binascii
 
 
 class TestProtocol(TestCase):
@@ -17,7 +19,7 @@ class TestProtocol(TestCase):
 
         encrypted = Utils.encrypt(payload, token)
         decrypted = Utils.decrypt(encrypted, token)
-        self.assertEquals(payload, decrypted)
+        assert payload == decrypted
 
     def test_invalid_token(self):
         payload = b"hello world"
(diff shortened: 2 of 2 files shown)


miio/protocol.py
```diff
@@ -151,7 +151,7 @@ def _decode(self, obj, context, path):
         try:
             # pp(context)
             decrypted = Utils.decrypt(obj, context['_']['token'])
-            decrypted = decrypted.rstrip(b"\x00")
+            decrypted = decrypted[:decrypted.rfind(b"\x00")]
```
syssi: Your solution isn't accurate. rfind returns -1 if there is no \x00 at all. decrypted will be shortened by one character in this case. The previous code didn't touch the payload.

```python
>>> decrypted = "asdfg"
>>> decrypted.rstrip("f")
'asdfg'
>>> decrypted[:decrypted.rfind("f")]
'asd'
>>> decrypted[:decrypted.rfind("h")]
'asdf'
```

jschmer: Correct. There is already a fixup for a different invalid json string scenario in that method so I'll extend that to add this special case.

syssi: You didn't get my point. Your code introduces a new case. If 0x00 isn't present the json will be destroyed now because the last char (`}`) will be removed.

jschmer: Sorry, I wasn't clear enough in my comment. I'll revert this change and instead provide a similar way to fix the decrypted data like is done for the powerstrip edge case.

miio/protocol.py
```diff
@@ -156,17 +156,31 @@ def _decode(self, obj, context, path):
...
+            lambda decrypted_bytes: decrypted_bytes.replace(b',,"otu_stat"', b',"otu_stat"'),
+            # xiaomi cloud returns invalid json when answering _sync.batch_gen_room_up_url
+            # command so try to sanitize it
+            lambda decrypted_bytes:
+                decrypted_bytes[:decrypted_bytes.rfind(b'\x00')]
+                if b'\x00' in decrypted_bytes
+                else decrypted_bytes
+        ]
+
+        for i, fix in enumerate(decrypted_fixes):
```
jschmer: @syssi This loop tries to fix the decrypted data as long as it can't be loaded as json. This makes it easy to add additional fixes for new edge cases and it won't meddle with the data when it can be loaded the first time.

miio/tests/test_protocol.py
```diff
@@ -32,3 +34,54 @@ def test_invalid_token(self):
...
+            # header
+            magic = binascii.unhexlify(b'2131')
+            length = (32 + len(encrypted_data)).to_bytes(2, byteorder='big')
+            unknown = binascii.unhexlify(b'00000000')
+            did = binascii.unhexlify(b'01234567')
+            epoch = binascii.unhexlify(b'00000000')
+
+            checksum = Utils.md5(magic+length+unknown+did+epoch+token+encrypted_data)
+
+            return magic+length+unknown+did+epoch+checksum+encrypted_data
```
jschmer: Had to write my own message building function because Message.build() will always produce valid json but we want to explicitly test if Message.parse() can handle invalid json input.

miio/tests/test_protocol.py
```diff
@@ -17,7 +19,7 @@ def test_encrypt(self):
 
         encrypted = Utils.encrypt(payload, token)
         decrypted = Utils.decrypt(encrypted, token)
-        self.assertEquals(payload, decrypted)
+        self.assertEqual(payload, decrypted)
```
rytilahti: As we use pytest, this could be simply `assert payload == decrypted` I suppose.

miio/tests/test_protocol.py
rytilahti: It's great to have tests for such behaviours, so thanks for this!

miio/tests/test_protocol.py
```diff
@@ -32,3 +34,54 @@ def test_invalid_token(self):
...
+
+        # can parse message with valid json
+        serialized_msg = build_msg(b'{"id": 123456}')
+
+        parsed_msg = Message.parse(serialized_msg, **ctx)
+        self.assertIsNotNone(parsed_msg)
+        self.assertIsNotNone(parsed_msg.data)
+        self.assertIsNotNone(parsed_msg.data.value)
+        self.assertIsInstance(parsed_msg.data.value, dict)
+        self.assertEqual(parsed_msg.data.value['id'], 123456)
```
rytilahti: See the above comment about pytest & assert, also https://docs.pytest.org/en/latest/assert.html .

miio/protocol.py
rytilahti: Looks great! I think such pieces are called quirks sometimes, but this will do fine :+1: I have not yet tested the code itself, but will do it later this week.

dgiese:
Great fix ;) Was wondering where the weird errors were originating from.

rytilahti:
Tested and works fine with vacuum & strip, so let's merge this. Thanks again! :+1:

PR #245, opened 2018-03-04
Some misleading docstrings updated

syssi:


rytilahti:
Uff, updated the commit message to include philips eyecare and was managed to typo it... Anyway, I think it's clear what is meant so thanks :-)

Issue #231, opened 2018-02-22
Support Xiaomi Air Purifier v3


syssi:
**zhimi.airpurifier.v3**

{'power': 'off', 'aqi': 0, 'humidity': None, 'temp_dec': None, 'mode': 'idle', 'led': 'off', 'led_b': 10, 'buzzer': 'on', 'child_lock': 'off', 'bright': 43, 'favorite_level': None, 'filter1_life': 26, 'f1_hour_used': 2573, 'use_time': None, 'motor1_speed': 0, 

{‘power’: 'on', ‘aqi’: 18, ‘humidity’: None, ‘temp_dec’: None, ‘mode’: 'silent', ‘led’: 'off', ‘led_b’: 10, ‘buzzer’: 'on', ‘child_lock’: 'off', ‘bright’: 4, ‘favorite_level’: None, ‘filter1_life’: 26, ‘f1_hour_used’: 2574, ‘use_time’: None, ‘motor1_speed’: 648}

set_mode returns ok for: auto, silent, favorite, medium, high, strong, idle

cp. https://community.home-assistant.io/t/xiaomi-mi-wifi-plug-air-purifier/16189/244

syssi:
```
if state.led_brightness:
File “/srv/homeassistant/lib/python3.4/site-packages/mirobo/airpurifier.py”, line 138, in led_brightness
return LedBrightness(self.data[“led_b”])
File “/usr/lib/python3.4/enum.py”, line 222, in call
return cls.new(cls, value)
File “/usr/lib/python3.4/enum.py”, line 457, in new
raise ValueError("%r is not a valid %s" % (value, cls.name))
ValueError: 10 is not a valid LedBrightness
```

syssi:
> It’s possible to control the brightness of the led?
No,only on or off.

PR #254, opened 2018-03-07
Support of the Air Purifier V3 added (Closes: #231)

syssi:
The Air Purifier V3 doesn't support led brightness but provides a strange value for the property nevertheless. cp. #231


rytilahti:
Looks god, at some point when we have more proper tests for the device protocols we may want to do runtime checking per model basis.

PR #257, opened 2018-03-08
Support of the Chuang Mi Plug V3 

syssi:


rytilahti:
Would you like to get this into the next release or shall it wait?

miio/plug_v3.py
```diff
@@ -0,0 +1,115 @@
...
+
+        properties_count = len(properties)
+        values_count = len(values)
+        if properties_count != values_count:
+            _LOGGER.debug(
+                "Count (%s) of requested properties does not match the "
+                "count (%s) of received values.",
+                properties_count, values_count)
+
+        load_power = self.send("get_power", [])  # Response: [300]
```
syssi: @rytilahti This is the only exotic part. Feel free to merge.

rytilahti: Ok, let's merge this for the next release and check later if we can generalize the plugs at some later point.

PR #258, opened 2018-03-08
Tests: Non-essential code removed

syssi:


syssi:
I think these lines aren't needed / explanatory. Did I miss something?

rytilahti:
Yeah, those are not needed.

PR #260, opened 2018-03-09
Provide the mac address as property of the device info

syssi:
I want to use the mac address as `unique_id` at home assistant.

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -95,6 +95,12 @@ class DeviceInfo:
             return self.data["hw_ver"]
         return None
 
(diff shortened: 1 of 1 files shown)


miio/device.py
```diff
@@ -95,6 +95,12 @@ def hardware_version(self) -> Optional[str]:
             return self.data["hw_ver"]
         return None
 
+    def mac_address(self) -> Optional[str]:
```
rytilahti: Would it be better if this would also be a property like the rest of them?

syssi:
Sure! It was too late for me.

PR #270, opened 2018-03-16
Merge of the Chuangmi Plug V1, V2, V3 and M1

syssi:
@rytilahti Could you provide a recommendation for the "device model enum" and the discovery device map?


miio/device.py
```diff
@@ -115,7 +115,7 @@ class Device:
     This class should not be initialized directly but a device-specific class inheriting
     it should be used instead of it."""
     def __init__(self, ip: str = None, token: str = None,
-                 start_id: int=0, debug: int=0, lazy_discover: bool=True) -> None:
+                 start_id: int=0, debug: int=0, lazy_discover: bool=True, model: str=None) -> None:
```
rytilahti: Is it truly necessary to introduce this also on the parent level, or would it be fine to have it just in the subclass?

miio/discovery.py
```diff
@@ -17,9 +17,9 @@
     "roborock-vacuum-s5": Vacuum,
     "chuangmi-plug-m1": Plug,
     "chuangmi-plug-v2": Plug,
-    "chuangmi-plug-v1": PlugV1,
-    "chuangmi-plug_": PlugV1,
-    "chuangmi-plug-v3": PlugV3,
+    "chuangmi-plug-v1": ChuangmiPlug,
+    "chuangmi-plug_": ChuangmiPlug,
+    "chuangmi-plug-v3": ChuangmiPlug,
```
rytilahti: For passing a parameter here you can use `functools.partial`, so something like:
```
partial(ChuangmiPlug, model=MyModel)
```
can be used in this case.

I'm wondering whether that should be an enum, or simply an integer `model_version`, which would default to either of them? We could also derive the model (if not given) per default from the status information (to adapt the setters accordingly), as it's just one extra attribute & different naming?

miio/__init__.py
```diff
@@ -1,11 +1,10 @@
 # flake8: noqa
 from miio.protocol import Message, Utils
 from miio.vacuumcontainers import (VacuumStatus, ConsumableStatus, DNDStatus,
-                                   CleaningDetails, CleaningSummary, Timer)
+                                   CleaningDetails, CleaningSummary, Timer, )
```
rytilahti: Is this a necessary change?

syssi: No. :-)

miio/discovery.py
syssi: I will care about tomorrow.

syssi: The `check_and_create_device` method doesn't enter the `inspect.isclass(v)` branch anymore if I use a partial. I'm unsure about a good implementation.

rytilahti: Indeed.. Hmm, maybe adding a check if it's partial and extracting the class out from it?
```
>>> from functools import partial
>>> class A:
>>>     pass

>>> partial(A, foo=1).func
<class '__main__.A'>
```

rytilahti:
One open question is whether we want to keep backwards compatibility with the old class naming? If yes, that could be simply done by creating small wrapper classes located inside `chuangmi_plug`, which are also exported (and yielding a warning about deprecation).

syssi:
Could you review the changes? I don't know how to improve the tests. The constructor of the old classes should be called and the base class must be mocked somehow.

miio/__init__.py
```diff
@@ -3,9 +3,10 @@
 from miio.vacuumcontainers import (VacuumStatus, ConsumableStatus, DNDStatus,
                                    CleaningDetails, CleaningSummary, Timer)
 from miio.vacuum import Vacuum, VacuumException
-from miio.plug import Plug
-from miio.plug_v1 import PlugV1
-from miio.plug_v3 import PlugV3
+from miio.chuangmi_plug import Plug
+from miio.chuangmi_plug import PlugV1
+from miio.chuangmi_plug import PlugV3
+from miio.chuangmi_plug import ChuangmiPlug
```
rytilahti: `from miio.chuangmi_plug import (Plug, PlugV1, PlugV3, ChuangiPlug)` would be nicer here.

miio/chuangmi_plug.py
```diff
@@ -0,0 +1,188 @@
...
+        warnings.warn(
+            "This device class is deprecated. Please use the ChuangmiPlug "
+            "in future and select a model by parameter 'model'.",
+            DeprecationWarning, stacklevel=2)
+
+        super().__init__(ip, token, start_id, debug, lazy_discover,
+                         model=model)
+
+
+class Plug(ClassDeprecatedPlug):
```
rytilahti: I created PR #276 to move the deprecated decorator out from vacuumcontainers, it should be used to decorate these classes to inform anyone trying to initialize these classes.

syssi:
Done :-)

rytilahti:
Great, thanks! :100:

PR #271, opened 2018-03-17
Water Purifier: Some properties added

syssi:
May be it's easier to retrieve some feedback if the device support is more complete.


rytilahti:
:+1:

PR #272, opened 2018-03-17
Air Conditioning Companion: LED status fixed

syssi:


rytilahti:
I was merging these as they seem fairly simple, thanks for that. Anyway, I noticed that the tests are failing now, do you mind fixing that?

PR #285, opened 2018-03-29
WiFi Repeater: Wifi roaming and signal strange indicator added

syssi:


miio/wifirepeater.py
```diff
@@ -86,12 +91,38 @@ def set_wifi_roaming(self, wifi_roaming: bool):
             'wifi_explorer': int(wifi_roaming)
         }])
 
-    def set_configuration(self, ssid: str, password: str, hidden: bool = False,
-                          wifi_roaming: bool = False):
+    def set_configuration(self, wifi_roaming: bool = False, ssid: str = None,
```
rytilahti: Would it make sense to have a separate method for enabling and disabling the roaming? Otherwise this looks good to be merged.

Issue #282, opened 2018-03-28
Support for Xiaomi Mi Smart Electric Rice Cooker


syssi:
```
get_prop '["func", "menu", "stage", "temp", "t_func", "t_precook", "t_cook", "setting", "delay", "version", 
get_prop '[{"all"}]'
"favorite"]'
get_temp_history '[]'
miIO.info '[]'
set_func '["nowarn"]'
set_func '["ack"]'
# stopCook
set_func '["end02"]' # if (model == COOKER_MODEL_PRESS1 && firmwareVersion >= 39) OR (model == COOKER_MODEL_NORMAL2 && firmwareVersion >= 22) OR (model == COOKER_MODEL_NORMAL3 && firmwareVersion >= 9) OR (model == COOKER_MODEL_PRESS2) OR (model == COOKER_MODEL_NORMAL1) OR (model == COOKER_MODEL_NORMAL4) OR (model == COOKER_MODEL_NORMAL5)
set_func '["end"]' # else

set_start '["cookingProfileHexdata"]'
set_interaction '["new String[]{setting + "," + shutLedDelay + "," + lidOpenTimeout + "," + lidOpenTimeoutAlarmTime"]'
set_menu '["hexdata"]' -> CookingActivity.this.recipe.getProfile().toHexData()
```

```
$ mirobo info
chunmi.cooker.normal2 v1.2.8 ([MAC]) @ [IP] - token: [TOKEN]

$ mirobo raw_command get_prop '[“func”, “menu”, “stage”, “temp”, “t_func”, “t_precook”, “t_cook”, “setting”, “delay”, “version” ]'
Sending cmd get_prop with params [‘func’, ‘menu’, ‘stage’, ‘temp’, ‘t_func’, ‘t_precook’, ‘t_cook’, ‘setting’, ‘delay’, ‘version’]
[‘waiting’, ‘0001’, ‘null’, ‘22’, ‘60’, ‘-1’, ‘60’, ‘1407’, ‘05060f’, ‘00030017’, ‘0100’, ‘13281323ffff011effff010000001617’]

$ mirobo raw_command get_prop '[“all” ]'
Sending cmd get_prop with params [‘all’]
[‘waiting’, ‘0001’, ‘null’, ‘22’, ‘60’, ‘-1’, ‘60’, ‘1407’, ‘05060f’, ‘00030017’, ‘0100’, ‘13281323ffff011effff010000001617’]
```

[...]

PR #287, opened 2018-03-29
Basic support for Xiaomi Mi Smart Electric Rice Cooker

syssi:
Closes: #282


miio/cooker.py
```diff
@@ -0,0 +1,199 @@
...
+MODEL_NORMAL4 = 'chunmi.cooker.normal3'
+MODEL_NORMAL3 = 'chunmi.cooker.normal4'
+MODEL_NORMAL5 = 'chunmi.cooker.normal5'
+
+MODEL_PRESS = [MODEL_PRESS1, MODEL_PRESS2]
+MODEL_NORMAL = [MODEL_NORMAL1, MODEL_NORMAL2, MODEL_NORMAL3, MODEL_NORMAL4,
+                MODEL_NORMAL5]
+
+
+class OperationMode(enum.Enum):
```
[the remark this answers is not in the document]
syssi: Fixed

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+MODEL_NORMAL5 = 'chunmi.cooker.normal5'
+
+MODEL_PRESSURE = [MODEL_PRESSURE1, MODEL_PRESSURE2]
+MODEL_NORMAL = [MODEL_NORMAL1, MODEL_NORMAL2, MODEL_NORMAL3, MODEL_NORMAL4,
+                MODEL_NORMAL5]
+
+MODEL_NORMAL_GROUP1 = [MODEL_NORMAL2, MODEL_NORMAL5]
+MODEL_NRRMAL_GROUP2 = [MODEL_NORMAL3, MODEL_NORMAL4]
+
+COOKING_STAGE_NAME = [
```
rytilahti: Please merge this and cooking state descriptions into a dict.

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+        32 minutes:
+        161515161c242a3031302f2eaa2f2f2e2f2e302f2e2d302f2f2e2f2f2f2f343a3f3f3d3e3c3d3c3f3d3d3d3f3d3d3d3d3e3d3e3c3f3f3d3e3d3e3e3d3f3d3c3e3d3d3e3d3f3e3d3f3e3d3c3f3e3d3c3f3e3d3c3f3f3d3d3e3d3d3f3f3d3d3f3f3e3d3d3d3e3e3d3daa3f3f3f3f3f414446474a4e53575e5c5c5b59585755555353545454555554555555565656575757575858585859595b5b5c5c5c5c5d5daa5d5e5f5f606061
+
+        55 minutes:
+        161515161c242a3031302f2eaa2f2f2e2f2e302f2e2d302f2f2e2f2f2f2f343a3f3f3d3e3c3d3c3f3d3d3d3f3d3d3d3d3e3d3e3c3f3f3d3e3d3e3e3d3f3d3c3e3d3d3e3d3f3e3d3f3e3d3c3f3e3d3c3f3e3d3c3f3f3d3d3e3d3d3f3f3d3d3f3f3e3d3d3d3e3e3d3daa3f3f3f3f3f414446474a4e53575e5c5c5b59585755555353545454555554555555565656575757575858585859595b5b5c5c5c5c5d5daa5d5e5f5f60606161616162626263636363646464646464646464646464646464646464646364646464646464646464646464646464646464646464646464646464aa5a59585756555554545453535352525252525151515151
+        """
+        self.data = [int(data[i:i + 2], 16) for i in range(0, len(data), 2)]
+
+    @property
+    def temperatures(self) -> list[int]:
```
rytilahti: You need to `from typing import List` and use `List` here (and elsewhere).

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+    Online = 'online'
+    PreCook = 'precook'
+    Resume = 'resume'
+    ResumeP = 'resumep'
+    Start = 'start'
+    StartP = 'startp'
+    Cancel = 'Отмена'
+
+
+class TemperatureHistory:
```
rytilahti: doctstrings would be nice for this class, too. What does this dow and how can I use it? It's a container of temperatures recorded every X time interval? Is there a limit how many historical values are available?

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+    KeepTemp = 'keep_temp'
+    Notice = 'notice'
+    Offline = 'offline'
+    Online = 'online'
+    PreCook = 'precook'
+    Resume = 'resume'
+    ResumeP = 'resumep'
+    Start = 'start'
+    StartP = 'startp'
+    Cancel = 'Отмена'
```
rytilahti: That looks weird. What does it mean that they are potential candidates? Is it necessary that these are included?

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+        s = "<TemperatureHistory temperatures=%s>" % self.data
+        return s
+
+
+class CookerCustomizations:
+    def __init__(self, custom: str):
+        """
+        Example values:
+
+        ffffffffffff011effff010000001d1f,
```
rytilahti: It would be nice to have some sort of short description for the bytes / how to interpret these.

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+             self.zhuzhou_cooking,
+             self.favorite_appointment,
+             self.favorite_cooking)
+        return s
+
+
+class CookingStage:
+    def __init__(self, stage: str):
+        """
+        Example timeouts: 'null', 02000000ff, 03000000ff, 0a000000ff, 1000000000
```
rytilahti: Same as above here. Description of bytes that is.

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+    @property
+    def taste(self) -> int:
+        return int(self.stage[6:8], 16)
+
+    @property
+    def taste_phase(self) -> int:
+        phase = self.taste / 33
+
+        if phase > 2:
+            return 2
```
rytilahti: This returns always either None or 2, is that correct? If yes, I'd say make returning None explicit + add `Optional` to return value.

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+
+    @property
+    def name(self) -> str:
+        try:
+            return COOKING_STAGE_NAME[self.state]
+        except IndexError:
+            return 'Cooking finished'
```
rytilahti: Why "cooking finished" is not in stages?

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+             self.taste,
+             self.taste_phase,
+             self.stage)
+        return s
+
+
+class InteractionTimeouts:
+    def __init__(self, timeouts: str = None):
+        """
+        Example timeouts: 05040f, 05060f
```
rytilahti: How are these interpreted and what are timeouts?

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+
+    @pressure_supported.setter
+    def pressure_supported(self, supported: bool):
+        if supported:
+            self.settings[0] |= 1
+        else:
+            self.settings[0] &= 254
+
+    @led_on.setter
+    def led_on(self, on: bool):
```
rytilahti: I'd recommend ordering getters & setters next to each other.

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+        return int(self.data['menu'], 16)
+
+    @property
+    def stage(self) -> Optional[CookingStage]:
+        """Current stage if cooking."""
+        stage = self.data['stage']
+        if len(stage) == 10 and stage.hexdigits:
+            return CookingStage(stage)
+
+        return None
```
rytilahti: When does this happen? Would it be better to raise an exception if that's an error case?

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+        """
+        Current temperature, if idle.
+
+        Example values: *29*, 031e0b23, 031e0b23031e
+        """
+        value = self.data['temp']
+        if len(value) == 2 and value.isdigit():
+            return int(value)
+
+        return None
```
rytilahti: Exception instead of returning None here too? (and below also, for error cases)

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+        default_output=format_output("Cooking stopped"),
+    )
+    def stop(self):
+        """Stop cooking."""
+        self.send('set_func', ['end02'])
+
+    @command(
+        default_output=format_output("Cooking stopped"),
+    )
+    def stop_outdated_firmware(self):
```
rytilahti: Is it possible to check if outdated stop is necessary, and just expose the `stop` command?

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+
+MODEL_PRESSURE1 = 'chunmi.cooker.press1'
+MODEL_PRESSURE2 = 'chunmi.cooker.press2'
+MODEL_NORMAL1 = 'chunmi.cooker.normal1'
+MODEL_NORMAL2 = 'chunmi.cooker.normal2'
+MODEL_NORMAL4 = 'chunmi.cooker.normal3'
+MODEL_NORMAL3 = 'chunmi.cooker.normal4'
+MODEL_NORMAL5 = 'chunmi.cooker.normal5'
+
+MODEL_PRESSURE = [MODEL_PRESSURE1, MODEL_PRESSURE2]
```
rytilahti: If MODEL_PRESSURE{1,2} are used nowehere else than here, better just simply use them instead of exposing a global variable.

miio/cooker.py
```diff
@@ -0,0 +1,706 @@
...
+MODEL_NORMAL = [MODEL_NORMAL1, MODEL_NORMAL2, MODEL_NORMAL3, MODEL_NORMAL4,
+                MODEL_NORMAL5]
+
+MODEL_NORMAL_GROUP1 = [MODEL_NORMAL2, MODEL_NORMAL5]
+MODEL_NRRMAL_GROUP2 = [MODEL_NORMAL3, MODEL_NORMAL4]
```
rytilahti: Typo in nrrmal. `MODEL_NORMAL1` does not belong to any group? Are these global variables necessary?

miio/cooker_profiles.json
```diff
@@ -0,0 +1,195 @@
+{
```
rytilahti: Maybe `miio/data/cooker_profiles.json` to keep it from polluting the main source dir?

miio/cooker.py
syssi: These are possible values of the property "func". A few are internal states of the android app probably. I've just collected all occurrences of string comparisons.

miio/cooker.py
syssi: If the cooker isn't cooking it returns the room temperature here (property: "temp").

miio/cooker.py
syssi: It's possible but ugly. I've to compare a lot of firmware versions and model types. It was to complicated to implement at the first step.

miio/cooker.py
syssi: These consts are important in future for all the edge-cases.

miio/cooker.py
rytilahti: Ok, then let it be like that for now.

miio/cooker.py
rytilahti: Ok, what are the values otherwise? Those hex-encoded ones? Should this be printing a warning when used and the temperature is not available, and point out what to do instead of it?

miio/cooker_profiles.json
rytilahti: To add, the setup.py and/or manifest file will require also an update, otherwise the json file won't be added when generating a package.

miio/cooker.py
syssi: Good catch! Fixed.

miio/cooker.py
syssi: There are a few stages called "Cooking" finished:

```
        if (state.equals("10") || state.equals("11") || state.equals("12")) {
            setStage(Constants.COOK_STATUS_FINISH);
```

miio/cooker.py
syssi: Note: Discussed at discord.

PR #291, opened 2018-04-01
WiFi Speaker support improved

syssi:


rytilahti:
I suppose we can close #69 now. As the speaker unfortunately doesn't allow direct playback of urls, I haven't had much use for my speaker :(

PR #300, opened 2018-04-04
WIP: Air Humidifier: Docstrings for trans_level, depth and dry mode added

syssi:


rytilahti:
Is this ready to be merged? :-)

syssi:
I've concerns with this PR. The improvement is tiny and the risk to break the humidifier support is high. I will try to do some synthetic testing against my Air Purifier to be sure.

PR #308, opened 2018-04-14
Device: Provide a cached device info always

syssi:
The firmware and hardware version of a device is important sometimes. It should be retrieved once and always available afterwards.

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -114,7 +114,8 @@ class Device(metaclass=DeviceGroupMeta):
     This class should not be initialized directly but a device-specific class inheriting
     it should be used instead of it."""
     def __init__(self, ip: str = None, token: str = None,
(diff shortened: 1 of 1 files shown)


miio/device.py
```diff
@@ -114,7 +114,8 @@ class Device(metaclass=DeviceGroupMeta):
     This class should not be initialized directly but a device-specific class inheriting
     it should be used instead of it."""
     def __init__(self, ip: str = None, token: str = None,
-                 start_id: int=0, debug: int=0, lazy_discover: bool=True) -> None:
+                 start_id: int=0, debug: int=0, lazy_discover: bool=True,
+                 device_info: DeviceInfo=None) -> None:
```
rytilahti: What's the use-case for passing this as a parameter? Using the results from discovery directly or..?

syssi:
Correct. I'm querying the info to identify the device type. I would like to pass the result instead of request it again.

PR #310, opened 2018-04-16
Added new commands app_goto_target and app_zoned_clean

ciB89:
In theory, app_zoned_clean should be able to take multiple zones at once - but I do not know how to implement this.


ciB89:
Yeah, I'm newbie sorry :P

PR #311, opened 2018-04-23
Support of multiple zones for app_zoned_cleaning

ciB89:
The vacuum only accepts zones within a list, so even a single list has to be passed like this:

[[x1,y1,x2,y2, iterations]]

The current code only sends "[x1,y1,x2,y2, iterations]" to the vacuum, so it does not work. I tried to change the code so that it accepts multiple zones, but I am not sure if it works like this. In addition, I have no no idea how to define the click.argument for a list. So I definitely going to need help on this one.


syssi:
Do you still need help or does it work already?

PR #317, opened 2018-05-10
Air Conditioning Companion: Rewrite a captured command before replay

syssi:
cp. https://github.com/syssi/xiaomi_airconditioningcompanion/issues/5


miio/airconditioningcompanion.py
```diff
@@ -306,14 +306,29 @@ def learn_stop(self, slot: int=STORAGE_SLOT_ID):
         return self.send("end_ir_learn", [slot])
 
     @command(
-        click.argument("command", type=str),
+        click.argument("model", type=str),
+        click.argument("code", type=str),
         default_output=format_output("Sending the supplied infrared command")
     )
-    def send_ir_code(self, command: str):
+    def send_ir_code(self, model: str, code: str):
```
yawor: I'm not a fan of working on a hex string. I think it would be better to convert both model and code strings to bytearrays and work on them. Then at the end, convert it back to a hex string.

miio/airconditioningcompanion.py
```diff
@@ -306,14 +306,29 @@ def learn_stop(self, slot: int=STORAGE_SLOT_ID):
...
+        """
+
+        self_function = "{:02X}".format(121)
+        # FE + 0487 + 00007145 + 9470 + 1FFF7FFF + 06 + 0042 + 27 + 4E + 0025002D008500AC01...
+        command = code[0:2] + model[4:8] + model[8:16] + '9470' + \
+            '1FFF' + self_function + 'FF' + \
+            code[26:28] + code[28:32] + '27'
+
+        checksum = sum([int(command[i:i + 2], 16) for i in range(0, len(command), 2)])
+        checksum = "{:02X}".format(checksum % 256)
```
yawor: `checksum & 0xFF` produces the same value as `checksum % 256`, but should be much faster (bit-wise operation vs arithmetic one).

syssi:
@yawor The bytearray variant is a bit hard to read now. I'm doing it correct?

miio/airconditioningcompanion.py
```diff
@@ -310,33 +310,29 @@ def learn_stop(self, slot: int=STORAGE_SLOT_ID):
...
 
     @command(
-        click.argument("model", type=str),
-        click.argument("code", type=str),
+        click.argument("model", type=bytes),
+        click.argument("code", type=bytes),
         default_output=format_output("Sending the supplied infrared command")
     )
-    def send_ir_code(self, model: str, code: str, slot: int=0):
+    def send_ir_code(self, model: bytes, code: bytes, slot: int=0):
```
yawor: I've been thinking about keeping strings here as args but converting them to `bytearray`/`bytes` right away. Having `bytes` as arg type for command line is probably not very good idea :). Maybe there should be two methods? Main one taking `bytes` like this one, but the `@command` should be applied to a wrapper method, which takes hex strings, converts them and calls the `bytes` one.

yawor: Optionally (maybe even better solution) would be to add converting functions to click arguments definition to convert hex strings from command line to bytes.

yawor:
@syssi seems to be OK. I would only do something about click args, because it'll probably be hard to enter raw bytes on command line. Unless click is smart enough to do this conversion automatically for bytes type.

PR #321, opened 2018-05-20
Preparation of release 0.4.0

syssi:
A new release is needed. There are so many new and cool features. I know a proper documentation of the miiocli is missing yet but may be would get some things done here? I lost the overview.


rytilahti:
Thanks for pushing this onwards, I agree that it's time to have a new release! I think it'd be enough to have a short introduction on the new cli tool, the documentation can always be extended as needed, so it doesn't have to be anything spectacular but just a reminder for anyone reading that there's such a tool, too.

PR #324, opened 2018-05-20
Update README

syssi:
Both powerstrip versions are supported and needs to selected.


rytilahti:
:+1: what do you mean with beselect? Other than that, feel free to merge.

Issue #325, opened 2018-05-28
IR remote chuangmiir module


AlbrtoSam:
I have got a IR module working using mirobo:

mirobo --ip 192.168.1.225 --token 12345 raw_command miIO.ir_play '{"freq": 38400, "code": "Z6VrAJwAAAAjAQAAugEAAHICAAAOAwAAfFoBAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAASEDAgQDAwEBAwMCAwEBAQMFASEDAgQDAwEBAwMCAwEBAQMFASEDAgQDAwEBAwMCAwEBAQMAA="}'
Sending cmd miIO.ir_play with params {'freq': 38400, 'code': 'Z6VrAJwAAAAjAQAAugEAAHICAAAOAwAAfFoBAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAASEDAgQDAwEBAwMCAwEBAQMFASEDAgQDAwEBAwMCAwEBAQMFASEDAgQDAwEBAwMCAwEBAQMAA='}

And it works fine.

Now i wanted to start using miiocli chuangmiir but i dont know the sintaxis, i also saw the help, but no luck:

Info works fine
miiocli chuangmiir --ip 192.168.1.225 --token be6778439fdc154b676c7c4376cc5968 info

But i dont know the sintaxis for play command: I got this error:

root@debian:/home/pi# miiocli chuangmiir --ip 192.168.1.225 --token be6778439fdc154b676c7c4376cc5968 play 2
Playing the supplied command
Traceback (most recent call last):
  File "/usr/local/bin/miiocli", line 11, in <module>
    load_entry_point('python-miio==0.3.9', 'console_scripts', 'miiocli')()
  File "/usr/local/lib/python3.5/dist-packages/miio/cli.py", line 43, in create_cli
    return cli(auto_envvar_prefix="MIIO")
  File "/usr/local/lib/python3.5/dist-packages/miio/click_common.py", line 53, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.5/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #328, opened 2018-06-02
Fix a name clash of click_common and the argument "command"

syssi:
of device methods (Closes: #325)


syssi:
There are some more methods with a parameter called "command". That's the cause, why I choose this approach.

PR #341, opened 2018-06-17
Fix ChuangmiPlugStatus repr format

syssi:


rytilahti:
Do you think it'd be a good idea simply to expose the available attributes in an array (like `__device_attrs__` or similar) and simply do one `__repr__()` in `Device`? Something akin to:
```
def __repr__(self):
    prefix = "<%s" % self.__class__.__name__
    attr_s = ""
    for attr in self.__device_attrs__:
        attr_s += "%s: %s " % (attr, self.data[attr])
    postfix = "@ %s" % self.host
    return prefix + attr_s + postfix
```

syssi:
I prefer the class properties instead of the miio api property names because of the naming and conversion.

rytilahti:
Ah, maybe that was put badly. So yeah, of course exposing the proper names & converted outputs. This is also something that the `json-output` could use (which currently simply delivers the `data` contents).

rytilahti:
I'm unsure why this wasn't merged already, seems to be no-brainer so let's get it done.

PR #343, opened 2018-06-18
RFC: Add discover method to miiocli

syssi:


yawor:
@syssi on which level do you want the discover method to be? Should it be implemented in each device group (for example miiocli airpurifier discover) or as the top command (miiocli discover)?

The second approach should be easier regarding the `--ip` and `--token` params, as they are added by each device group, so they are non existent at the top level. Is there any requirement that the discover should be in the device group?

rytilahti:
@yawor discovery should be done on the global scope, as its main purpose is for users to find to locate (all) available, supported devices. So `miiocli discover` is what is wanted here :-)

syssi:
@yawor On global scope cp. https://github.com/rytilahti/python-miio/pull/343#issuecomment-412377249

rytilahti:
Closing this in favor of #1013

Issue #351, opened 2018-07-11
Xiaomi Philips Zhirui Bedside Lamp


ardeus-ua:
Just received the lamp today and want to connect it to Home Assistant.
_https://brightbestlight.com/xiaomi-philips-zhirui-lamp-review/
How can I help to add the support for this lamp?

syssi:
Please extract the token and provide the output of the "info" command (mirobo/miiocli).

ardeus-ua:
philips.light.moonlight v1.5.6

syssi:
I'm looking for a mihome plugin which supports the device right now. Are you able to control the light with the MiHome app?

ardeus-ua:
Yes, MiHome automatically downloaded plugin for the lamp.

syssi:
Do you own a rooted phone and can you provide the plugin (apk file)?

ardeus-ua:
I have checked it with rooted android phone but didn’t found it in the folder where the plugins are usually stored (found 2 plugins for 2 other devices I use but not for the moonlight) will dig additionally.

ardeus-ua:
Found it! Here it is https://drive.google.com/file/d/1E4rbqvgnpd21y_lLfq-XLeMPq7IyKMlS/view?usp=drivesdk

ardeus-ua:
Is there any way I can help in testing? I can forward ports and send you the details via personal message

syssi:
```
# Model name
philips.light.moonlight

# Supported properties
["pow", "sta", "bri", "rgb", "cct", "snm", "spr", "spt", "wke", "bl", "ms", "mb", "wkp"]

# Supported methods
add_mb # Add miband
enable_bl # Night light
enable_sub_voice # Sub voice control?
enable_voice # Voice control
get_band_period # Bracelet work time
get_mb_rssi # Miband RSSI
miIO.bleGetNearbyBandList
get_wakeup_time
get_mb_mac # Miband MAC address
miIO.bleStartSearchBand
en_wakeup
set_band_period
set_bricct # Brightness & Color temperature
set_brirgb # Brightness & RGB
set_bright # Brightness
set_cct # Color temperature
set_rgb # RGB
set_mb_rssi
go_night # Night light / read mode
set_power
apply_fixed_scene
skip_breath
en_sleep
set_sleep_time
enable_mibs
set_wakeup_time
```

syssi:
@ardeus-ua Could you provide the output of

```
[...]

PR #359, opened 2018-08-10
Add basic Philips Moonlight support (Closes: #351)

syssi:


syssi:
I will try to improve the implementation gradually. I don't know the meaning of some properties right now.

PR #360, opened 2018-08-10
Add Xiaomi Airfresh VA2 support

syssi:
cp. https://github.com/syssi/xiaomi_airpurifier/issues/25

### DIFF
--- a/miio/airfresh.py
+++ b/miio/airfresh.py
@@ -87,6 +87,7 @@ class AirFreshStatus:
             try:
                 return LedBrightness(self.data["led_level"])
             except ValueError:
+                _LOGGER.error("Unsupported LED brightness discarded: %s", self.data["led_level"])
                 return None
 
         return None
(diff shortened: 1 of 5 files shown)


miio/airfresh.py
```diff
@@ -0,0 +1,292 @@
...
+        return OperationMode(self.data["mode"])
+
+    @property
+    def led_brightness(self) -> Optional[LedBrightness]:
+        """Brightness of the LED."""
+        if self.data["led_level"] is not None:
+            try:
+                return LedBrightness(self.data["led_level"])
+            except ValueError:
+                return None
```
rytilahti: Would it be a good idea to log the error here?

miio/airfresh.py
```diff
@@ -0,0 +1,292 @@
...
+        """Return True if buzzer is on."""
+        if self.data["buzzer"] is not None:
+            return self.data["buzzer"] == "on"
+
+        return None
+
+    @property
+    def child_lock(self) -> bool:
+        """Return True if child lock is on."""
+        return self.data["child_lock"] == "on"
```
rytilahti: Should this and the other properties also be protected against None?

syssi: This property is always available and doesn't need to be safeguarded.

PR #369, opened 2018-08-23
Add tqdm to requirements list

pluehne:
`tqdm` was introduced as a dependency earlier and added to [`setup.py`](https://github.com/rytilahti/python-miio/blob/e8b6489bb956be465f7738f58ee1c7835d817623/setup.py#L57) accordingly. However, this new dependency was not reflected in the textual requirements list.


rytilahti:
Thanks! :+1:

PR #371, opened 2018-08-29
Use enums for vacuum states and errors

rytilahti:
This PR adds two new enums, `VacuumError` and `VacuumState`, to report known states and errors for downstream users to use instead of mapping the error codes themselves.

For the time being calling `VacuumStatus.state` keeps returning the state as a human-readable string, which will be changed to return the VacuumState enum later on. The enum is available in `state_enum` and `state_str` can be used for the human-readable presentation.

`VacuumError` works similarly, `error` keeps returning `error_str` for the time being, `error_enum` returns the enum and `error_str` the human-readable presentation.

### DIFF
--- a/miio/vacuumcontainers.py
+++ b/miio/vacuumcontainers.py
@@ -107,7 +107,8 @@ class VacuumStatus:
         return int(self.data["state"])
 
     @property
-    @deprecated("Use state_str for string presentation, this will return an enum in the future.")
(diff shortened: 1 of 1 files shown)


rytilahti:
This is now obsolete. If any work in this direction is to be made, it should be a common API for all vacuum devices to allow easier downstream integrations.

PR #379, opened 2018-09-11
vacuum's got_error: compare against error code, not against the state

rytilahti:
The vacuum changes the state to error state only for a short period of time before returning to a "Disconnected" state if nothing else happens.

This will change the behaviour of `got_error` to check the error code against 0.


rytilahti:
Merging as there's been no objections.

PR #382, opened 2018-09-30
Expand documentation for token from Android backup

sgtio:
Add information about the initial setup of the device, and the steps
needed prior to creating the backup.

### DIFF
--- a/docs/discovery.rst
+++ b/docs/discovery.rst
@@ -104,14 +104,38 @@ or database from the Mi Home app.
 The procedure is briefly described below,
 but you may find the following links also useful:
 
-- https://github.com/jghaanstra/com.xiaomi-miio/blob/master/docs/obtain\_token\_mirobot\_new.md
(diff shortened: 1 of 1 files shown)


docs/discovery.rst
```diff
@@ -174,7 +198,7 @@ to the tool with ``--password <password>``.
             MAC: 28:6C:07:XX:XX:XX
     room1
             Model: yeelink.light.color1
-            IP address: 192.168.XXX.XXX
+            \IP address: 192.168.XXX.XXX
```
rytilahti: Why change this?

sgtio: yup, this is accidental. I'll fix it!
Thanks for the feedback!

sgtio:
I have squashed the commit with the typo fix. Please let me know if you prefer having 2 separate commits. Thanks!

rytilahti:
It is fine, we'll simply squash & merge it. Thanks a lot for helping with the documentation!

PR #376, opened 2018-09-02
Bump construct version (Closes: #374)

syssi:

PR #384, opened 2018-10-02
Prepare 0.4.2 - Dropping the Pin

### LINKED ISSUE #374: python-construct-2.9.45
first of all thank you for that great project! :)
current version of python-construct on arch linux is 2.9.45, but python-miio is pinned to 'construct==2.9.41'.
just bumped the required version within my local repo and looks like python-miio works just fine with 2.9.45.

rytilahti:
This release removes the version pinning for "construct" library as its API has been stabilized and we don't want to force our downstreams for our version choices.

Another notable change is dropping the "mirobo" package which has been deprecated for a very long time, and everyone using it should have had converted to use "miio" already.

This release also changes the behavior of vacuum's `got_error` property to signal properly if an error has occured. The previous behavior was based on checking the state instead of the error number, which changed after an error to 'idle' after a short while.

Replaces #376. Closes #374. Relates to https://github.com/home-assistant/home-assistant/pull/16562 .

edit: to add, this also changes the pypi classifier to stable, I suppose we are are stable enough for that!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,37 @@
 # Change Log
 
+## [0.4.2](https://github.com/rytilahti/python-miio/tree/0.4.2)
+
+This release removes the version pinning for "construct" library as its API has been stabilized and we don't want to force our downstreams for our version choices.
+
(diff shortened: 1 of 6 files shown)


rytilahti:
This PR should be rebased after #387 gets merged and before doing the release.

PR #397, opened 2018-10-24
Add cli interface for yeelight devices

rytilahti:
This extends yeelight support to allow controlling yeelight devices with `miiocli`, in case someone prefers this instead of using the developer mode.

```
Usage: miiocli yeelight [OPTIONS] COMMAND [ARGS]...

Options:
  --ip TEXT     [required]
  --token TEXT  [required]

Commands:
  info                      Get miIO protocol information from the device.
  off                       Power off.
  on                        Power on.
  raw_command               Send a raw command to the device.
  set_brightness            Set brightness.
  set_color_temp            Set color temp in kelvin.
  set_default               Set current state as default.
  set_developer_mode        Enable or disable the developer mode.
  set_name                  Set an internal name for the bulb.
  set_save_state_on_change  Enable or disable saving the state on changes.
  status                    Retrieve properties.
  toggle                    Toggle bulb state.
```

This can also be helpful for those who want to enable the developer mode without installing the app and binding the bulb to the cloud. Slightly related to https://github.com/home-assistant/home-assistant/pull/17699 this could use some documentation for potential users.

### DIFF
--- a/miio/yeelight.py
+++ b/miio/yeelight.py
@@ -1,7 +1,9 @@
 import warnings
+import click
 from enum import IntEnum
 from typing import Tuple, Optional
 
+from .click_common import command, format_output
 from .device import Device, DeviceException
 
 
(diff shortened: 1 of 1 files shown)


SmaginPV:
Any chance to cut the 0.4.3 release?

PR #398, opened 2018-10-27
Update commands in manual

olskar:
Update commands in manual to reflect commands that is in --help and that does work.

>  mirobo install_sound
Usage: mirobo [OPTIONS] COMMAND [ARGS]...
Try "mirobo --help" for help.

Error: No such command "install_sound".


rytilahti:
Thanks for the PR! I think this change was introduced also by click7+.

Issue #73, opened 2017-09-11
Send multiple params broken result


nutinshell:
Send multiple params to water purifier like `raw_command get_prop '["press","mode"]'`, the result always broken like `[300, 'error']`, only first one is right, no matter how much params sent.

`https://github.com/aholstenson/miio/` cli works well.

Capture shows very different behavior with mirobo, below is a single command ` --method get_prop --params '["press","temperature"]'` :

```
 ->  10.0.1.9 data= {"id":1,"method":"miIO.info","params":[]}
 <-  10.0.1.8 data= {"result":{"life":30300,"token":"a3d7f9****683","mac":"28***:03","fw_ver":"1.2.8","hw_ver":"ESP8266","model":"yunmi.waterpuri.v2","mcu_fw_ver":"0019","wifi_fw_ver":"1.4.0(30e0bd0)","ap":{"rssi":-58,"ssid":"Deep Space Voyager","bssid":"00***:A1"},"netif":{"localIp":"10.0.1.8","mask":"255.255.0.0","gw":"10.0.0.15"},"mmfree":10256},"id":1}
 ->  10.0.1.9 data= N/A
 <-  10.0.1.8 data= N/A
 ->  10.0.1.9 data= {"id":1,"method":"get_prop","params":["press","temperature"]}
 <-  10.0.1.8 data= {"result":{"life":30300,"token":"a3d7f9****683","mac":"28***:03","fw_ver":"1.2.8","hw_ver":"ESP8266","model":"yunmi.waterpuri.v2","mcu_fw_ver":"0019","wifi_fw_ver":"1.4.0(30e0bd0)","ap":{"rssi":-58,"ssid":"Deep Space Voyager","bssid":"00***:A1"},"netif":{"localIp":"10.0.1.8","mask":"255.255.0.0","gw":"10.0.0.15"},"mmfree":10256},"id":1}
 ->  10.0.1.9 data= {"id":1,"method":"get_prop","params":["press","temperature"]}
 <-  10.0.1.8 data= {"result":[300,29],"id":1}
```
any idea?

rytilahti:
Could you please post also the same mirobo command with `-d` and check out if the JSON payloads differ?

nutinshell:
> Could you please post also the same mirobo command with -d and check out if the JSON payloads differ?

hi @rytilahti  it's like:

```
INFO:mirobo.vacuum_cli:Debug mode active
[...]

PR #405, opened 2018-11-07
Waterpurifier: Divide properties into multiple requests (Closes: #73)

syssi:


rytilahti:
I'm wondering how it works with the javascript miio just fine, but maybe they are also splitting it in the background? Feel free to merge though :+1:

PR #411, opened 2018-11-16
Use pause instead of stop for home command

rytilahti:
This extends the support to more vacuum models, tested to work on gen1 vacuum as previously.

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -54,7 +54,11 @@ class Vacuum(Device):
 
     @command()
     def stop(self):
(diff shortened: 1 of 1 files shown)


Luc3as:
Oh I see you are working on the same as me, I did not work with GIT for some time. So basically you did some nice comments in code, I let this to you. 
Thank you for making this better ;)

rytilahti:
Hehe, yeah, I wanted to get it into a releasable/testable state asap. I hope we can make a new release this weekend! :-)

PR #414, opened 2018-11-17
Moonlight: Provide property rgb as tuple

syssi:


syssi:
May be `set_rgb` should be improved, too and we need a method per direction? Do you prefer a name? @rytilahti

rytilahti:
I would say we want to have it both ways, maybe just `int_to_rgb` and `rgb_to_int` to do the conversions, and changing `set_rgb` to take a tuple as an input?

syssi:
Alright. I will care about!

syssi:
TODO: set_rgb + miiocli compat

syssi:
@rytilahti I'm done. :-)

rytilahti:
Looking good! :+1:

PR #435, opened 2018-12-12
Moonlight: Fix parameters of the set_rgb api call

syssi:
cp. https://github.com/syssi/philipslight/issues/21

### DIFF
--- a/miio/philips_moonlight.py
+++ b/miio/philips_moonlight.py
@@ -186,7 +186,7 @@ class PhilipsMoonlight(Device):
             if color < 0 or color > 255:
                 raise PhilipsMoonlightException("Invalid color: %s" % color)
 
-        return self.send("set_rgb", [rgb_to_int(rgb)])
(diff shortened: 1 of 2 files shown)


miio/philips_moonlight.py
```diff
@@ -241,7 +241,9 @@ def set_brightness_and_rgb(self, brightness: int, rgb: Tuple[int, int, int]):
             if color < 0 or color > 255:
                 raise PhilipsMoonlightException("Invalid color: %s" % color)
 
-        return self.send("set_brirgb", [brightness, rgb_to_int(rgb)])
+        params = list(rgb)
+        params.append(brightness)
```
rytilahti: I think `[brightness, *rgb]` would also work here, but this is fine to me, so feel free to merge as you wish.

syssi: Yep. Work like a charm!

Issue #422, opened 2018-11-25
Some devices not responding across subnets.


cnrd:
Hi

I'm trying to change around a couple of things in my HASS setup, which seems to have surfaced a bug in the communication with chuangmi_ir devices.

I can ping the device just fine:
```
ping 10.0.3.18
PING 10.0.3.18 (10.0.3.18) 56(84) bytes of data.
64 bytes from 10.0.3.18: icmp_seq=1 ttl=254 time=5.14 ms
64 bytes from 10.0.3.18: icmp_seq=2 ttl=254 time=6.86 ms
64 bytes from 10.0.3.18: icmp_seq=3 ttl=254 time=3.88 ms
64 bytes from 10.0.3.18: icmp_seq=4 ttl=254 time=2.10 ms
--- 10.0.3.18 ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3005ms
rtt min/avg/max/mdev = 2.107/4.502/6.866/1.742 ms
```

but if i try to get info:
```
mirobo --ip 10.0.3.18 --token [TOKEN] info
ERROR:miio.device:Unable to discover a device at address 10.0.3.18
Error: Unable to discover the device 10.0.3.18
```
doing the same thing with my Roborock Vacuum works fine:
```
mirobo --ip 10.0.3.17 --token [TOKEN]
State: Charging
Battery: 100 %
Fanspeed: 100 %
Cleaning since: 0:02:34
Cleaned area: 2.765 m
```

It does work fine if I have an IP in the same subnet:
```
mirobo --ip 10.0.3.18 --token [TOKEN] info
chuangmi.ir.v2 v1.2.4_38 ([MAC]) @ 10.0.3.18 - token: [TOKEN]
```

I have tested on 2 different physical machines in different subnets, so it shouldn't be a problem with my network configuration.

ivancwc:
yes this is definitely a bug.  i bought two chuangmi_ir devices and encountered same issue

cnrd:
@syssi any chance you know what is going on here? :-)

syssi:
I own the device and will try to reproduce the issue. I don't know what's going on here. The code is untouched for month. I've no clue why it's broken now.

cnrd:
I just don't think we have catched this bug before, is it only manifests when trying to communicate across subnets (in my case routed VLANs)

In my case the IR device is in 10.0.3.0/24
If I try to access it from another device in 10.0.3.0/24 then everything is fine.
[...]

Issue #152, opened 2018-01-03
Unable to discover using API


CODeRUS:
Currently discover API is very strange. It prints ip addresses to log, but this info is not accessible using api. When trying Device.discover('<broadcast>') or object.do_discover(), i can not obtain discovered device ip address for doing things.

rytilahti:
The discovery interface (based on hello messages) is very ugly indeed, `do_discover()` should probably be renamed to `send_handshake()` or similar as it just sends a hello packet to already known IP address.

Using `Discovery` class is the preferred method for discovering devices, which however does not work on the vacuum if it e.g. is paired and has no cloud connectivity.

CODeRUS:
Can you point to example using Discovery for searching for unpaired device? I am a bit stuck here.

rytilahti:
Oh, I'm unsure if the unpaired one responds over mDNS, but you could try to connect to it (when it's unpaired) and simply run `mirobo discover` to see if it detects it. If yes, then something like this should work:
```
from miio import Discovery
devices = Discovery.discover_mdns()
for ip, dev in devices.items():
    print("%s at addr %s" % (dev, ip))
```

Note that the devices are returned first after enter or some key is pressed (I think a timeout could be added as an option here), and they are initialized accordingly (so a vacuum should be of type `Vacuum`).

CODeRUS:
`mirobo discover` after couple of minutes show nothing...

rytilahti:
Apparently it doesn't use mdns when unpaired :( this leaves us with the fact that the discovery/handshake api should be made better, so I'll leave this issue open.

PR #437, opened 2018-12-12
Revise device (broadcast-using) discovery process

rytilahti:
This is the initial commit to rework the handshake/discovery process,
separating them to separate entities.

The discovery part from Device class is moved into MiIODiscovery class usable for broadcast discovery of available devices.
Also, instead of printing out the details from discovered devices, a list of Devices is being returned.
The discover() method of the class is extended to allow discovery on all network interfaces (instead of using 255.255.255.255),
making it easier to perform the initial configuration of non-provisioned devices and/or work with devices in other subnets.

Although the change cause some duplicate code (mainly the handshake payload),
this will make the responsibilities of corresponding classes clearer and code more usable for 3rd party developers.

There will be cleanups and (likely also) renamings of some of the parts, but I wanted to make
this available for others to comment.

TODO:
* Upcasting the Device to a corresponding implementation class (miIO.info model information could be useful, too..)
* Decide how to handle the returning of discovered devices. Use a separate DiscoveredDevice class which can be used
  to initialize the real device, or use the existing Device and do the casting trick?

Related to #152 (and maybe #422).

### DIFF
--- a/miio/ceil_cli.py
+++ b/miio/ceil_cli.py
@@ -67,7 +67,7 @@ def cli(ctx, ip: str, token: str, debug: int):
 @cli.command()
 def discover():
     """Search for plugs in the network."""
-    miio.Ceil.discover()
+    miio.Ceil.send_handshake()
 
 
 @cli.command()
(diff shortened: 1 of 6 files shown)


rytilahti:
The codebase has changed so much that improving the discovery process should be done from scratch.

PR #438, opened 2018-12-17
Add control for persistent maps, no-go zones and barriers

rytilahti:
Completely untested for now, works only on gen2 and requires a fairly
recent firmware version supporting no-go zones and barriers.

Based on information from https://github.com/marcelrv/XiaomiRobotVacuumProtocol/issues/15


rytilahti:
I think this can be merged, hopefully someone with access to testing these features will add some documentation on how to use these features and improve them where needed.

Issue #205, opened 2018-02-08
DeviceError code -30001 (Resp Invalid JSON) - Philips Bulb


mouth4war:
This is a sporadic error I see happening about once in 30 minutes. It seems to be relevant to this library so I'm posting it here.

> 2018-02-08 16:18:00 ERROR (MainThread) [homeassistant.core] Error executing service <ServiceCall light.turn_on: transition=30.0, color_temp=283, entity_id=['light.guest_bedroom_wall_north']>
Traceback (most recent call last):
  File "/srv/homeassistant/lib/python3.6/site-packages/homeassistant/components/light/xiaomi_miio.py", line 194, in _try_command
    partial(func, *args, **kwargs))
  File "/usr/local/lib/python3.6/asyncio/futures.py", line 332, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/usr/local/lib/python3.6/asyncio/tasks.py", line 250, in _wakeup
    future.result()
  File "/usr/local/lib/python3.6/asyncio/futures.py", line 245, in result
    raise self._exception
  File "/usr/local/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/srv/homeassistant/lib/python3.6/site-packages/miio/philips_bulb.py", line 76, in on
    return self.send("set_power", ["on"])
  File "/srv/homeassistant/lib/python3.6/site-packages/miio/device.py", line 251, in send
    raise DeviceError(m.data.value["error"])
miio.device.DeviceError: {'code': -30001, 'message': 'Resp invalid json.'}

syssi:
Has your philips bulb internet access?

syssi:
I've seen the error today the first time too while doing load tests (45req/s):

```
DEBUG:miio.device:192.168.132.37:54321 >>: {'params': ['power', 'bright', 'cct', 'snm', 'dv'], 'method': 'get_prop', 'id': 4274}
DEBUG:miio.device:192.168.132.37:54321 (ts: 1970-01-01 00:36:18, id: 4274) << {'id': 4274, 'error': {'code': -30001, 'message': 'Resp invalid json.'}}
DEBUG:miio.click_common:Exception: {'code': -30001, 'message': 'Resp invalid json.'}
Traceback (most recent call last):
```
I will try to handle the error properly in future: retry / replay of the failed command.

mouth4war:
[...]

PR #449, opened 2019-01-07
Handle "resp invalid json" error (Closes: #205)

syssi:
Fixes https://github.com/home-assistant/home-assistant/issues/19787, too.

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -12,7 +12,7 @@ import construct
 from .click_common import (
     DeviceGroupMeta, command, format_output, LiteralParamType
 )
-from .exceptions import DeviceException, DeviceError
+from .exceptions import DeviceException, DeviceError, RecoverableError
 from .protocol import Message
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


miio/exceptions.py
```diff
@@ -6,3 +6,8 @@ class DeviceException(Exception):
 class DeviceError(DeviceException):
     """Exception communicating an error delivered by the target device."""
     pass
+
+
+class RecoverableError(DeviceException):
```
rytilahti: Looks good to me in general, but maybe `RecoverableDeviceError` deriving from `DeviceError` instead?

miio/device.py
```diff
@@ -269,7 +269,10 @@ def send(self, command: str, parameters: Any=None, retry_count=3) -> Any:
                           m.data.value["id"],
                           m.data.value)
             if "error" in m.data.value:
-                raise DeviceError(m.data.value["error"])
+                error = m.data.value["error"]
+                if "code" in error and error["code"] == -30001:
+                    raise RecoverableError(error)
```
rytilahti: We want to have this codepath tested too (at some point, maybe there are also other recoverable cases?) :-)

miio/exceptions.py
syssi: Fixed.

miio/device.py
syssi: Do you've an idea how to test this part?

rytilahti: I think the way to go would be mocking the Message.parse() to return with that error code, and then checking that it gets  called retry_count times. #437 may simplify the tests (or we might split the code even a bit more to make it easier to test). I think this can wait for now, feel free to merge.

PR #462, opened 2019-01-20
Extend miio-extract-tokens to allow extraction from yeelight app db

rytilahti:
In case someone prefers yeelight app over mi home..


rytilahti:
Maybe it is just badly named, but considering this is just a simple tool I think it is fine :-)

PR #465, opened 2019-01-21
Corrected link in discovery.rst and Xiaomi Air Purifier Pro fix

swiergot:
I hope you don't mind I also include a small fix for Air Purifier Pro. I cannot see any distinction between different Air Purifier models. Either this code doesn't care that some models might not accept higher values or so far all models have had a max of 16. Please advise.


syssi:
@swiergot Could you update the unit tests, too! :-)

PR #466, opened 2019-01-22
Rename speed property to avoid a name clash at HA

syssi:
cp. https://github.com/home-assistant/home-assistant/issues/19967


syssi:
This line must be updated as soon as python-miio 0.4.5 is released:

```
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/fan/xiaomi_miio.py#L235
```

PR #468, opened 2019-01-23
WIP: Vacuum: Provide human readable fan speeds

syssi:


miio/vacuum.py
```diff
@@ -306,19 +334,34 @@ def disable_dnd(self):
         return self.send("close_dnd_timer", [""])
 
     @command(
-        click.argument("speed", type=int),
+        click.argument("speed", type=str),
     )
-    def set_fan_speed(self, speed: int):
+    def set_fan_speed(self, speed: str):
```
syssi: I would prefer to pass a Enum here. @rytilahti any suggestions?

poma:
The new vacuum still supports setting speed in percentage (0-100) like in the old one, and it can still return percentage values that will not fit any of predefined values. So methods that work with raw values are still needed.

syssi:
Thanks for the hint! I will update my proposal soon!

tribut:
Not sure if this is the correct place to report it, but given that #523 is closed -
for the Xiaowa E2, the official app sets the fan speeds as following:

~~~
Silent 50
Standard 68
Strong 79
Max 100
~~~

rytilahti:
@syssi should we close this now that #643 is merged in?

rytilahti:
This is obsolete now.

Issue #467, opened 2019-01-23
miio-extract-tokens raises a TypeError when running against extracted SQLite database


domibarton:
When running `miio-extract-tokens` against an extracted SQLite database from an iOS backup, the script raises a `TypeError`:

```
(homeassistant) root@rpi:/opt/homeassistant# miio-extract-tokens /tmp/1672094281_mihome.sqlite
INFO:miio.extract_tokens:Reading database from /tmp/1672094281_mihome.sqlite
INFO:miio.extract_tokens:Reading tokens from Apple DB
Traceback (most recent call last):
  File "/opt/homeassistant/homeassistant/bin/miio-extract-tokens", line 11, in <module>
    sys.exit(main())
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/miio/extract_tokens.py", line 155, in main
    devices = list(reader.read_tokens(backup))
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/miio/extract_tokens.py", line 113, in read_tokens
    yield from self.read_apple()
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/miio/extract_tokens.py", line 74, in read_apple
    token = BackupDatabaseReader.decrypt_ztoken(dev['ZTOKEN'])
  File "/opt/homeassistant/homeassistant/lib/python3.5/site-packages/miio/extract_tokens.py", line 50, in decrypt_ztoken
    if len(ztoken) <= 32:
TypeError: object of type 'NoneType' has no len()
```

Please note this is a follow up of the PR #460, as I want to leave issue-related discussions in the issue section.

I'll post more as soon as I've done some additional testing.

domibarton:
@rytilahti I think I found the bug:

```
[...]

PR #469, opened 2019-01-24
Fix miio-extract-tokens for devices without tokens

domibarton:
The `miio.extra_tokens.BackupDatabaseReader.decrypt_ztoken` method will
raise a `TypeError` when a device in the SQLite database file has no
token defined. This occurs for local (bluetooth) devices, such as the
Mi Action Cam.

This change will fix the `miio-extra-tokens` utility, so that it will
work again, even if there are devices without tokens.

See also issue #467 for more informations.


rytilahti:
Looks good, thanks! :+1:

Issue #471, opened 2019-01-31
Xiaomi Vacuum - resume clean-up after pause


pszafer:
Hello,

Is it possible somehow to resume zone clean up and regular clean up?
This task is possible from Mi Home app.
I can send pause command, but with start command vacuum wants to clean everything.

rytilahti:
This would need to know what command is being used to continue the cleaning after a pause, either by checking what miio (javascript) does if it supports this, or analyzing the traffic to see what the app does.

pszafer:
I sniffed the command:
```
resume_zoned_clean
```

so `if last cleaning state was Zoned cleaning && paused`:
```
mirobo raw-command resume_zoned_clean
```

I guess it is more a job for home assistant to remember state before paused...

rytilahti:
Ok, what happens if you try to resume the cleaning with that command when it has been paused during a normal cleanup? Could you also check what does it send to resume from a regular pause?

pszafer:
During regular cleanup this command do nothing.

To resume from a regular cleanup it just send `app_start`, but there is some timeout I think because once it started fresh cleanup after pause...

I'm not sure how Mi Home app knows that vacuum is paused from zone clean up, but I think it is `if "in_cleaning" == 2 ` then it is zone clean up.
At least that is what I found

This is full info sent to app during paused in zone clean up:
```
ver":2,"msg_seq":6023,"state":2,"battery":100,"clean_time":9,"clean_area":122500,"error_code":0,"map_present":1,"in_cleaning":2,"in_returning":0,"in_fresh_state":0,"lab_status":1,"fan_power":60,"dnd_enabled":0}]
```

This is how it looks like when vacuum is paused:
```
ver":2,"msg_seq":6707,"state":10,"battery":99,"clean_time":9,"clean_area":122500,"error_code":0,"map_present":1,"in_cleaning":1,"in_returning":0,"in_fresh_state":0,"lab_status":1,"fan_power":60,"dnd_enabled":0}]
```
One thing, MI Home app after restart doesn't know that vacuum is resuming clean up if you press clean.

rytilahti:
Uhh, that's ugly.. So from the looks of it they reused the `in_cleaning` indeed for that, while swapping the state from "Paused" (code 10) to "Charger disconnected" (code 2).

I'm all ears if you have a proposed way to handle exposing this, my proposal:
1. Add a new, custom state code when the state is 2 and the in_cleaning is 2.
[...]

PR #473, opened 2019-02-03
Vacuum: add resume_zoned_clean() and resume_or_start() helper

rytilahti:
First take on fixing #471. resume_or_start() can be used to either unpause
or start a new cleanup depending on the state of the vacuum.


rytilahti:
I suppose this can be merged.

PR #488, opened 2019-03-24
Add easily accessible properties to DeviceError exception

syssi:


miio/exceptions.py
```diff
@@ -5,7 +5,9 @@ class DeviceException(Exception):
 
 class DeviceError(DeviceException):
     """Exception communicating an error delivered by the target device."""
-    pass
+    def __init__(self, error):
+        self.code = error.get('code')
+        self.message = error.get('message')
```
rytilahti: Please add `__str__` to provide the information (in case someone is printing the exception). Otherwise it looks good to go!

rytilahti: One question though, is this always rised with the error dict, i.e., there is no need to pass a message in like commonly done? Like `raise DeviceError("unable to do something", error)`

syssi: Does this answer your questions?

```
>>> class DeviceException(Exception):
...     """Exception wrapping any communication errors with the device."""
...     pass
... 
>>> 
>>> class DeviceError(DeviceException):
...     """Exception communicating an error delivered by the target device."""
...     def __init__(self, error):
...         self.code = error.get('code')
...         self.message = error.get('message')
... 
>>> 
>>> try:
...     error = {'code': 3000, 'message': 'adsf'}
...     raise DeviceError(error)
... except DeviceError as ex:
...     print(ex)
...     print(ex.code)
... 
{'code': 3000, 'message': 'adsf'}
3000
>>> try:
...     raise DeviceError()
... except DeviceError as ex:
...     print(ex)
...     print(ex.code)
... 
  File "<stdin>", line 2, in <module>
TypeError: __init__() missing 1 required positional argument: 'error'
>>> 
>>> 
```

rytilahti: Looks fine to me, let's keep it like this.

PR #499, opened 2019-04-14
Chuangmi Plug: Fix set_wifi_led cli command

syssi:


rytilahti:
This seems to just change the name of the variable, am I missing something here? Renaming the variable is fine for sure, but does that fix some specific, known problem?

syssi:
The method signature must align with the decorator. At the moment it's out of sync and the miiocli command won't work.

rytilahti:
Ah, sorry, that was (partially) hidden from the diff context. Looks good! :+1:

Issue #495, opened 2019-04-09
Support for pronto codes on chuangmi.remote.h102a03


nikrolls:
I have a model of the Xiaomi IR remote that reports as `chuangmi.remote.h102a03`. It works fine when capturing and replaying these captured raw commands, but any commands converted from Pronto do nothing - no flashing light, no response from the target device, but an `ok` response from the device's API.

[Here is the device on Ali Express.](https://www.aliexpress.com/item/In-Stock-Xiaomi-Mi-Universal-Smart-Remote-Controller-Home-Appliances-WIFI-IR-Switch-360-Degree-Smart/32965570092.html?spm=a2g0s.9042311.0.0.509b4c4djFoQzD)

Here are some examples:

```python
>>> ir.info()
DEBUG:miio.device:10.1.1.23:54321 >>: {'id': 3, 'method': 'miIO.info', 'params': []}
DEBUG:miio.device:10.1.1.23:54321 >>: {'id': 104, 'method': 'miIO.info', 'params': []}
DEBUG:miio.device:10.1.1.23:54321 (ts: 1970-01-30 15:52:36, id: 104) << {'result':
  {'life': 2562756, 'cfg_time': 0, 'token': 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx', 
   'mac': 'xx:xx:xx:xx:xx:xx', 'fw_ver': '1.4.3_10', 'hw_ver': 'MC200', 
   'model': 'chuangmi.remote.h102a03', 'wifi_fw_ver': 'SD878x-14.76.36.p79-702.1.0-WM', 
   'ap': {'rssi': -63, 'ssid': 'xxxxxxxxx', 'bssid': 'xx:xx:xx:xx:xx:xx'}, 
   'netif': {'localIp': '10.1.1.23', 'mask': '255.255.255.0', 'gw': '10.1.1.1', 
             'gw_mac': 'xx:xx:xx:xx:xx:xx' },
  'mmfree': 56428, 'ot': 'otu', 'otu_stat': [325, 333, 3647, 125, 3134, 10], 
  'ott_stat': [19, 4, 3, 612]}, 'id': 104}
chuangmi.remote.h102a03 v1.4.3_10 (xx:xx:xx:xx:xx:xx) @ 10.1.1.23 - token: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx


Works (previously recorded command):

[...]

PR #501, opened 2019-04-15
Support for pronto codes on chuangmi.remote.h102a03

yawor:
A separate device class for ChuangmiRemote. It subclasses ChuangmiIr and overrides `pronto_to_raw` class method to apply new learned signal encoding scheme.
It requires two new dependencies: cython and heatshrink.
There's a Python 3 compatibility issue with current version of heatshrink on PyPi, as https://github.com/johan-sports/pyheatshrink/pull/15 waits to be merged for quite some time now. Because of that I've forked the library and applied the patch from PR and I've added dependency on the fork. For the library to install automatically from the git, a `pip>=18.1` is required, as that version added support for PEP 508 dependency specification. I've added a check to setup.py to make sure at least that version is installed before adding cython and heatshrink to `install_requires` (it may still not work correctly if `python setup.py install` is used instead of pip). I couldn't find a better way until heatshrink author merges the PR.
I'm also waiting for confirmation in #495 to be sure that signals generated by the new implementation work correctly.


miio/chuangmi_ir.py
```diff
@@ -145,6 +149,18 @@ def play(self, command: str):
         return play_method(command, *command_args)
 
 
+class ChuangmiRemote(ChuangmiIr):
+
```
rytilahti: Maybe add a short docstring to explain what type of device this is, and why it requires overloading? For future readers :-)

setup.py
```diff
@@ -13,6 +14,15 @@ def readme():
         return re.sub(reg, r'\1', f.read())
 
 
+try:
+    pkg_resources.get_distribution('pip>=18.1')
+    heatshrink_requirement = [
+        'cython',
+        'heatshrink @ git+https://github.com/yawor/pyheatshrink@py3fix',
```
rytilahti: Would it make sense to add these to `extras_require`? I hope that there will be a new pypi release for heatshrink with your fix, so that we need no git paths here.

yawor: Actually that's not my fix. It was done by a different GitHub user some time ago, but the PR still waits to be merged. I've only created this fork to be 100% sure it won't suddenly disappear over night.
The extras_require may be a good option, but pip 18.1 or newer is still required to install patched version directly from git until a new version is released. Another way would be to temporarily publish a patched version to pypi under a different package name.

yawor:
There's a chance for a new Heatshrink release on pypi. The PR has been merged few weeks ago, so I've asked if they could publish the new version.

rytilahti:
Great! This fine to be merged from my side after the new release can be added into the extras_require.

rytilahti:
@yawor any updates on this?

kiwikern:
I would love to use remote.v2 with pronto codes. Is there any way we can revive this PR? I'd be happy to help

rytilahti:
Obsoleted by #1021 - thanks for the PR and groundwork on this @yawor! :-)

PR #503, opened 2019-04-24
Require click version >=7

fvollmer:
The command line interface is different for versions <7 (underscores instead of dash, see https://github.com/pallets/click/issues/1123). The documentation only documents the commands with dash. So click>=7 is required.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(
 
     install_requires=[
         'construct',
-        'click',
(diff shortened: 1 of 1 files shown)


rytilahti:
Even when I don't like pinning, I think this is fine (and forces people with older installations to upgrade the click, so that the instructions match with the commands). Thanks! :tada:

Issue #518, opened 2019-06-06
Add swing mode of a Chigo Air Conditioner


vaguecupid:
I have three xiaomi airconditioningcompanion, but one of them can't use after turning on the air conditioner. The error code is:

> 
2019-06-07 02:01:31 ERROR (MainThread) [homeassistant.components.climate] xiaomi_miio_airconditioningcompanion: Error on device update!
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 261, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 375, in async_device_update
    await self.async_update()
  File "/config/custom_components/xiaomi_miio_airconditioningcompanion/climate.py", line 276, in async_update
    ATTR_SWING_MODE: state.swing_mode.name.lower(),
  File "/usr/local/lib/python3.7/site-packages/miio/airconditioningcompanion.py", line 215, in swing_mode
    mode = int(self.state[5:6])
ValueError: invalid literal for int() with base 10: 'C'

syssi:
Do you own different air conditions? Does the problematic one support special swing modes?

vaguecupid:
> Do you own different air conditions? Does the problematic one support special swing modes?

I saw from the app, there is no difference between this and other in swing mode ,only on and off. Three air conditioners are of different brands(Panasonic,Midea,Chigo),The brand with problems is Chigo,

syssi:
If you change the swing mode does the error change, too?

vaguecupid:
yes, I turn off the swing mode,  and then restart hassio, the error code is:

> 019-06-07 03:38:00 ERROR (MainThread) [homeassistant.components.climate] xiaomi_miio_airconditioningcompanion: Error on device update!
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 261, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 375, in async_device_update
    await self.async_update()
  File "/config/custom_components/xiaomi_miio_airconditioningcompanion/climate.py", line 276, in async_update
    ATTR_SWING_MODE: state.swing_mode.name.lower(),
  File "/usr/local/lib/python3.7/site-packages/miio/airconditioningcompanion.py", line 215, in swing_mode
[...]

PR #519, opened 2019-06-06
Add particular swing mode values of a chigo air conditioner

syssi:
Closes: #518

### DIFF
--- a/miio/airconditioningcompanion.py
+++ b/miio/airconditioningcompanion.py
@@ -35,10 +35,12 @@ class FanSpeed(enum.Enum):
 
 
 class SwingMode(enum.Enum):
-    On = 0
-    Off = 1
-    Unknown2 = 2
-    Unknown7 = 7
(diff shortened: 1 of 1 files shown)


miio/airconditioningcompanion.py
```diff
@@ -212,7 +214,7 @@ def target_temperature(self) -> Optional[int]:
     def swing_mode(self) -> Optional[SwingMode]:
         """Current swing mode."""
         try:
-            mode = int(self.state[5:6])
+            mode = self.state[5:6]
```
rytilahti: Should this actually be returning a boolean instead of an enum (and exposing some extra information otherwise, if/when those unknowns get handled)? Otherwise feel free to merge as you will.

syssi: It's more or less a workaround. I'm collecting some knowledge to provide a proper solution some day.

PR #523, opened 2019-06-25
My shot at 'Vacuum: Provide human readable fan speeds (rytilahti#468)'

bugficks:
Since rytilahti#468 seems to be stalled for some time I took the freedom to give it a try.

It supports both enums and integer levels. 
The issue with V2 levels is that those 100 > levels < 105 get converted to percentage level **on device** so when queried later they return a different value.

```
# mirobo fanspeed
Current fan speed: FanSpeed.Max

# mirobo fanspeed 10
Setting fan speed to 10

# mirobo fanspeed
Current fan speed: FanSpeed.Level_10

# mirobo fanspeed 60
Setting fan speed to 60

# mirobo fanspeed
Current fan speed: FanSpeed.Balanced

# mirobo fanspeed mop
Setting fan speed to mop

# mirobo fanspeed
Current fan speed: FanSpeed.Mop

# mirobo fanspeed 105
Setting fan speed to 105

# mirobo fanspeed
Current fan speed: FanSpeed.Mop
```

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -23,6 +23,23 @@ from .vacuumcontainers import (VacuumStatus, ConsumableStatus, DNDStatus,
 _LOGGER = logging.getLogger(__name__)
 
 
+MODELS_VACUUM_V1 = ['rockrobo.vacuum.v1']
+MODELS_VACUUM_V2 = ['roborock.vacuum.s5']
+
+
+FanSpeed = enum.Enum('FanSpeed',
+    [
+        ('Quiet', 'quiet'),
+        ('Balanced', 'balanced'),
+        ('Turbo', 'turbo'),
+        ('Max', 'max'),
+        ('Mop', 'mop'),
+    ]
+    # comment if only levels above are wanted...
+    + [("Level_%d" % x, x) for x in range(0, 101)]
+)
+
+
 class VacuumException(DeviceException):
     pass
 
@@ -31,6 +48,21 @@ class TimerState(enum.Enum):
     On = "on"
     Off = "off"
 
+
+@enum.unique
+class _FanSpeedV1(enum.Enum):
+    Quiet = 38
+    Balanced = 60
+    Turbo = 77
+    Max = 90
+
+@enum.unique
+class _FanSpeedV2(enum.Enum):
+    Quiet = 101     # -> 38
+    Balanced = 102  # -> 60
+    Turbo = 103     # -> 75
+    Max = 104       # -> 100
+    Mop = 105
 
 class Consumable(enum.Enum):
     MainBrush = "main_brush_work_time"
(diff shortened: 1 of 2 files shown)


bugficks:
seems there is no interest...

miio/vacuum.py
```diff
@@ -46,6 +79,32 @@ def __init__(self, ip: str, token: str = None, start_id: int = 0,
                  debug: int = 0) -> None:
         super().__init__(ip, token, start_id, debug)
         self.manual_seqnum = -1
+        self._model = self.info().model
+        self._init_fan_speeds()
```
rytilahti: Until now the library does not do any I/O operations in constructors, so I think this should be deferred until it's really necessary. One way would be to have an internal `_fanspeedMap` function/property, which queries and stores the data during the first fetch for later uses.

miio/vacuum.py
```diff
@@ -412,19 +471,50 @@ def disable_dnd(self):
         return self.send("close_dnd_timer", [""])
 
     @command(
-        click.argument("speed", type=int),
+        click.argument("speed", type=str),
     )
-    def set_fan_speed(self, speed: int):
+    def set_fan_speed(self, speed: [int, FanSpeed]):
```
rytilahti: This could be an `Union`, but I think it is better to keep it as integer; or is there some specific need why both should be accepted?

miio/vacuum.py
```diff
@@ -412,19 +471,50 @@ def disable_dnd(self):
...
+
         return self.send("set_custom_mode", [speed])
 
     @command()
     def fan_speed(self):
         """Return fan speed."""
-        return self.send("get_custom_mode")[0]
+        speed = self.send("get_custom_mode")[0]
+        try:
+            if hasattr(self, '_fanspeedMap'):
```
rytilahti: This should be already defined when it gets this far, so no need for checking, right?

miio/vacuum.py
```diff
@@ -412,19 +471,50 @@ def disable_dnd(self):
...
+                return self._fanspeedMap[speed]
+        except KeyError:
+            pass
+
+        return speed
+
+    @command()
+    def fan_speed_list(self):
```
rytilahti: Maybe `fan_speed_presets`?

miio/vacuum_cli.py
```diff
@@ -314,17 +314,31 @@ def dnd(vac: miio.Vacuum, cmd: str,
...
 @cli.command()
-@click.argument('speed', type=int, required=False)
+@click.argument('speed', type=str, required=False)
 @pass_dev
 def fanspeed(vac: miio.Vacuum, speed):
     """Query and adjust the fan speed."""
+    from miio.vacuum import FanSpeed
     if speed:
         click.echo("Setting fan speed to %s" % speed)
+        try:
```
rytilahti: I think it makes sense first to check for the preset name, and then fall back to int conversion. Furthermore, if the conversion is not possible, and the exception should be captured and reported back.

PR #526, opened 2019-06-29
Add preliminary Roborock M1S / Mi Robot S1 support

syssi:


rytilahti:
@syssi Should we wait for more input or merge this as it is?

PR #531, opened 2019-07-17
add mi robot s1 (m1s) to discovery

rytilahti:


rytilahti:
As this has no chance of breaking something, I think it's fair to merge as it is.

PR #537, opened 2019-08-04
Add pwzn relay

SchumyHao:
Signed-off-by: SchumyHao <schumyhaojl@126.com>

### DIFF
--- a/miio/pwzn_relay.py
+++ b/miio/pwzn_relay.py
@@ -1,12 +1,11 @@
 import logging
 from collections import defaultdict
-from typing import Dict, Any, Optional
+from typing import Dict, Any
 
 import click
 
 from .click_common import command, format_output
 from .device import Device
-from .utils import deprecated
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -293,6 +292,7 @@ class PwznRelayStatus:
 
     def __json__(self):
         return self.data
+
 
 class PwznRelay(Device):
     """Main class representing the PWZN Relay."""
--- a/miio/pwzn_relay.py
+++ b/miio/pwzn_relay.py
@@ -1,0 +1,380 @@
+import logging
+from collections import defaultdict
+from typing import Dict, Any, Optional
+
+import click
+
+from .click_common import command, format_output
+from .device import Device
+from .utils import deprecated
+
+_LOGGER = logging.getLogger(__name__)
+
+MODEL_PWZN_RELAY_APPLE = 'pwzn.relay.apple'
+
+AVAILABLE_PROPERTIES = {
+    MODEL_PWZN_RELAY_APPLE: [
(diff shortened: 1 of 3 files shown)


miio/pwzn_relay.py
```diff
@@ -0,0 +1,380 @@
...
+        """
+        self.data = data
+
+    def relay(self, num: int) -> bool:
+        """Current relay X state."""
+        if "relay_status" in self.data:
+            return True if self.data["relay_status"] & (1 << num) else False
+
+    @property
+    def relay0(self) -> bool:
```
rytilahti: I think it would be better to let the users of the library to call the `relay` (maybe rename to `relay_state` given it returns whether the given relay is on?).

miio/pwzn_relay.py
```diff
@@ -0,0 +1,380 @@
...
+        """Current relay14 state."""
+        return self.relay(14)
+
+    @property
+    def relay15(self) -> bool:
+        """Current relay15 state."""
+        return self.relay(15)
+
+    @property
+    def name0(self) -> str:
```
rytilahti: Instead of having separate methods to return single names, I'd prefer to have `relay_names` or simply `names` returning a dict of (index, name), something like this:
```
def relay_names() -> Dict[int, str]:
    def _extract_index_from_key(name) -> int:
        """extract the index from the variable"""

    return {_extract_index(name): value for name, value in self.data.items() if name.startswith("name")}
```

SchumyHao:
Hi @rytilahti 

There are two device models `pwzn.relay.apple` `pwzn.relay.banana` 
Device is like this 
image

miio/pwzn_relay.py
syssi: I would be happy about this solution, too.

SchumyHao: I'm not so familar with python language.
Can you show me more detail about how to fix this?

syssi: Remove all nameX methods/properties and implement a single method as suggested by @rytilahti. This method should return a Dict with relay names. Do you need additional support to implement this?

SchumyHao:
I update code as @rytilahti  requied

rytilahti:
Looks good, thanks! Can you update README's supported devices section? Also, are these relays discoverable? If so, please adjust the discovery.py accordingly, and this can then be merged!

SchumyHao:
Hi @rytilahti 

Sorry for late response. I'm busy with my work those days.

This device can not discovered token is not accessable.
So I only update README.

SchumyHao:
Any update about this PR?
I checked the CI error logs, it seems not caused by my patches

rytilahti:
Indeed, the errors are odd, do you have an idea @syssi? This can be merged now though, thanks @SchumyHao! :tada:

PR #539, opened 2019-08-09
Add support for Air Quality Monitor S1 (cgllc.airmonitor.s1)

zhumuht:
already test pass.
device details: https://www.gearbest.com/weather-station/pp_009986206287.html

```
Got new state:  <AirQualityMonitorStatus power=None, aqi=None, battery=100, usb_power=False, temperature=25.8, humidity=59, co2=540, pm25=11.8, tvoc=243, display_clock=False>
```

if merge on python_miio 0.4.6, I will update xiaomi_miio sensor on home assistant to support AirQualityMonitor:cgllc.airmonitor.s1

image

### DIFF
--- a/miio/airqualitymonitor.py
+++ b/miio/airqualitymonitor.py
@@ -10,6 +10,7 @@ _LOGGER = logging.getLogger(__name__)
 
 MODEL_AIRQUALITYMONITOR_V1 = 'zhimi.airmonitor.v1'
 MODEL_AIRQUALITYMONITOR_B1 = 'cgllc.airmonitor.b1'
+MODEL_AIRQUALITYMONITOR_S1 = 'cgllc.airmonitor.s1'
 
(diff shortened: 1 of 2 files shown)


miio/airqualitymonitor.py
```diff
@@ -92,16 +102,46 @@ def sensor_state(self) -> str:
         """Sensor state."""
         return self.data["sensor_state"]
 
+    @property
+    def co2(self) -> int:
+        return self.data["co2"]
```
rytilahti: Please add docstrings to the new properties. Also, in case these methods do not work on the other devices, maybe consider rising an exception (or maybe even having a separate status class inheriting from the common one (if there are common elements)?

syssi:
@zhumuht I've updated the official HA component. Do you like to test the implementation? 

https://github.com/home-assistant/home-assistant/pull/27255

PR #544, opened 2019-09-02
Add tinymu smart toiletlid

scp10011:
#  select toiletlid status
    Work: False
    State: 1
    Ambient Light: White
    Filter remaining: 100%
    Filter remaining time: 180
    
# set toiletlid ambient light
    Support color:
    blue
    green
    orange
    powder
    purple
    red
    white
    yellow

# nozzle clean

### DIFF
--- a/miio/tests/test_toiletlid.py
+++ b/miio/tests/test_toiletlid.py
@@ -1,0 +1,83 @@
+from unittest import TestCase
+
+import pytest
+
+from miio.toiletlid import (
+    Toiletlid,
+    ToiletlidStatus,
+    AmbientLightColor,
+    MODEL_TOILETLID_V1,
+)
+from .dummies import DummyDevice
+
+
+class DummyToiletlidV1(DummyDevice, Toiletlid):
+    def __init__(self, *args, **kwargs):
+        self.model = MODEL_TOILETLID_V1
+        self.state = {
+            "is_on": False,
+            "work_state": 1,
+            "ambient_light": "Yellow",
+            "filter_use_flux": "100",
+            "filter_use_time": "180",
+        }
+
+        self.return_values = {
+            "get_prop": self._get_state,
+            "nozzle_clean": lambda x: self._set_state("work_state", [97]),
+            "set_aled_v_of_uid": self.set_aled_v_of_uid,
+            "get_aled_v_of_uid": self.get_aled_v_of_uid,
+        }
+        super().__init__(args, kwargs)
+
+    def set_aled_v_of_uid(self, x):
+        uid, color = x
+        return self._set_state("ambient_light", [AmbientLightColor(color).name])
+
+    def get_aled_v_of_uid(self, uid):
+        color = self._get_state(["ambient_light"])
+        if not AmbientLightColor._member_map_.get(color[0]):
+            raise ValueError(color)
+        return AmbientLightColor._member_map_.get(color[0]).value
+
+
+@pytest.fixture(scope="class")
+def toiletlidv1(request):
+    request.cls.device = DummyToiletlidV1()
+    # TODO add ability to test on a real device
+
+
+@pytest.mark.usefixtures("toiletlidv1")
+class TestToiletlidV1(TestCase):
+    def is_on(self):
+        return self.device.status().is_on
+
+    def state(self):
+        return self.device.status()
+
+    def test_status(self):
+        self.device._reset_state()
+
+        assert repr(self.state()) == repr(ToiletlidStatus(self.device.start_state))
+
+        assert self.is_on() is False
+        assert self.state().work_state == self.device.start_state["work_state"]
+        assert self.state().ambient_light == self.device.start_state["ambient_light"]
+        assert (
+                self.state().filter_use_percentage
+                == "%s%%" % self.device.start_state["filter_use_flux"]
+        )
+        assert (
+                self.state().filter_remaining_time
+                == self.device.start_state["filter_use_time"]
+        )
+
+    def test_set_ambient_light(self):
+        for value, enum in AmbientLightColor._member_map_.items():
+            self.device.set_ambient_light(enum)
+            assert self.device.status().ambient_light == value
+
+    def test_nozzle_clean(self):
+        self.device.nozzle_clean()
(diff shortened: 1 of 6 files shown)


miio/toiletlid.py
```diff
@@ -0,0 +1,147 @@
...
+    Orange = "6"
+    Red = "7"
+
+
+class ToiletlidStatus:
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
+
+    @property
+    def work_state(self) -> int:
```
rytilahti: Do you have idea about potential state code? I.e. could this be turned into an enum?

miio/toiletlid.py
```diff
@@ -0,0 +1,147 @@
...
+        """Device state code"""
+        return self.data["work_state"]
+
+    @property
+    def is_on(self) -> bool:
+        return self.work
+
+    @property
+    def work(self) -> bool:
```
rytilahti: Is there a need for this property, considering there's already a more generic `is_on`?

miio/toiletlid.py
```diff
@@ -0,0 +1,147 @@
...
+        debug: int = 0,
+        lazy_discover: bool = True,
+        model: str = MODEL_TOILETLID_V1,
+    ) -> None:
+        super().__init__(ip, token, start_id, debug, lazy_discover)
+
+        if model in AVAILABLE_PROPERTIES:
+            self.model = model
+        else:
+            self.model = MODEL_TOILETLID_V1
```
rytilahti: Do you think there'll be more models than this one? Because if not, the code could be made much simpler by aoiding the model handling altogether.

miio/toiletlid.py
```diff
@@ -0,0 +1,147 @@
...
+            "State: {result.work_state}\n"
+            "Ambient Light: {result.ambient_light}\n"
+            "Filter remaining: {result.filter_use_percentage}\n"
+            "Filter remaining time: {result.filter_remaining_time}\n",
+        )
+    )
+    def status(self) -> ToiletlidStatus:
+        """Retrieve properties."""
+        properties = AVAILABLE_PROPERTIES[self.model]
+        _props_per_request = 15
```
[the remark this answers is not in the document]
rytilahti: Do you mind fixing this?

miio/toiletlid.py
```diff
@@ -0,0 +1,147 @@
...
+                properties_count,
+                values_count,
+            )
+        color = self.get_ambient_light()
+        return ToiletlidStatus(dict(zip(properties, values), ambient_light=color))
+
+    @command(default_output=format_output("Nozzle clean"))
+    def nozzle_clean(self):
+        """Nozzle clean."""
+        return self.send("nozzle_clean", ["on"])
```
rytilahti: Should there be an "off" variant of this?

miio/toiletlid.py
```diff
@@ -0,0 +1,147 @@
...
+        return self.send("set_aled_v_of_uid", ["", color.value])
+
+    @command(default_output=format_output("Get the Ambient light color."))
+    def get_ambient_light(self) -> str:
+        """Set Ambient light color."""
+        color = self.send("get_aled_v_of_uid", [""])
+        if color:
+            return AmbientLightColor(color[0]).name
+        else:
+            return AmbientLightColor("0").name
```
rytilahti: I think this else should be removed. If the device does not respond, it's better to error out.

miio/toiletlid.py
scp10011: This operation cannot be terminated

miio/toiletlid.py
scp10011: There are two brands and three models with similar functions. They are `Tiny mu smart toiletdled` `Tiny mu smart toiletdled voice AI version` `little whale wash smart toilet pro`

miio/toiletlid.py
scp10011: The merchant does not have a development document, the state value is a black box.

miio/toiletlid.py
```diff
@@ -27,6 +27,10 @@ class AmbientLightColor(enum.Enum):
     Red = "7"
 
 
+class AmbientLightException(DeviceException):
```
rytilahti: This does not need to be so specific, please name it `ToiledLidException`.

miio/toiletlid.py
```diff
@@ -139,9 +136,13 @@ def set_ambient_light(self, color: AmbientLightColor):
...
+        try:
             return AmbientLightColor(color[0]).name
-        else:
-            return AmbientLightColor("0").name
+        except TypeError:
+            raise DeviceException(
+                "Get Ambient Light Operating Device Not Responding!"
+            ) from None
+        except ValueError as e:
+            raise AmbientLightException(e) from None
```
rytilahti: Is there really need to handle this case (and to that end, is it even necessary to have a custom exception for this)? How often can this happen? I think it would make more sense to check for the value of `color[0]` and raise an exception on unexpected value if that's necessary at all.

miio/toiletlid.py
rytilahti: Ok, thanks for being farsighted! :+1:

miio/toiletlid.py
rytilahti: Fair enough.

rytilahti:
Looks good to me now, thanks! :+1: One small thing, please add the full name of the device to README.md's list of supported devices.

@syssi, you are more knowledgeable on props based integrations, ok to merge?

miio/tests/test_toiletlid.py
```diff
@@ -0,0 +1,94 @@
...
+"""
+Response instance
+>> status
+
+Work: False
+State: 1
+Ambient Light: Yellow
+Filter remaining: 100%
+Filter remaining time: 180
+"""
```
rytilahti: Sorry for being unclear, I meant with my comment about the response for `get_prop` inside (documented under the state container, see e.g. https://github.com/rytilahti/python-miio/blob/master/miio/ceil.py#L20). Other than that this is ready to be merged :+1:

rytilahti:
Thanks :tada:

PR #562, opened 2019-10-13
Add support for the cgllc.airmonitor.b1

fwestenberg:
Added support for cgllc.airmonitor.b1

image

### DIFF
--- a/miio/airqualitymonitor.py
+++ b/miio/airqualitymonitor.py
@@ -239,29 +239,16 @@ class AirQualityMonitor(Device):
     def status(self) -> AirQualityMonitorStatus:
         """Return device status."""
 
-        if info.model == None:
+        if info.model is None:
             info = self.info()
             self.model = info.model
 
         properties = AVAILABLE_PROPERTIES[self.model]
 
-        if self.model != "cgllc.airmonitor.b1":
-            values = self.send("get_prop", properties)
-        else:
+        if self.model == MODEL_AIRQUALITYMONITOR_B1:
             values = self.send("get_air_data")
-            """
-            Format like on the display of the unit
-            In the current situation tvoc_unit and temperature_unit does not translate the tvoc/temperature values
-            """
-            values['co2']              = round(values['co2e'],1)
-            # Replace key co2e by just co2
-            values.pop('co2e', None)
-            values['humidity']         = round(values['humidity'],1)
-            values['pm25']             = round(values['pm25'],1)
-            values['temperature']      = round(values['temperature'],1)
-            values.pop('temperature_unit', None)
-            values['tvoc']             = round(values['tvoc'],3)
-            values.pop('tvoc_unit', None)
+        else:
+            values = self.send("get_prop", properties)
 
         properties_count = len(properties)
         values_count = len(values)
--- a/miio/airqualitymonitor.py
+++ b/miio/airqualitymonitor.py
@@ -70,10 +70,7 @@ class AirQualityMonitorStatus:
     @property
     def power(self) -> Optional[str]:
         """Current power state."""
-        if "power" in self.data and self.data["power"] is not None:
-            return self.data["power"]
-
-        return None
+        return self.data.get("power", None)
 
     @property
     def is_on(self) -> bool:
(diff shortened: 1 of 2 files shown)


miio/airqualitymonitor.py
```diff
@@ -200,17 +211,19 @@ def __init__(
         start_id: int = 0,
         debug: int = 0,
         lazy_discover: bool = True,
-        model: str = MODEL_AIRQUALITYMONITOR_V1,
+        model: str = None,
```
syssi: Please don't change the default here.

miio/airqualitymonitor.py
```diff
@@ -200,17 +211,19 @@ def __init__(
...
     ) -> None:
         super().__init__(ip, token, start_id, debug, lazy_discover)
 
+        if model == None:
```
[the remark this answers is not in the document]
syssi: Please don't add I/O at the constructor. The correct model should be passed to the constructor if needed.

miio/airqualitymonitor.py
```diff
@@ -232,7 +245,23 @@ def status(self) -> AirQualityMonitorStatus:
 
         properties = AVAILABLE_PROPERTIES[self.model]
 
-        values = self.send("get_prop", properties)
+        if self.model != "cgllc.airmonitor.b1":
```
syssi: Please use the const here and invert the condition:

```
if self.model == MODEL_AIRQUALITYMONITOR_B1:
    values = self.send("get_air_data")
else:
```

miio/airqualitymonitor.py
```diff
@@ -232,7 +245,23 @@ def status(self) -> AirQualityMonitorStatus:
...
+        else:
+            values = self.send("get_air_data")
+            """
+            Format like on the display of the unit
+            In the current situation tvoc_unit and temperature_unit does not translate the tvoc/temperature values
+            """
+            values['co2']              = round(values['co2e'],1)
+            # Replace key co2e by just co2
+            values.pop('co2e', None)
+            values['humidity']         = round(values['humidity'],1)
```
syssi: Please don't manipulate the measurements. Pass all measurements unchanged to a new status class (cp. FanStatusP5 of fan.py).

miio/airqualitymonitor.py
```diff
@@ -244,7 +273,7 @@ def status(self) -> AirQualityMonitorStatus:
                 values_count,
             )
 
-        if self.model == MODEL_AIRQUALITYMONITOR_S1:
+        if self.model == MODEL_AIRQUALITYMONITOR_S1 or self.model == MODEL_AIRQUALITYMONITOR_B1:
```
syssi: return the new status class here

miio/airqualitymonitor.py
```diff
@@ -206,11 +241,12 @@ def __init__(
 
         if model in AVAILABLE_PROPERTIES:
             self.model = model
-        else:
+        elif model is not None:
             self.model = MODEL_AIRQUALITYMONITOR_V1
```
fwestenberg: When a model is provided, but not in available_properties fallback is still there.

miio/airqualitymonitor.py
```diff
@@ -230,9 +267,20 @@ def __init__(
     def status(self) -> AirQualityMonitorStatus:
         """Return device status."""
 
+        if self.model is None:
+            """Autodetection"""
+            info = self.info()
+            self.model = info.model
+            self.mac_address = info.mac_address
+            self.firmware_version = info.firmware_version
+            self.hardware_version = info.hardware_version
```
rytilahti: I would avoid exposing the mac, and the versions like this. Prefer `self._firmware_version` etc for internal use.

fwestenberg:
Thanks for your suggestions! Please let me know if anything else should be changed.

miio/airqualitymonitor.py
```diff
@@ -230,9 +222,20 @@ def __init__(
...
         """Return device status."""
 
-        properties = AVAILABLE_PROPERTIES[self.model]
+        if self._model is None:
+            """Autodetection"""
+            info = self.info()
+            self._model = info.model
+            self._mac_address = info.mac_address
+            self._firmware_version = info.firmware_version
+            self._hardware_version = info.hardware_version
```
rytilahti: Where are you planning to use these? To me it looks like they are not used at all, so it'd be better simply to remove the last three lines, no?

If it's for homeassistant integration, it should simply query the info itself and parse the information out from it (instead of accessing these now-private variables) :-)

fwestenberg: It is instead for the HA integration. But I don't like the idea of querying the info again. Also I don't see any trouble in saving the info on device-level. So now the remaining question to me is: where to store them? I suppose all of the Xiaomi Air Quality devices have the same response here?

fwestenberg: How about creating the following in AirQualityMonitor (for mac_address, model, firmware_version and hardware_version:

`    @property `
`    def mac_address(self) -> Optional[int]: `
`        """Return mac address.""" `
`        return self._mac_address `

rytilahti: I'm thinking if it would make simply sense to extend Device's `info()` to cache the result (and add a kwarg `refresh` if there's a need to force re-fetching), as I assume the information returned is quite static (see https://github.com/rytilahti/python-miio/blob/master/miio/device.py#L33).

But that should be done in a separate PR, for the time being I would do the caching downstream (as you already do in the homeassistant PR) and we can think about adapting the behavior of `info()` (https://github.com/rytilahti/python-miio/blob/master/miio/device.py#L348) to cache the results for some time period or until explicitly asked for a refresh.

@syssi do you have an opinion on this?

fwestenberg: First it started with getting the device model, which is needed in case of te B1 to simply get te status. That's why I added the info( ) call in the first place to the init, and then moved it to the status method. Here the info( ) is only called once, when model is None.

rytilahti: Yes, I got that, but what I'm trying to say is that those internal (info) variables should not be exposed like this only for a single device type. There is already DeviceInfo container holding this information, so the homeassistant part should simply call info (once) and save those details for its own use.

fwestenberg: I reversed some changes (model, firmware etc). All should be fine now (I hope).

rytilahti: Looks good, thanks! Let's get it merged! :tada:

miio/tests/test_airqualitymonitor.py
```diff
@@ -128,3 +129,55 @@ def test_status(self):
...
+        assert self.state().aqi is None
+        assert self.state().temperature == self.device.start_state["temperature"]
+        assert self.state().humidity == self.device.start_state["humidity"]
+        assert self.state().co2 == None
+        assert self.state().co2e == self.device.start_state["co2e"]
+        assert self.state().pm25 == self.device.start_state["pm25"]
+        assert self.state().tvoc == self.device.start_state["tvoc"]
+        assert self.state().display_clock is None
+        assert self.state().night_mode is None
+
```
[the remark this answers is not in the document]
rytilahti: Could you still fix the hound warnings instead of marking them resolved?

PR #569, opened 2019-10-31
Remove UTF-8 encoding comment

quamilek:


rytilahti:
Thanks, these are not indeed needed when using python3 :+1:

PR #575, opened 2019-11-11
Add check-manifest, check for pypi-description, flake8-docstrings

rytilahti:
Also, add already shipped docs/yeelight.rst that wasn't in the repo


rytilahti:
Considering there's no objections, I'm going to merge this and fix later if needed.

PR #579, opened 2019-11-22
Improve toiletlid various parameters

scp10011:
1.add work_state analysis
2.add xiaomi_band interface

### DIFF
--- a/miio/toiletlid.py
+++ b/miio/toiletlid.py
@@ -1,6 +1,6 @@
 import enum
 import logging
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 import click
 
@@ -27,6 +27,14 @@ class AmbientLightColor(enum.Enum):
     Red = "7"
 
 
+class ToiletlidOperatingMode(enum.Enum):
+    Vacant = 0
+    Occupied = 1
+    RearCleanse = 2
+    FrontCleanse = 3
+    NozzleClean = 6
+
+
 class ToiletlidStatus:
     def __init__(self, data: Dict[str, Any]) -> None:
         # {"work_state": 1,"filter_use_flux": 100,"filter_use_time": 180, "ambient_light": "Red"}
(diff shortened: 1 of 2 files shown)


miio/toiletlid.py
```diff
@@ -37,6 +45,11 @@ def work_state(self) -> int:
         """Device state code"""
         return self.data["work_state"]
 
+    @property
+    def work_mode(self) -> str:
+        """Device working mode"""
+        return ToiletlidOperatingMode((self.work_state - 1) // 16).name
```
rytilahti: Should this be returning rather the `ToiletLidOperatingMode` instead of t he string? Also, should the work_state be adapted already in the `__init__`?

miio/toiletlid.py
```diff
@@ -123,22 +139,52 @@ def nozzle_clean(self):
 
     @command(
         click.argument("color", type=EnumType(AmbientLightColor, False)),
+        click.argument("xiaomi_id", type=str, default=""),
         default_output=format_output(
             "Set the ambient light to {color} color the next time you start it."
         ),
     )
-    def set_ambient_light(self, color: AmbientLightColor):
+    def set_ambient_light(self, color: AmbientLightColor, xiaomi_id: str = ""):
```
rytilahti: This one and the `bind_xiaomi_band` have a different position for the `xiaomi_id`. Should it be both cases the first argument? Or should it be an option defaulting to `""`, in case that is what works for those who have no `xiaomi_id`?

scp10011: They are the same parameter. `set_ambient_light` if you do not select `xiaomi_id` then set as the default user `""`. But `xiaomi_id` is a required parameter in `bind_xiaomi_band`.

rytilahti: Fair enough, let's merge this, it can be improved later on if necessary. Thanks again! :100:

Issue #550, opened 2019-09-20
Support for new vaccum Xiaomi Mijia STYJ02YM 


kwetnico:
Hello,

It's possible to add the new vaccum xaomi.
This model : https://fr.gearbest.com/vacuum-cleaners/pp_009487950353.html

Thanks

ABRT29:
hi, please mee too

ABRT29:
Model number : STYJ02YM
![Capture d’écran 2019-10-01 à 09 17 11](https://user-images.githubusercontent.com/16398440/65941900-6bd00680-e42c-11e9-8eec-2a2a96091b2d.png)

with mi home 5.4.54 : 

{"did":"27XXX6986","token":"72676b4a764bXXXXXXXXXXXXXX5363","longitude":"XXX","latitude":"XXX","**name":"Mi Robot Vacuum-Mop P"**,"pid":"0","localip":"192.168.0.35","mac":"XX:XX:XX:XX:XX:XX","ssid":"XX","bssid":"XX:XX:XX:XX:XX:XX","parent_id":"","parent_model":"","show_mode":1,"**model":"viomi.vacuum.v7",**"adminFlag":1,"shareFlag":0,"permitLevel":16,"isOnline":true,"desc":"Appareil en ligne ","extra":{"isSetPincode":0,"fw_version":"3.5.3_0042","needVerifyCode":0,"isPasswordEncrypt":0}

alex7614:
Hi
Please too

HydrelioxGitHub:
I have this vacuum too. Is there anything we can do for help ?

NSDmitry:
Hi, please too

ABRT29:
have you any news ?

edubox:
> Model number : STYJ02YM
> ![Capture d’écran 2019-10-01 à 09 17 11](https://user-images.githubusercontent.com/16398440/65941900-6bd00680-e42c-11e9-8eec-2a2a96091b2d.png)
> 
> with mi home 5.4.54 :
> 
[...]

PR #581, opened 2019-11-27
Add support for a few more commands of the STYJ02YM

rumpeltux:
Tested these with a local device.

#550


rytilahti:
Thanks a lot for the PR, looking good! :tada:

Issue #533, opened 2019-07-25
Support for Mi Smart Humidifier (deerma.humidifier.mjjsq)


rezmus:
it's new mi humidifier, miio model deerma.humidifier.mjjsq

https://www.xiaomiyoupin.com/detail?gid=109486
https://ibb.co/n1dwCgS

while i don't have device (yet?) i took a look at the plugin so i leave info here if anybody wants to add support. below you can find all props/methods supported by firmware.

{"method":"get_prop","params":["Humidifier_Gear","Humidity_Value","HumiSet_Value","Led_State","OnOff_State","TemperatureValue","TipSound_State","waterstatus","watertankstatus"]}

props

Humidifier_Gear - speed (1-4, 1 Low, 2 Medium, 3 High, 4 Humidity)
Humidity_Value - humidity value (0-100)
HumiSet_Value - desired humidity value? (not used by plugin atm)
Led_State - led state (0 - off, 1 - on)
OnOff_State - power state (0 - off, 1 - on)
TemperatureValue - temperature value
TipSound_State - sound state (0 - off, 1 - on)
waterstatus - water status (0 - insufficient, 1 - normal)
watertankstatus - water tank status (0 - detached, 1 - installed)

methods

{"method":"Set_HumidifierGears","params":[1-4]} - set speed (Humidifier_Gear prop)
{"method":"Set_OnOff","params":[0/1]} - set power off/on (OnOff_State prop)
{"method":"SetLedState","params":[0/1]} - set led off/on (Led_State prop)
{"method":"SetTipSound_Status","params":[0/1]} - set sound off/on (TipSound_State prop)
{"method":"Set_HumiValue","params":[0-100]} - set desired humidity value? (not used by plugin atm)

syssi:
@rezmus Good job! This should be enough to support the device.

dovchinnikov:
@rezmus hi, could you please teach how "to look at the plugin" to obtain properties and methods?

PR #586, opened 2019-12-03
Add support for deerma.humidifier.mjjsq

syssi:
Fixes: #533


miio/tests/test_airhumidifier.py
```diff
@@ -3,14 +3,17 @@
 import pytest
 
 from miio import AirHumidifier
+from miio import AirHumidifierMjjsq
```
rytilahti: Looks good to me, but I'm wondering if this should be in a separate file (considering it's a different product line)? Otherwise feel free to merge :tada:

syssi: I will move it to a separate file.

miio/airhumidifier_mjjsq.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class OperationModeMjjsq(enum.Enum):
+    Low = 1
+    Medium = 2
+    High = 3
+    Humidity = 4
+
+
+class AirHumidifierMjjsqStatus:
```
rytilahti: How about removing Mjssq from class names now that they are not clashing? The downstreams can do similarly what homeassistant does and import them as whatever they prefer.

miio/airhumidifier_mjjsq.py
```diff
@@ -0,0 +1,236 @@
...
+         'Led_State': 1, 'OnOff_State': 0, 'TemperatureValue': 21,
+         'TipSound_State': 1, 'waterstatus': 1, 'watertankstatus': 1}
+        """
+
+        self.data = data
+
+    @property
+    def power(self) -> str:
+        """Power state."""
+        return "on" if self.data["OnOff_State"] == 1 else "off"
```
rytilahti: Is there a need to have both `power` and `is_on`? I'd prefer simply a single boolean variant of one of these.

miio/airhumidifier_mjjsq.py
```diff
@@ -0,0 +1,236 @@
...
+        return self.power == "on"
+
+    @property
+    def mode(self) -> OperationModeMjjsq:
+        """Operation mode. Can be either low, medium, high or humidity."""
+        return OperationModeMjjsq(self.data["Humidifier_Gear"])
+
+    @property
+    def temperature(self) -> int:
+        """Current temperature."""
```
rytilahti: `Current temperature in celsius.` maybe?

miio/airhumidifier_mjjsq.py
```diff
@@ -0,0 +1,236 @@
...
+        return self.data["TemperatureValue"]
+
+    @property
+    def humidity(self) -> int:
+        """Current humidity."""
```
rytilahti: What's the unit here?

miio/airhumidifier_mjjsq.py
```diff
@@ -0,0 +1,236 @@
...
+        return self.data["TipSound_State"] == 1
+
+    @property
+    def led(self) -> bool:
+        """True if LED is turned on."""
+        return self.data["Led_State"] == 1
+
+    @property
+    def target_humidity(self) -> int:
+        """Target humiditiy."""
```
rytilahti: Unit missing.

miio/airhumidifier_mjjsq.py
syssi: I would like to stick on this because of consistency and a nice status output of: "Power: on"

rytilahti: Ok, your call.

PR #576, opened 2019-11-12
Add support for Xiaomi Mijia STYJ02YM (viomi.vacuum.v7)

rytilahti:
Based on findings done by @nqkdev and contributions by @rumpeltux - thanks a lot for help!
This is the initial PR to fix #550 - plenty of values are still unknown so any help is welcome!

PR #590, opened 2019-12-05
STYJ02YM: Manual movement and mop mode support

rumpeltux:
#576

### DIFF
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -156,7 +156,8 @@ class ViomiVacuumStatus:
         return bool(self.data["has_newmap"])
 
     @property
-    def is_mop(self) -> ViomiMopMode:
+    def mop_mode(self) -> ViomiMopMode:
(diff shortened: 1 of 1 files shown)


miio/viomivacuum.py
```diff
@@ -140,9 +156,8 @@ def has_new_map(self) -> bool:
         return bool(self.data["has_newmap"])
 
     @property
-    def is_mop(self) -> bool:
-        """True if mopping."""
-        return bool(self.data["is_mop"])
+    def is_mop(self) -> ViomiMopMode:
+        return ViomiMopMode(self.data["is_mop"])
```
rytilahti: Should this be called `mop_mode` instead of `is_mop`? Also, please add a docstring for the property, too.

rumpeltux: Done.

rytilahti:
Thanks, merging! :tada:

PR #585, opened 2019-12-01
Add basic support for Xiaomi Mi Air Purifier 3/3H

petrkotek:
Add support for basic operations for new Xiaomi Air Purifier 3/3H (`zhimi-airpurifier-mb3` and `zhimi-airpurifier-ma4` (https://github.com/rytilahti/python-miio/issues/577)).

In order to support that, also implement `MiotDevice` class with basic support for MIoT protocol which the `AirPurifierMiot` extends to retrieve & set properties.

Mapping is based on https://github.com/rytilahti/python-miio/issues/543#issuecomment-544745035 - thanks, it helped a lot!

Note 
- I have not added any tests yet. I'd appreciate first round of a review before I do so.
- I have not tried the template/generative approach. Thought it would be better just to get it working. I can attempt to do that separately later.

Disclaimer: this is my first contribution to this project & I don't have Python experience.

ToDo:
- [x] extract Filter Type detection to a separate class
- [ ] add test coverage

Manually tested operations:
- `info` (see output below)
- `status` (see output below)
- `off`
- `on`
- `set_buzzer <true|false>`
- `set_child_lock <true|false>`
- `set_fan_level <1-3>`
- `set_favorite_level <0-14>`
- `set_led <true|false>`
- `set_led_brightness <off|dim|bright>`
- `set_mode <auto|fan|favorite|silent>`

Example output of `info`:
```
$ python miio/cli.py airpurifiermiot --ip $AIR_IP --token $TOKEN info
Model: zhimi.airpurifier.mb3
Hardware version: esp32
Firmware version: 2.0.5
Network: {'localIp': '192.168.0.22', 'mask': '255.255.255.0', 'gw': '192.168.0.1'}
AP: {'ssid': 'coolwifi', 'bssid': '20:E5:2A:3C:BC:82', 'rssi': -59, 'primary': 8}
```

Example output of `status`:
```
$ python miio/cli.py airpurifiermiot --ip $AIR_IP --token $TOKEN status
Power: on
AQI: 3 μg/m³
Average AQI: 6 μg/m³
Humidity: 58 %
Temperature: 25.799999 °C
Mode: OperationMode.Favorite
LED: True
LED brightness: LedBrightness.Dim
Buzzer: False
Child lock: False
Favorite level: 14
Filter life remaining: 98 %
Filter hours used: 59
Use time: 212700 s
Purify volume: 8606 m³
Motor speed: 2148 rpm
Filter RFID product id: 0:0:31:31
[...]

PR #592, opened 2019-12-08
Refactoring: Split Device class into Device+Protocol

petrkotek:
Before adding support for the first MIoT device (https://github.com/rytilahti/python-miio/pull/585) I thought it would be good to do a little bit of refactoring (i.e. to make the https://github.com/rytilahti/python-miio/pull/585 smaller).

This PR extracts command-sending functionality from Device class to new Protocol class.

- Basically the `send()`, `discover()` and `do_discover()` methods were moved from `Device` to `Protocol`
- `Protocol` is instantiated in `Device` constructor.
- `Protocol#send()` is deprecated & forwards the commands to CommandSender.

Note: tried discovery and looks to be working fine:
```
$ python miio/vacuum_cli.py discover
INFO:miio.discovery:Discovering devices with mDNS, press any key to quit...
INFO:miio.discovery:Found a supported 'Vacuum' at 192.168.0.10 - token: ffffffffffffffffffffffffffffffff
/Users/petrkotek/Workspace/python-miio/miio/discovery.py:177: UserWarning: Please consider using python-yeelight for more complete support.
  dev = device_cls(ip=addr)
INFO:miio.discovery:Found a supported 'Yeelight' at 192.168.0.17 - token: 00000000000000000000000000000000
INFO:miio.discovery:Found a supported 'AirQualityMonitor' at 192.168.0.2 - token: bb835d94f8e75c74a231e445c9716616
INFO:miio.discovery:Found a supported 'Yeelight' at 192.168.0.9 - token: 00000000000000000000000000000000
```

Edit: renames `CommandSender` to `Protocol`


petrkotek:
@rytilahti Looking forward to hear your feedback when you have a bit of time to review this. Thank you!

petrkotek:
@rytilahti friendly ping in case you missed it. If you're busy, that's ok :)

miio/command_sender.py
```diff
@@ -0,0 +1,235 @@
...
+
+import construct
+
+from .exceptions import DeviceError, DeviceException, RecoverableError
+from .protocol import Message
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class CommandSender:
```
rytilahti: Rename this to `Protocol`.

miio/device.py
```diff
@@ -122,205 +117,11 @@ def __init__(
...
-                    )
-                    seen_addrs.append(addr[0])
-            except socket.timeout:
-                if is_broadcast:
-                    _LOGGER.info("Discovery done")
-                return  # ignore timeouts on discover
-            except Exception as ex:
-                _LOGGER.warning("error while reading discover results: %s", ex)
-                break
+        self.command_sender = CommandSender(ip, token, start_id, debug, lazy_discover)
```
rytilahti: `self.protocol`.

miio/device.py
```diff
@@ -122,205 +117,11 @@ def __init__(
...
-        except RecoverableError as ex:
-            if retry_count > 0:
-                _LOGGER.debug(
-                    "Retrying to send failed command, retries left: %s", retry_count
-                )
-                return self.send(command, parameters, retry_count - 1)
-
-            _LOGGER.error("Got error when receiving: %s", ex)
-            raise DeviceException("Unable to recover failed command") from ex
+        return self.command_sender.send(command, parameters, retry_count)
```
rytilahti: No need to deprecate this, simply keep the docstring and make this pass the parameters to the implementation.

miio/discovery.py
```diff
@@ -207,7 +207,7 @@ def check_and_create_device(self, info, addr) -> Optional[Device]:
                     _LOGGER.info(
                         "%s: token: %s",
                         v(info),
-                        pretty_token(dev.do_discover().checksum),
+                        pretty_token(dev.command_sender.do_discover().checksum),
```
rytilahti: I think we should move `do_discover` out from the protocol to be a separate `handshake()` or `hello()` for the device. The first step could be to create a wrapper as with `send()` for this, though.

miio/tests/test_airconditioningcompanion.py
```diff
@@ -56,21 +57,19 @@ def __init__(self, *args, **kwargs):
...
+            {
+                "get_model_and_state": self._get_state,
+                "start_ir_learn": lambda x: True,
+                "end_ir_learn": lambda x: True,
+                "get_ir_learn_result": lambda x: True,
+                "send_ir_code": lambda x: self._send_input_validation(x),
+                "send_cmd": lambda x: self._send_input_validation(x),
+                "set_power": lambda x: self._set_power(x),
+            }
+        )
```
rytilahti: Please keep the `self.return_values` for now, and simply pass it as a parameter to the implementation. This will keep the diff much more readable and there is no real need to modify these parts at the moment (imo).

miio/vacuum.py
```diff
@@ -596,10 +597,12 @@ def callback(ctx, *args, id_file, **kwargs):
         @dg.resultcallback()
         @dg.device_pass
         def cleanup(vac: Vacuum, *args, **kwargs):
-            if vac.ip is None:  # dummy Device for discovery, skip teardown
+            if (
+                vac.command_sender.ip is None
```
rytilahti: IMO ip is a property of the device, not of the protocol.

miio/device.py
petrkotek: My thinking was that the consumers of Device shouldn't be sending "raw commands" and if they really want to do that, they can still do it using the `raw_command()` method which passes it to the `self.protocol.send()`

miio/vacuum.py
petrkotek: ✅ Brought `ip` and `token` back. Token because of the `dev.token = m.checksum` in `discovery.py` (to be on the safe side).

The discovery will indeed need more work at some point.

miio/tests/test_airconditioningcompanion.py
rytilahti: Honestly I'd prefer it that'd be done in a separate PR. This PR contains already so many lines to review, and as these are tests, not having `return_values` does not really bring much.

miio/device.py
rytilahti: I think that'd be fine, but I'd reword the deprecation message to state that `raw_command` should be used instead. We can drop the `send()` then at some point.

miio/command_sender.py
rytilahti: I started to wonder if this should be under `miioprotocol.py` (maybe current `protocol.py` is misnamed...) considering you are going to add `miotprotocol.py` in your another PR, no?

This way the pieces shared by both protocols would live separately from the specific implementation. What do you think?

petrkotek: yep, we'll probably end up with these 3 classes. maybe we can do the split in the next PR?

miio/device.py
petrkotek: ✅

miio/tests/test_airconditioningcompanion.py
petrkotek: Fair enough. I'm initialising the `DummyProtocol` in `DummyDevice` now and passing in `self` so it can access `return_values`. That way the diff is indeed smaller.

Had to add call to the parent constructor for some dummy devices in tests and add inheritance from `DummyDevice`, but that should faily uncontroversial :)

petrkotek:
Actioned the 2nd round of review. Hope it looks better now.

Thank you for looking into this, @rytilahti!

miio/command_sender.py
rytilahti: Ok, fair enough :+1:

miio/device.py
```diff
@@ -122,205 +117,16 @@ def __init__(
...
-                    seen_addrs.append(addr[0])
-            except socket.timeout:
-                if is_broadcast:
-                    _LOGGER.info("Discovery done")
-                return  # ignore timeouts on discover
-            except Exception as ex:
-                _LOGGER.warning("error while reading discover results: %s", ex)
-                break
+        self.token = token
+        self.protocol = Protocol(ip, token, start_id, debug, lazy_discover)
```
rytilahti: This could be `_protocol` to make it obvious that it's an internal and should not be accessed.

miio/device.py
```diff
@@ -333,7 +139,7 @@ def raw_command(self, command, parameters):
 
         :param str command: Command to send
         :param dict parameters: Parameters to send"""
-        return self.send(command, parameters)
+        return self.protocol.send(command, parameters)
```
rytilahti: I'm starting to have second thoughts on deprecating the `send`. This code would generally be much cleaner if we'd keep it as it is, and simply make it wrap the call to the private protocol. Thoughts on this? What are the pros for deprecating it (and making all other methods more complicated)?

miio/protocol.py
```diff
@@ -3,20 +3,24 @@
 This module contains the implementation of the routines to encrypt and decrypt
 miIO payloads with a device-specific token.
 
-The payloads to be encrypted (to be passed to a device) are excpected to be
+The payloads to be encrypted (to be passed to a device) are expected to be
 JSON objects, the same applies for decryption where they are converted
 automatically to JSON objects.
 If the decryption fails, raw bytes as returned by the device are returned.
 
 An usage example can be seen in the source of :func:`miio.Device.send`.
```
rytilahti: docstring requires fixing here after the change.

miio/device.py
petrkotek: I guess the main reason why I decided to do that was to do less inheritance and more composition (to make things easier to follow). Keeping `send`, however, makes it more backwards compatible.

I've removed the deprecation.

miio/device.py
petrkotek: ✅ 

Had to expose `raw_id` property for `Vacuum` though.

miio/protocol.py
petrkotek: ✅ Updated.

petrkotek:
Done addressing the latest feedback.

Tested on my Air Quality Monitor:
```
(python-miio) bash-3.2$ python miio/cli.py airqualitymonitor --ip $AIR_IP --token $TOKEN status
Power: on
USB power: True
Battery: 100
AQI: 4
Temperature: None
Humidity: None
CO2: None
CO2e: None
PM2.5: None
TVOC: None
Display clock: False
(python-miio) bash-3.2$ python miio/cli.py airqualitymonitor --ip $AIR_IP --token $TOKEN raw_command get_prop "['power', 'aqi', 'battery', 'usb_state', 'time_state', 'night_state', 'night_beg_time', 'night_end_time', 'sensor_state']"
Running command raw_command
['on', 3, 100, 'on', 'off', 'on', 79200, 25200, 'ready']
```

Discovery worked too:
```
(python-miio) bash-3.2$ python miio/vacuum_cli.py discover
INFO:miio.discovery:Discovering devices with mDNS, press any key to quit...
INFO:miio.discovery:Found a supported 'AirQualityMonitor' at 192.168.0.2 - token: bb835d94<REDACTED>
```
I.e. it works fine, however, I don't have any other devices handy.

Further testing appreciated!

miio/protocol.py
```diff
@@ -217,3 +225,225 @@ def _decode(self, obj, context, path):
         Checksum(Bytes(16), Utils.md5, Utils.checksum_field_bytes),
     ),
 )
+
+
+class Protocol:
```
rytilahti: Could you move this to its own file (thus, avoid moving this again soonish and polluting the git logs), I'd say `miioprotocol.py` would be a good candidate.

petrkotek: ✅ Moved `Protocol` to `miioprotocol.py` as `MiIOProtocol`

rytilahti: Great, thanks for the effort! :tada:

PR #608, opened 2020-01-01
s6: two new commands for map handling

### LINKED ISSUE #607: Additional commands for vacuum
To simplify the map handling, you could probably add this lines to vacuum.py:

``` 
    @command()
    def get_recover_maps(self):
        """Get recover maps."""
        return self.send("get_recover_maps")

    @command(
        click.argument("id"),
        click.argument("map"),
    )
    def recover_map(self, id: int, map: int):
        """Set recover maps ."""
        payload = [int(id), int(map)]
        click.echo("Setting the map %s as active" % payload)
        return self.send("recover_map", payload)[0]
```

is working fine with my Roborock S6

please se PR: [https://github.com/rytilahti/python-miio/pull/608](url)

glompfine:
This two commands are working fine with my Roborock s6

Roborock s6 has three saved maps. Command `get_recover_maps` shows you the map id's.
With the command `recover_map` you can set one of the tree maps as active.

Use case:
You manage one map with some no-go-zones and another map without no-go-zones.
Depending on whether the apartment is prepared (chars are on the table, carpets are removed) or not, use one or the other map.

Closes #607

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -546,11 +546,10 @@ class Vacuum(Device):
         return self.send("get_recover_maps")
 
     @command(
-        click.argument("id"),
-        click.argument("map"),
+        click.argument("id"), click.argument("map"),
     )
-    def recover_map(self, id: int, map: int):
-        """Set recover maps."""
+    def use_recover_map(self, id: int, map: int):
+        """Set recover map."""
         payload = [int(id), int(map)]
         click.echo("Setting the map %s as active" % payload)
         return self.send("recover_map", payload)[0]
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -541,18 +541,15 @@ class Vacuum(Device):
         return self.send("get_room_mapping")
 
     @command()
-    def get_recover_maps(self):
-        """Get recover maps."""
+    def get_backup_maps(self):
+        """Get backup maps."""
         return self.send("get_recover_maps")
 
-    @command(
-        click.argument("id"), click.argument("map"),
-    )
-    def use_recover_map(self, id: int, map: int):
-        """Set recover map."""
-        payload = [int(id), int(map)]
-        click.echo("Setting the map %s as active" % payload)
-        return self.send("recover_map", payload)[0]
+    @command(click.argument("id", type=int))
+    def use_backup_map(self, id: int):
+        """Set backup map."""
+        click.echo("Setting the map %s as active" % id)
+        return self.send("recover_map", [id])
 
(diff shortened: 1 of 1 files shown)


miio/vacuum.py
```diff
@@ -540,6 +540,21 @@ def get_room_mapping(self):
         """Retrieves a list of segments."""
         return self.send("get_room_mapping")
 
+    @command()
+    def get_recover_maps(self):
+        """Get recover maps."""
```
rytilahti: Maybe `get_saved_maps`?

miio/vacuum.py
```diff
@@ -540,6 +540,21 @@ def get_room_mapping(self):
...
+        return self.send("get_recover_maps")
+
+    @command(
+        click.argument("id"),
+        click.argument("map"),
+    )
+    def recover_map(self, id: int, map: int):
+        """Set recover maps."""
```
rytilahti: `use_saved_map` in accordance to the above command? What are ID and map in this case? Can you post an example of what get recover maps returns?

miio/vacuum.py
glompfine: Actually its more like recover or backup maps than real saved maps. There is the current map and two backup maps, like this:
20200102_103607229_iOS

miio/vacuum.py
glompfine: Example:
```
miiocli vacuum --ip roboip --token robotoken get_recover_maps
Running command get_recover_maps
```
returns:
```
[[1, 1575723134], [2, 1577540948], [3, 1577807926]]
```
I named the first number (1,2,3) as id and the second (1575723134,1577540948, 1577807926) as map

So in my case the command:
```
miiocli vacuum --ip roboip --token robotoken recover_map 1 1575723134
Running command recover_map
Setting the map [1, 1575723134] as active
ok
```
sets the initial (not edited) map as active:
PNG image

```
miiocli vacuum --ip roboip --token robotoken recover_map 2 1577540948
Running command recover_map
Setting the map [2, 1577540948] as active
ok
```
sets the oldest backup map (the one from 2019.12.28 14:49, at the very bottom of the screeny, map not fully visible) as active:
PNG image 1

```
miiocli vacuum --ip roboip --token robotoken recover_map 3 1577807926
Running command recover_map
Setting the map [3, 1577807926] as active
ok
```
sets the other backup map (the one from 2019.12.31 16:58) as active:
PNG image 2


So I suggest to name it `use_recover_map` according to `get_recover_maps`?

PR #470, opened 2019-01-30
Initial support for xiaomi gateway devices

rytilahti:
This adds preliminary support for Xiaomi Aqara gateway devices. Currently implemented interfaces include controlling the alarm and the light facilities, as well as some other miscellaneous features.

Thanks to @approximatenumber and @starkillerOG for pushing to make this happen :tada:

--- 

This has been rotting on my computer for over a year, so I thought I'd just push this up in case someone wants to work on it - personally I'm not going to do that as I'm using zigbee2mqtt to control my xiaomi devices. The code does not work without modifications, but gives at least a starting point to build upon.

Related to https://github.com/home-assistant/home-assistant/pull/20294

PR #630, opened 2020-02-17
Add support for Xiaomi Aqara Gateway Alarm

starkillerOG:
I am working on integrating the Xiaomi Aqara Gateway Alarm into HomeAssistant as an official component.
Since the "Xiaomi Aqara API (LAN protocol)" does not support the Alarm of the Gateway I use the "Xiaomi's miIO protocol" as implemented in your awesom library.

I know that HomeAssistant requires commans such as included in this PR to be in the upstream library instead of directly in the HomeAssistant code. That is why I want this merged in python-miio.

@rytilahti Could you please merge this PR and push to a new version of python-miio such that I can then create the PR in HomeAssistant to add official support of the Xiaomi Gateway Alarm?

Thank you in advance for your time and effort!

### DIFF
--- a/miio/gateway_alarm.py
+++ b/miio/gateway_alarm.py
@@ -17,6 +17,7 @@ class XiaomiGatewayAlarm(Device):
         """Fetch alarm status from the device."""
         # Response: ['on'], ['off'], ['oning']
         alarm_status = self._device.send("get_arming")
+        alarm_status = alarm_status.pop()
 
         return alarm_status
 
--- a/miio/gateway_alarm.py
+++ b/miio/gateway_alarm.py
@@ -5,18 +5,18 @@ from .device import Device
 
 _LOGGER = logging.getLogger(__name__)
 
+
 class XiaomiGatewayAlarm(Device):
     """Main class representing the Xiaomi Gateway Alarm."""
 
     def __init__(self, Device) -> None:
         self._device = Device
 
-    @command(
-        default_output=format_output("[alarm_status]"))
+    @command(default_output=format_output("[alarm_status]"))
     def status(self):
         """Fetch alarm status from the device."""
         # Response: ['on'], ['off'], ['oning']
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti is the test coverage decrease a problem?
That is the only left failed check.

rytilahti:
Please see #470 – I think it makes sense to get that PR into a publishable state and merge it instead of adding each feature of the gateway separately.

starkillerOG:
@rytilahti I indeed already saw #470 and I agree that if that gets merged the class in this PR schould be combined with that PR in one file.

However for HomeAssistant it probablly does not make sense to chage the xiaomi aqara integration since the "Xiaomi Aqara API (LAN protocol)" allows for subscriptions and thus events beeing pushed to HomeAssistant rather than polling the status, therfore changes are instantly seen in HomeAssistant, which is greath for automations.
I am not sure, but i think the "Xiaomi's miIO protocol" only allows for polling but not subscriptions. (although I do see somthing about listen for events in this npm package https://github.com/aholstenson/miio)

I am not planning on using any other code than the alarm_on, alarm_off and alarm_status.
I am not familliar with most of the code in PR #470, so i do not think i am the right person to clean up that PR.

Would it be possible to merge this PR that only contains the alarm class, and then later if someone has cleaned up PR #470, merge the two in one file?

starkillerOG:
I could change the filename to "gateway.py" to be the same as in PR #470 such that the files can be merged more easily later on?

rytilahti:
I would very much prefer it to be contained in the same PR to avoid multiple variants (see also the another PR #632). I won't be able to do a cleanup in the upcoming weeks, but considering there is some interest on that device I will do that and prepare to merge it when I can.

starkillerOG:
@rytilahti I have moved everthing to my PR https://github.com/rytilahti/python-miio/pull/633, therefore I will close this one

PR #632, opened 2020-02-18
device.py (class Device): Add to method Device.send parametr "sid" fo…

igzero:
…r to send a command to a child zigbee device connected to the XiaomiGateway
Consider adding an additional SID parameter to the Device.send method to allow you to communicate with child zigbee devices connected to the XiaomiGateway. This patch is applied by me in practice, tested, works. What can I see about in my project https://github.com/igzero/xiaomigateway

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -221,20 +221,24 @@ class Device(metaclass=DeviceGroupMeta):
                 _LOGGER.warning("error while reading discover results: %s", ex)
                 break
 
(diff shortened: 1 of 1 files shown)


rytilahti:
Could you please rebase this on top of #470 as there is no need to have it on other devices?

rytilahti:
Replaced with a more generic #653, #651 adds support for the gateway class. Thanks for the PR anyway, @igzero!

PR #636, opened 2020-02-20
Random start_id by default

fsalomon:
I noticed that one of my devices (lumi.acpartner.mcn02) does not execute raw commands if they are sent with the same id. It answers ['ok'], but the desired action is not taken. So instead of starting with 1, this will start with a random id.

### DIFF
--- a/miio/philips_eyecare_cli.py
+++ b/miio/philips_eyecare_cli.py
@@ -105,6 +105,20 @@ def off(dev: miio.PhilipsEyecare):
 
 
 @cli.command()
+@pass_dev
+def eyecare_on(dev: miio.PhilipsEyecare):
+    """Turn eyecare on."""
+    click.echo("Eyecare on: %s" % dev.eyecare_on())
+
+
+@cli.command()
+@pass_dev
+def eyecare_off(dev: miio.PhilipsEyecare):
+    """Turn eyecare off."""
+    click.echo("Eyecare off: %s" % dev.eyecare_off())
+
+
+@cli.command()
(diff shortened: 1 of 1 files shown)


rytilahti:
Yes, this is a known problem and should be fixed, but random is not a good solution here. The reason is that some devices will (likely) require incrementing value for subsequent calls.

The original mirobo client (`vacuum_cli.py`) contains a very hacky way for saving the previously used `msgid`. The proper way would be to have a miiocli cache file for storing information about the known devices, including the last used msgid.

fsalomon:
I see. Well, my device (lumi.acpartner.mcn02) does not expect incremental ids. And I think if people are using the Xiaomi App in conjunction with this lib, incremental ids will fail anyways.

Would you agree that random ids are still a hell of a lot better than static ids? Would you agree that any devices that currently works with static ids will continue working with random ids?

If your answer to both questions with yes, this pull request should be merged.

rytilahti:
I'm not so sure that randomized IDs are going to be better.. Considering that there are so many different devices and likely implementations, it is hard to say if they keep working and I don't really want to get flooded with bug reports if something breaks.

fsalomon:
Hard to believe there could be any implementation _that_ messed up :-)

Should I make it a command line switch then? --randomid

rytilahti:
Fair enough, maybe `--randomize-msgid` flag defaulting (for the time being) to false? I'm just hesitant as I don't think saving that information per device into a cache would be that much more effort in the end. But I'll accept a PR adding a new flag to turn randomization, so feel free to go for it :-)

PR #643, opened 2020-03-15
Add fan_speed_presets() for querying available fan speeds

rytilahti:
This returns a dictionary {name, value} of the available fan speeds,
the main driver being the need to simplify homeassistant integrations for different devices.

For viomi vacuums this is currently straightforward and hard-coded, but we do special handling
for rockrobo devices (based on info() responses):

* If the query succeeds, newer firmware versions (3.5.7+) of v1 are handled as a special case,
  otherwise the new-style [100-105] mapping is returned.

* If the query fails, we fall back to rockrobo v1's percentage-based mapping.
  This happens, e.g., when the vacuum has no cloud connectivity.

Related to #523

Related downstream issues
https://github.com/home-assistant/core/pull/32821
https://github.com/home-assistant/core/issues/31268
https://github.com/home-assistant/core/issues/27268

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -470,8 +470,8 @@ class Vacuum(Device):
         )
 
     @command()
-    def supported_fanspeeds(self) -> Dict[str, int]:
-        """Return dictionary containing supported fanspeeds."""
+    def fan_speed_presets(self) -> Dict[str, int]:
+        """Return dictionary containing supported fan speeds."""
         if self.model is None:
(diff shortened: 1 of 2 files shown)


rytilahti:
This has been tested now with the gen1 (old firmware) as well as Roborock S4 and S55 v2. Unfortunately no updates from anyone running gen1 with a newer firmware, but let's get this merged and released to finally fix the issue for non-gen1 users at least.

Issue #650, opened 2020-03-24
lumi.gateway.v3


miway80:
Hi guys,  

  I am starting with python, I already have token and the gateway ip but I don't know how to start my code, can somebody send me a example or explain me how can i access my gateway and see zigbee devices?

I found it, but i dont know how can i put token and ip and read zigbee goods, for example controller my  Plug = 11

Thanks in advance

rytilahti:
Unfortunately that part of the current gateway is not currently implemented, and it's waiting for someone to step in to do that.

If you are interested in digging in, there is some test code I used during the initial implementation in this commit: https://github.com/rytilahti/python-miio/pull/470/commits/57ff02898dc34f516705ab65c56c6524f71e6c8b

The command to access zigbee devices differs per device (`get_prop_plug` for the plug), which requires passing the device id in the payload. The current code does not allow doing this, so to make it work this needs to be adjusted (see #632). 

If your goal is to access raw zigbee pieces, `GatewayZigbee` class contains some (unimplemented) methods for commands that could allow doing it. I have no clue what are the parameters and how they work (if they even work), just writing this down here in case you or someone else comes up looking and is interested in that.

Long story short, there is no way to do that currently without code changes. To get started accessing the gateway and list the devices connected to it:
```
from miio import Gateway

gw = Gateway("192.168.123.123", "tokentokentoken")
print(gw.devices())
```

miway80:
Thank you very much for answer, i will check these posts and will try to connect to my gateway 
at least

rytilahti:
See https://github.com/rytilahti/python-miio/blob/master/miio/gateway.py#L27 - that lists the known subdevices. I'll create soon a PR to make it not crash but report an error during initialization, but until that you can simply modify that enum to contain that value.

starkillerOG:
@rytilahti I think this ons can also be closed since the issues discussed here are all fixed I think

rytilahti:
The device type of 65 is still unknown/not implemented, so I think this can remain open?

@miway80 could you please test your script with the current master, if it is otherwise working for you? Do you happen to know which device is the one that is causing the error?

miway80:
Sorry for later answer, but i tried and no get info, con the current master i get it

from miio import Gateway

gw = Gateway("192.168.123.123", "tokentokentoken")
print(gw.devices())


pi@rpi:~ $ python3 mi.py
Traceback (most recent call last):
  File "mi.py", line 1, in <module>
    from miio import Gateway
[...]

PR #651, opened 2020-03-25
Gateway raw_child_command

fsalomon:
This adds _raw_child_command_ to _gateway_ device that lets the user specify a child device's SID for the command. This is different from _raw_command_ because the SID is not a "param", it is a top-level key.

Example command (e.g. parent device is lumi.acpartner.v3, child device is lumi.relay.c2acn01):
`miiocli --debug gateway --ip 192.168.xxx.xxx --token xxxx raw_child_command lumi.xxxx toggle_ctrl_neutral "[\"channel_0\",\"on\"]"`

Also added two devices types `AqaraRelay = 54` (lumi.relay.c2acn01) and `AqaraSwitch2 = 135` (lumi.remote.b286acn01) and fixed a small bug related to kwargs

### DIFF
--- a/miio/miioprotocol.py
+++ b/miio/miioprotocol.py
@@ -8,7 +8,7 @@ import codecs
 import datetime
 import logging
 import socket
-from typing import Any, List
+from typing import Any, Dict, List
 
 import construct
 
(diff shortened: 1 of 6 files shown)


fsalomon:
Relate to #632 and #650.

miio/miioprotocol.py
```diff
@@ -141,6 +142,9 @@ def send(self, command: str, parameters: Any = None, retry_count=3) -> Any:
 
         cmd = {"id": self._id, "method": command}
 
+        if self.sid is not None:
```
rytilahti: I would much prefer to add a new kwarg-only option `extra_params` instead of hardcoding sid here, something like this: https://github.com/rytilahti/python-miio/commit/57ff02898dc34f516705ab65c56c6524f71e6c8b#diff-91c92bb750c47e7a1d365652c76443edR102 – `send()` inside gateway could then be overridden to pass the sid parameter where needed.

What do you think?

fsalomon: Yes, you are right. It's just that I'm new to python, I'm just figuring out the syntax as I go. I didn't want to change the method's signature for this fringe feature, but if you have an IDE where you can do this refactoring easily, please do. I'm working with text editor here.

rytilahti: Ok, I'll update this PR when I get some time this weekend. Thanks :-)

rytilahti: Hi, I created a separate PR to add the necessary functionality (#653). Could you please rebase this on top of it? Thanks!

fsalomon: Okay done. Thanks!

rytilahti: Thanks! I think this will need to wait for a bit, I'll get that linked PR done and merged (after 0.5.0 release is done), and let's rebase this then on top of the master. Thanks for testing & patience :-)

rytilahti:
I just merged the extra_parameters PR, could you please rebase this on top of the current master?

starkillerOG:
@rytilahti I think this PR can be closed since I think almost everything is already included in the restructure PR: https://github.com/rytilahti/python-miio/pull/700

rytilahti:
Indeed, if there is still something missing, it is probably easier to start anew on top of the current master.

Issue #642, opened 2020-03-12
shuii.humidifier.jsq001 support


iromeo:
Please add Xiaomi "The fog free of humidifier" support

I've tried to get info, so it is: 
```
$ miiocli airhumidifier --ip ... --token <....> info
Model: shuii.humidifier.jsq001
Hardware version: ESP8266
Firmware version: 1.3.9
Network: {'localIp': '...', 'mask': '255.255.255.0', 'gw': '...'}
AP: {'rssi': -63, 'ssid': 'wifi', 'bssid': '...'}
```

Command `status` doesn't work, I get
```
ERROR:miio.device:Unable to discover a device at address 192.NN.NN.NN
Error: Unable to discover the device 192.NN.NN.NN
```
I'm not sure how to check what information this humidifier could return.

iromeo:
I've tried
```
from miio import AirHumidifier
dev = AirHumidifier(192.168.1.8, 'c9af1xxxxxxxxxxxxxxxxxx')
dev.status()
```
and got error `miio.exceptions.DeviceException: No response from the device`
same error for `dev.send("get_prop", ["power"])`

Seems device has some other communication protocol, because it ansers on empty command:
```
>>> dev.send("", [""])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/romeo/anaconda3/envs/xiaomi/lib/python3.7/site-packages/miio/device.py", line 291, in send
    raise DeviceError(error)
miio.exceptions.DeviceError: {'code': -32602, 'message': 'Invalid param.'}
```

#577

rezmus:
prop for power state is bstart (if get_prop does not work try get_props).

String[] PROPERTIES = { "temperature", "humidity", "mode", "buzzer", "lock", "brightness", "bstart", "waterYield", "lid" };

On (set_start 1)
Off (set_start 0)
Intelligent (set_mode 1)
Level 1 (set_mode 2)
Level 2 (set_mode 3)
Level 3 (set_mode 4)
Level 4 (set_mode 5)

set_brightness
set_buzzer
set_lock
get_timer
get_timer_count
set_timer

iromeo:
@rezmus thx, "get_props" works, thx! And you are right about other commands, only `timer` doesn't work for me
```
>>> dev.send("get_props","")
[24, 37, 3, 1, 0, 2, 1, 0, 0]

# status[0] : temp
# status[1]: humidity
[...]

PR #654, opened 2020-03-27
Add Xiaomi Zero Fog Humidifier (shuii.humidifier.jsq001) support (#642)

iromeo:
Adds support for `shuii.humidifier.jsq001` humidifier (#642)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # Change Log
 
-
 ## [0.4.8](https://github.com/rytilahti/python-miio/tree/0.4.8)
 
 This release adds support for the following new devices:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,5 @@
 # Change Log
 
-This release adds support for the following new devices:
-
-* Xiaomi Zero Fog Humidifier \(shuii.humidifier.jsq001\)
 
 ## [0.4.8](https://github.com/rytilahti/python-miio/tree/0.4.8)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+This release adds support for the following new devices:
+
+* Xiaomi Zero Fog Humidifier \(shuii.humidifier.jsq001\)
+
 ## [0.4.8](https://github.com/rytilahti/python-miio/tree/0.4.8)
 
 This release adds support for the following new devices:
--- a/miio/airhumidifier_jsq.py
+++ b/miio/airhumidifier_jsq.py
@@ -108,6 +108,11 @@ class AirHumidifierStatus:
         return brightness
 
     @property
+    def led(self) -> bool:
+        """True if LED is turned on."""
+        return self.led_brightness is not LedBrightness.Off
+
+    @property
     def child_lock(self) -> bool:
         """Return True if child lock is on."""
         return self.data["child_lock"] == 1
--- a/miio/airhumidifier_jsq.py
+++ b/miio/airhumidifier_jsq.py
@@ -6,7 +6,7 @@ import click
 
 from .airhumidifier import AirHumidifierException
 from .click_common import EnumType, command, format_output
-from .device import Device, DeviceInfo, Optional
+from .device import Device
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -46,7 +46,7 @@ class LedBrightness(enum.Enum):
 class AirHumidifierStatus:
     """Container for status reports from the air humidifier jsq."""
 
-    def __init__(self, data: Dict[str, Any], device_info: DeviceInfo) -> None:
+    def __init__(self, data: Dict[str, Any]) -> None:
         """
         Status of an Air Humidifier (shuii.humidifier.jsq001):
             [24, 30, 1, 1, 0, 2, 0, 0, 0]
@@ -57,7 +57,6 @@ class AirHumidifierStatus:
               'lid_opened': 0}
         """
         self.data = data
-        self.device_info = device_info
 
     @property
     def power(self) -> str:
@@ -127,16 +126,6 @@ class AirHumidifierStatus:
         """True if the water tank is detached."""
         return self.data["lid_opened"] == 1
 
-    @property
-    def firmware_version(self) -> str:
-        """Returns the fw_ver of miIO.info. For example 1.3.9."""
-        return self.device_info.firmware_version
-
-    @property
-    def hardware_version(self) -> Optional[str]:
-        """The hardware version."""
-        return self.device_info.hardware_version
-
     def __repr__(self) -> str:
         s = (
             "<AirHumidiferStatus power=%s, "
@@ -147,9 +136,7 @@ class AirHumidifierStatus:
             "buzzer=%s, "
             "child_lock=%s, "
             "no_water=%s, "
-            "lid_opened=%s, "
-            "hardware_version=%s, "
-            "firmware_version=%s>"
+            "lid_opened=%s>"
             % (
                 self.power,
                 self.mode,
@@ -160,8 +147,6 @@ class AirHumidifierStatus:
                 self.child_lock,
                 self.no_water,
                 self.lid_opened,
-                self.firmware_version,
-                self.hardware_version,
             )
         )
         return s
@@ -187,7 +172,6 @@ class AirHumidifierJsq(Device):
         super().__init__(ip, token, start_id, debug, lazy_discover)
 
         self.model = model if model in AVAILABLE_PROPERTIES else MODEL_HUMIDIFIER_JSQ001
-        self.device_info = None
 
     @command(
         default_output=format_output(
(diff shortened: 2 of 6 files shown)


miio/airhumidifier_jsq.py
```diff
@@ -0,0 +1,330 @@
...
+    @command(
+        click.argument("lock", type=bool),
+        default_output=format_output(
+            lambda lock: "Turning on child lock" if lock else "Turning off child lock"
+        ),
+    )
+    def set_child_lock(self, lock: bool):
+        """Set child lock on/off."""
+        if lock:
+            return self.send("set_lock", [1])
```
rytilahti: You could use only a single `send` here with `[int(lock)]` to cover both cases.

miio/airhumidifier_jsq.py
```diff
@@ -0,0 +1,330 @@
...
+    @command(
+        click.argument("buzzer", type=bool),
+        default_output=format_output(
+            lambda buzzer: "Turning on buzzer" if buzzer else "Turning off buzzer"
+        ),
+    )
+    def set_buzzer(self, buzzer: bool):
+        """Set buzzer on/off."""
+        if buzzer:
+            return self.send("set_buzzer", [1])
```
rytilahti: Same as below.

miio/airhumidifier_jsq.py
```diff
@@ -0,0 +1,330 @@
...
+        return self.send("set_start", [0])
+
+    @command(
+        click.argument("mode", type=EnumType(OperationMode, False)),
+        default_output=format_output("Setting mode to '{mode.value}'"),
+    )
+    def set_mode(self, mode: OperationMode):
+        """Set mode."""
+        value = mode.value
+        if value not in (
```
rytilahti: I'm wondering if this could be simplified by simply testing against all the name in the enum? Something based on `list(OperationModel)`?

miio/airhumidifier_jsq.py
```diff
@@ -0,0 +1,330 @@
...
+            "Lid opened: {result.lid_opened}\n"
+            "Firmware version: {result.firmware_version}\n"
+            "Hardware version: {result.hardware_version}\n",
+        )
+    )
+    def status(self) -> AirHumidifierStatus:
+        """Retrieve properties."""
+
+        if self.device_info is None:
+            self.device_info = self.info()
```
rytilahti: Is this call necessary for functionality?

miio/airhumidifier_jsq.py
```diff
@@ -0,0 +1,330 @@
...
+
+    @property
+    def lid_opened(self) -> bool:
+        """True if the water tank is detached."""
+        return self.data["lid_opened"] == 1
+
+    @property
+    def firmware_version(self) -> str:
+        """Returns the fw_ver of miIO.info. For example 1.3.9."""
+        return self.device_info.firmware_version
```
rytilahti: Better skip properties accessing the `info` here. They are unnecessary and easily accessible through the common info interface for the use cases that require that information.

miio/airhumidifier_jsq.py
```diff
@@ -0,0 +1,330 @@
...
+
+    @property
+    def power(self) -> str:
+        """Power state."""
+        return "on" if self.data["power"] == 1 else "off"
+
+    @property
+    def is_on(self) -> bool:
+        """True if device is turned on."""
+        return self.power == "on"
```
rytilahti: Is there a need for `power` at all? This could be `bool(self.data["power"])` otherwise.

miio/airhumidifier_jsq.py
iromeo: Ok, I agree that cleanup is required here. 

Instead of `int(lock)` I'd rather use smth more type safe. Your version assumes that `lock` is `bool` or `number`, but smb could call this API with wrong type, e.g. `lock="on"` and we will get an exception `ValueError: invalid literal for int() with base 10: 'on'`. I suggest to use smth more type-safe here, e.g. `[int(bool(lock))]` or `[1 if lock else 0]`

miio/airhumidifier_jsq.py
iromeo: The most evident way like `mode in OperationMode` is deprecated in 3.8
```
>>> OperationMode
<enum 'OperationMode'>
>>> 0 in OperationMode
__main__:1: DeprecationWarning: using non-Enums in containment checks will raise TypeError in Python 3.8
False
>>> 8 in OperationMode
False
```

List doesn't work here
```
>>> list(OperationMode)
[<OperationMode.Intelligent: 0>, <OperationMode.Level1: 1>, <OperationMode.Level2: 2>, <OperationMode.Level3: 3>, <OperationMode.Level4: 4>]
>>> 0 in list(OperationMode)
False
```

But we could use a generator. I agree it is much better that hardcoded enum options.
```
>>> [v.value for v in OperationMode]
[0, 1, 2, 3, 4]
```

miio/airhumidifier_jsq.py
iromeo: I've used other devices implementation as a reference. This patter is used for humidifiers. `device_info` is lazily initialized, so we need to init it if not done already

miio/airhumidifier_jsq.py
iromeo: Ok, so just remove `hardware_version` and `firmware_version` methods from `AirHumidifierStatus`? Impl is based on `airhumidifier.py` which defines a similar method. Maybe you need to do cleanup there also.

miio/airhumidifier_jsq.py
iromeo: To be honest I don't have any idea do we need a separate `power` method or not. I tried to make API similar to other humidifiers in implementation, they use the same pattern with `power` method. So should I get rid of it?

miio/airhumidifier_jsq.py
iromeo: Fixed, see updated version.

miio/airhumidifier_jsq.py
iromeo: Fixed, see updated version.

miio/airhumidifier_jsq.py
iromeo: Fixed, see the updated version.

miio/airhumidifier_jsq.py
rytilahti: Ok, fair enough!

miio/airhumidifier_jsq.py
rytilahti: We can leave it as it is, maybe the API will cleaned up at some point. I think personally that it's not really needed, as `is_on` boolean is much nicer API-wise.

miio/airhumidifier_jsq.py
rytilahti: Yes, I would say so.

Btw, is the device so much different than adding support for this device into the `airhumidified.py` would have unnecessarily complicated that?

miio/airhumidifier_jsq.py
rytilahti: I think click does automatic conversion for bool-like types, but that's fine too :-)

miio/airhumidifier_jsq.py
iromeo: `airhumidified.py` isn't a general minimal humidifier impl, it has options with not available for this one, e.g:  `target_humidity`, `motor speed`, `dry`, `depth`, `use_time`, `button_pressed`

Also `airhumidified.py` uses different `mode` and they are named string values, this humidifier uses int values. Brightness modes are different,  e.g. OFF state is 2 for  `airhumidified.py`  and 0 for this one. Status read/write is different, here we get all values at once, and in  `airhumidified.py` we fetch single attribute by name if I'm not mistaken.

> would have unnecessarily complicated that?

I don't think so, but it will be good to revise API and make one that covers all existing use cases.

rytilahti: Ok, sounds good to me! Is there something else you want to change before getting this merged?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+This release adds support for the following new devices:
+
+* Xiaomi Zero Fog Humidifier \(shuii.humidifier.jsq001\)
```
rytilahti: Please remove this for now. The changelog (or parts of it) will be automatically generated as part of the release procedures, which is the time when the high-level changelog gets also updated.

miio/airhumidifier_jsq.py
iromeo: Yes, I'll push firmware/hardware related cleanup soon + also ensure that everything still works in my Home Assistant with updated support of the humidifier in the corresponding HASS component

CHANGELOG.md
iromeo: Done

miio/airhumidifier_jsq.py
iromeo: Done

iromeo:
@rytilahti seems everything is ready, thank you for the detailed review

PR #657, opened 2020-03-28
Add Device.get_properties(), cleanup devices using get_prop

rytilahti:
This will move common handling for property request splitting,
and verifying the results, into the Device class.

Also, add tests for this functionality.


rytilahti:
Let's merge this as it's a showstopper for improving the miot support, this should be anyway a non-breaking change.

PR #660, opened 2020-03-29
Make flake8 happy again

rytilahti:


rytilahti:
Okay, something is terribly, terribly broken here. The original way was working apparently just fine, no idea what is going on with my local flake8...

PR #665, opened 2020-04-11
chuangmi.plug.v3: Fixed power state status for updated firmware

ad:
from [this comment](https://github.com/home-assistant/core/issues/31653#issuecomment-611667270)

### DIFF
--- a/miio/chuangmi_plug.py
+++ b/miio/chuangmi_plug.py
@@ -46,7 +46,7 @@ class ChuangmiPlugStatus:
     def power(self) -> bool:
         """Current power state."""
         if "on" in self.data:
-            return self.data["on"] == True or self.data["on"] == "on"
+            return self.data["on"] or self.data["on"] == "on"
         if "power" in self.data:
             return self.data["power"] == "on"
 
--- a/miio/chuangmi_plug.py
+++ b/miio/chuangmi_plug.py
@@ -46,7 +46,7 @@ class ChuangmiPlugStatus:
     def power(self) -> bool:
         """Current power state."""
         if "on" in self.data:
-            return self.data["on"] or self.data["on"] == "on"
+            return self.data["on"] is True or self.data["on"] == "on"
         if "power" in self.data:
             return self.data["power"] == "on"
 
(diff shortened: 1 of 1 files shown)


rytilahti:
I don't think this this really fixes the issue, considering that the first check will always return True for non-empty string values too (i.e., `off` would be matched, too), right?

miio/chuangmi_plug.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, data: Dict[str, Any]) -> None:
     def power(self) -> bool:
         """Current power state."""
         if "on" in self.data:
-            return self.data["on"]
+            return self.data["on"] or self.data["on"] == "on"
```
rytilahti, line 49:
```suggestion
            return self.data["on"] is True or self.data["on"] == "on"
```

matteos1:
is it possible to correct this error in home assistant?

rytilahti:
It will correct itself when we make a new release, and homeassistant updates to that.

matteos1:
in 0.109?

rytilahti:
No, 0.109 beta is already out so it will likely land in 0.110.

matteos1:
is not possisible in future release of 0.109? thank you

> No, 0.109 beta is already out so it will likely land in 0.110.

rytilahti:
I wanted to get some of the currently open PRs merged first, but maybe I'll cut a new release over the weekend and see if we can get it into some .x homeassistant release!

matteos1:
thank you

matteos1:
> I wanted to get some of the currently open PRs merged first, but maybe I'll cut a new release over the weekend and see if we can get it into some .x homeassistant release!

same problem on 0.110 the switch is already on :(

matteos1:
> I wanted to get some of the currently open PRs merged first, but maybe I'll cut a new release over the weekend and see if we can get it into some .x homeassistant release!

have  you update chuangmi_plug.py?

acelauwc:
As of today, its still not in the home assistant release 0.114.4 

For those who cannot wait, you can manually edit chuangmi_plug.py (line 49) in your home assistant miiio directory to reflect the change.

Issue #666, opened 2020-04-11
error timeout - adding supported to viomi-vacuum-v8_miio 309248236 


Dmxsir:
Hello,
tried to add the vacuum to Home assistant but i'm receiving
`Got error when receiving: timed out` in HA log.
i tried different command with miio and mirobo with the same error

Error: `No response from the device`
model number: **STYTJ02YM**

token got from mi home 5.4.54:
```
SmartHome 813489:[DEBUG]-04-10 14:02:13.919 processResult in result=
{"code":0,"message":"ok","result":{"list":[{"did":"309248236","token":"796XXXXXXXX773","longitude":"0.00000000","latitude":"0.00000000","*
*name":"Mi Robot Vacuum-Mop** P","pid":"0","localip":"10.0.0.18","mac":"XX:XX:XX:XX:XX:XX,"ssid":"XXX","bssid":"XX:XX:XX:XX:XX:XX","
parent_id":"","parent_model":"","show_mode":1,"model":"viomi.vacuum.v8","adminFlag":1,"shareFlag":0
,"permitLevel":16,"isOnline":false,"desc":"Device offline ","extra":
{"isSetPincode":0,"fw_version":"3.5.3_0017","needVerifyCode":0,"isPasswordEncrypt":0}"
```

mirobo discover
`WARNING:miio.discovery:Found unsupported device viomi-vacuum-v8_miio309248236._miio._udp.local. at 10.0.0.18, please report to developers`

miio discover
```
Device ID: 309248236
Model info: Unknown
Address: 10.0.0.18
Token: ???
Support: Unknown
```

 mirobo  --ip 10.0.0.18 --token 796XXXXXXXX773 info
```
Error: No response from the deviceERROR:miio.miioprotocol:Got error when receiving: timed out
```

rytilahti:
I have no real answer for this problem, but timeout usually indicates incorrect token.

Dmxsir:
working commend:
**miiocli viomivacuum** --ip 10.0.0.18 --token XXXXXXXXXXXX info
Model: viomi.vacuum.v8
Hardware version: Linux
Firmware version: 3.5.3_0017
Network: {'localIp': '10.0.0.18', 'mask': '255.0.0.0', 'gw': '10.0.0.1'}
AP: {'ssid': 'XXXX', 'bssid': 'XXXXXXXXX}

PR #668, opened 2020-04-13
add viomi.vacuum.v8 to discovery

rytilahti:
Removes the warning from this already supported device.

Related to #666


rytilahti:
Merging as this is very straightforward.

PR #672, opened 2020-04-18
WIP: improve miotdevice API, add gosund plug support (cuco.plug.cp1)

rytilahti:
This PR aims to add better support for miot devices, and build a common API to simplify contributions for new devices.

* add miot_info() to get the common information from the device
  * this includes manufacturer, model, firmeware version and serial number

* add get_properties_for_dataclass(cls) which allows easy implementation for get_properties mappings
  * each field() can define metadata containing siid and piid, these are mapped automatically to the response container
  * _siid can be used to define common siid, _max_properties can be used to set number of maximum properties per request

* "old" get_properties_for_mapping() requires explicit passing of the mapping, no more passing over `__init__`

* rename old set_property to set_property_from_mapping
    * add set_properties_from_dataclass (allows passing a mapping object with all wanted values at once)
    
    ```
    device.set_properties_from_dataclass(GosundPlugStatus(state=True, some_other_property=123)
    ```
    
    * new set_property helper which takes kwargs that are used automatically with the help of the class-given _MAPPING:
    
    ```
    device.set_property(state=True)
    ```

* Add support for gosund miot plug (cuco.plug.cp1) to show how the API could work.

## TODO
* Naming of helper methods, `_MAPPING`
* Unit tests for all MiotDevice interfaces
* Split set property requests based on `_max_properties`.
* python 3.6 compatibility (dataclasses are available only python 3.7+)
* helper to create fields with wanted metadata, maybe `add_property(piid, siid)`?
* document the API, figure out the best way to do conversions, if needed.
* handling read-only/write-only properties
* common API for actions

### DIFF
--- a/miio/__init__.py
+++ b/miio/__init__.py
@@ -22,6 +22,7 @@ from miio.device import Device
 from miio.exceptions import DeviceError, DeviceException
 from miio.fan import Fan, FanP5, FanSA1, FanV2, FanZA1, FanZA4
 from miio.gateway import Gateway
+from miio.gosund_plug import GosundPlug
 from miio.heater import Heater
 from miio.philips_bulb import PhilipsBulb, PhilipsWhiteBulb
 from miio.philips_eyecare import PhilipsEyecare
--- a/miio/airpurifier_miot.py
+++ b/miio/airpurifier_miot.py
@@ -255,16 +255,6 @@ class AirPurifierMiotStatus:
 class AirPurifierMiot(MiotDevice):
     """Main class representing the air purifier which uses MIoT protocol."""
 
-    def __init__(
-        self,
-        ip: str = None,
-        token: str = None,
-        start_id: int = 0,
-        debug: int = 0,
-        lazy_discover: bool = True,
-    ) -> None:
-        super().__init__(_MAPPING, ip, token, start_id, debug, lazy_discover)
-
     @command(
         default_output=format_output(
             "",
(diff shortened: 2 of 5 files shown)


rytilahti:
@foxel if you have some time, your feedback on how/if the API needs adaptations to allow porting airpurifier_miot over it :-) for the time being this is just an experimental try to see if this approach makes sense in general, alas, the only miot device I have has only a single service (besides the common miot_info) with a single property...

marcelrv:
FYI.
For the zhimi.airpurifier.mb3  all siids respond with value exept for siid 1. siid1 does not respond to get_properties so below ones are not available in reality but are in the spec file.

Device Information-Current Firmware Version
Device Information-Device Manufacturer
Device Information-Device Model
Device Information-Device Serial Number

rytilahti:
Yeah, probably not all manufacturers implement it properly (considering miio.info also reveals enough information). Here's what I'm getting out with that plug:
```
MiotInfo(manufacturer='hello', model='hello', serial_number='hello', firmware_version='hello')
```

xuefer:
the code looks much easier for miot in this branch, can you improve it asap? i want to add more devices like curtain wifi version
i wonder if someone already modify devtools/miottemplate.py to generate for new miio.miot_device class instead of for non-existing miio.MiOTService class

rytilahti:
I'll try to find some time soonish to push this onwards. I have a locally modified devtools that will make integrating a bit easier. I'll create another PR for that when I can.

mkohns:
Hi. Just an idea: What about downloading and parsing the miot spec on new device detection? On this page:
https://home.miot-spec.com/s/<model>
the spec JSON can be retrieved. With a small parser the full device feature set could be utilized.

rytilahti:
@mkohns see the soon-to-be-linked PR, I'm closing this now as obsolete :-)

PR #675, opened 2020-04-19
Xiaomi vacuum. Add property for water box (water tank) attach status

ckesc:


ckesc:
@rytilahti please check this small PR :)

miio/vacuumcontainers.py
```diff
@@ -167,6 +167,11 @@ def is_on(self) -> bool:
             or self.state_code == 17
         )
 
+    @property
+    def is_water_box_attached(self) -> int:
+        """Return True is water box is installed."""
+        return self.data["water_box_status"] == 1
```
rytilahti, line 173: This is not always available, so let's avoid crashing just in case someone tries to access it :-)
```suggestion
        return "water_box_status" in self.data and self.data["water_box_status"] == 1
```

miio/vacuumcontainers.py
```diff
@@ -167,6 +167,11 @@ def is_on(self) -> bool:
             or self.state_code == 17
         )
 
+    @property
+    def is_water_box_attached(self) -> int:
```
rytilahti, line 171:
```suggestion
    def is_water_box_attached(self) -> bool:
```

ckesc:
@rytilahti Cool! 
BTW when will be release? Looking forward for this fix and fix from master about `set_fan_speed`. Currently HomeAssistant can't set fan mode except Medium :( But master version of miio can :)

Offtopic: Maybe there's some chat about this project? Discord / Telegram / whatever?

rytilahti:
There is no real release plan, we just do it after some number of changes / when it seems to be worth to due for downstreams (mainly homeassistant, I suppose). What's the problem with fan speed setting? Is it the one that will get fixed when 0.109 gets released (differing fan speeds on different vacuums)?

There is no chat for the project, but I'm available with the same handle on discord.

ckesc:
@rytilahti Yep looks like this one. Fixed it fro now with passing direct fan values like 104 to service call.

Looks like Discord links users to chats. So I can't find you only by username :( can you make chat for this project and post link?

rytilahti:
I created a matrix room (assuming the link is working correctly): https://matrix.to/#/#python-miio-chat:matrix.org - I might add that to the README file to see if there's interest in such, but it's probably a good idea to have some sort of real-time communication channel for easier debugging etc. if needed :-)

ckesc:
@rytilahti ok) But it seems that group is private and invite only :) I got error "You are not invited to this room."

rytilahti:
@ckesc it should work now, I hope!

PR #686, opened 2020-04-28
send multiple handshake requests

rytilahti:
send several handshake requests as an experiment to improve the transmission rate. reuses retry_count to decide on the number of requests to send.

potential fix for discovery problems, e.g., https://github.com/home-assistant/core/issues/21899


rytilahti:
As it doesn't do much harm to send multiple handshakes, I'll merge it even when it doesn't fix the connectivity issues for everyone.

PR #693, opened 2020-05-07
Add support for chuangmi.plug.hmi208

### LINKED ISSUE #616: Unsupported device found - chuangmi.plug.hmi208
Hello,
This is a Xiaomi Mijia Smart Wi-fi plug with Bluetooth and can act as Bluetooth Gateway. I just saw the code and seems there is nothing like it.  
https://www.gearbest.com/access-control/pp_009225999378.html

Any chances to be integrated ?

rytilahti:
Closes #616


rytilahti:
I don't see why this wouldn't work, so not worth stalling this PR.

biscuitee:
hi, what changes must I do to add support for hmi208 into home assistant? do i need to download the new miio? I am having 2 of the hmi208 switch, but i can only configure it as hmi205.  It works for the power switch, but not the usb switch.  Any help is greatly appreciated.

PR #696, opened 2020-05-15
Added support of Aqara Wireless Relay 2ch (LLKZMK11LM)

bskaplou:
Found commands here => https://github.com/igzero/xiaomigateway

Works as follows:

> $ miiocli gateway --ip 192.168.2.109 --token XXXX devices
> Running command devices
> [<Subdevice DeviceType.Cube: lumi.158d000103ec74 fw: 3 bat: 60>, <Subdevice DeviceType.SensorHT: lumi.158d00040e2330 fw: 2 bat: 60>, <Subdevice DeviceType.Cube: lumi.158d00010202de fw: 3 bat: 60>, <Subdevice DeviceType.Unknown62: lumi.158d000429d96f fw: 5 bat: 60>, <Subdevice DeviceType.AqaraHT: lumi.158d000465ade7 fw: 3 bat: 60>, <Subdevice DeviceType.Unknown15: lumi.158d0003f0d874 fw: 1 bat: 60>, <Subdevice DeviceType.Unknown54: lumi.158d0004454133 fw: 36 bat: 60>]
> $ miiocli gateway --ip 192.168.2.109 --token XXXX relay_get_state lumi.158d0004454133 first
> Running command relay_get_state
> on
> $ miiocli gateway --ip 192.168.2.109 --token XXXX relay_toggle lumi.158d0004454133 first toggle
> Running command relay_toggle
> ok
> $ miiocli gateway --ip 192.168.2.109 --token XXXX relay_toggle lumi.158d0004454133 second on
> Running command relay_toggle
> ok
> $ miiocli gateway --ip 192.168.2.109 --token XXXX relay_get_load_power lumi.158d0004454133
> Running command relay_get_load_power
> 0.06

PR #700, opened 2020-05-20
restructure and improve gateway subdevices

starkillerOG:
Restructure and improve gateway subdevices
- add GatewayException
- add SubDeviceInfo class
- make GatewayAlarm, GatewayRadio, GatewayZigbee and GatewayLight cli compatible
- add 'devices' property that returns already discoverd devices
- Implement "discover_devices" such that it also initilizes subclasses containing the devices, these subclasses are returned as a list by the "discover_devices" command and can be retrieved again from the 'devices' property
- improve and use general getter/setter functions for both the gateway itself and the subdevices
- structure device specific subclasses
- add GatewayDevice class for consistant initialization of gateway devices (such as GatewayAlarm, GatewayRadio, GatewayZigbee and GatewayLight)
- expand SubDevice baseclass with: props class, status property, device_type property, update method, send method, send_arg method, get_property method, get_property_exp method and set_property method
- implement SubDevice Specif classes: AqaraHT, AqaraMagnet, AqaraPlug and AqaraRelayTwoChannels

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -1,5 +1,4 @@
 import logging
-import sys
 from datetime import datetime
 from enum import Enum, IntEnum
 from typing import Optional
@@ -8,9 +7,11 @@ import click
 
 from .click_common import EnumType, command, format_output
 from .device import Device
+from .exceptions import DeviceException
 from .utils import brightness_and_color_to_int, int_to_brightness, int_to_rgb
 
 _LOGGER = logging.getLogger(__name__)
+
 
 color_map = {
     "red": (255, 0, 0),
@@ -24,6 +25,8 @@ color_map = {
     "purple": (128, 0, 128),
 }
 
+class GatewayException(DeviceException):
+    """Exception for the Xioami Gateway communication."""
 
 class DeviceType(IntEnum):
     Unknown = -1
@@ -45,6 +48,8 @@ class DeviceType(IntEnum):
     AqaraMagnet = 53
     AqaraRelayTwoChannels = 54
     AqaraSquareButton = 62
+    RemoteSwitchSingle = 134
+    RemoteSwitchDouble = 135
 
 class Gateway(Device):
     """Main class representing the Xiaomi Gateway.
@@ -126,22 +131,23 @@ class Gateway(Device):
     def discover_devices(self):
         """Discovers SubDevices and returns a list of the discovered devices."""
         # from https://github.com/aholstenson/miio/issues/26
+        device_type_mapping = {"AqaraRelayTwoChannels": AqaraRelayTwoChannels, "Plug": Plug, "SensorHT": SensorHT, "AqaraHT": AqaraHT, "AqaraMagnet": AqaraMagnet}
         devices_raw = self.get_gateway_prop("device_list")
         self._devices = []
-        thismodule = sys.modules[__name__]
         
         for x in range(0, len(devices_raw), 5):
             try:
                 Device_name = DeviceType(devices_raw[x+1]).name
             except ValueError:
-                _LOGGER.warning("Unknown subdevice type '%i' discovered, of Xiaomi gateway with ip: %s", devices_raw[x+1], self.ip)
+                _LOGGER.warning("Unknown subdevice type '%i': %s discovered, of Xiaomi gateway with ip: %s", devices_raw[x+1], devices_raw[x], self.ip)
                 Device_name = DeviceType(-1).name
             
-            SubDeviceClass = getattr(thismodule, Device_name, SubDevice)
-            if SubDeviceClass == SubDevice and Device_name != DeviceType(-1).name:
+            SubDeviceClass = device_type_mapping.get(Device_name, SubDevice)
+            if SubDeviceClass == SubDevice and Device_name != DeviceType(-1).name and Device_name != DeviceType(0).name:
                 _LOGGER.warning("Gateway device type '%s' does not have device specific methods defined, only basic default methods will be available", Device_name)
             
-            self._devices.append(SubDeviceClass(self, *devices_raw[x : x + 5]))
+            if(devices_raw[x]!='lumi.0'):
+                self._devices.append(SubDeviceClass(self, *devices_raw[x : x + 5]))
         
         return self._devices
 
@@ -552,32 +558,37 @@ class SubDevice:
         try:
             return self.gw.send(command, arguments, extra_parameters={"sid": self.sid})
         except Exception as ex:
-            _LOGGER.error(
-                "Got an exception while sending command '%s' with arguments '%s': %s", command, str(arguments), ex, exc_info=True
-            )
-            return None
+            raise GatewayException("Got an exception while sending command '%s' with arguments '%s': %s" % (command, str(arguments), ex))
 
     @command(click.argument("property"))
     def get_subdevice_prop(self, property):
         """Get the value of a property of the subdevice."""
         try:
-            return self.gw.send("get_device_prop", [self.sid, property])
+            response = self.gw.send("get_device_prop", [self.sid, property])
         except Exception as ex:
-            _LOGGER.debug(
-                "Got an exception while fetching property %s: %s", property, ex, exc_info=True
-            )
-            return None
+            raise GatewayException("Got an exception while fetching property %s: %s" % (property, ex))
+        
+        if not response:
+            raise GatewayException("Empty response while fetching property %s: %s" % (property, response))
+        
+        return response
 
     @command(click.argument("properties", nargs=-1))
     def get_subdevice_prop_exp(self, properties):
         """Get the value of a bunch of properties of the subdevice."""
         try:
-            return self.gw.send("get_device_prop_exp", [[self.sid] + list(properties)]).pop()
+            response = self.gw.send("get_device_prop_exp", [[self.sid] + list(properties)]).pop()
         except Exception as ex:
-            _LOGGER.debug(
-                "Got an exception while fetching properties %s: %s", properties, ex, exc_info=True
-            )
-            return None
+            raise GatewayException("Got an exception while fetching properties %s: %s" % (properties, ex))
+
+        if len(list(properties)) != len(response):
+            raise GatewayException("unexpected result while fetching properties %s: %s" % (properties, response))
+        
+        for item in response:
+            if not item:
+               raise GatewayException("One or more empty results while fetching properties %s: %s" % (properties, response)) 
+        
+        return response
 
     @command(click.argument("property"), click.argument("value"))
     def set_subdevice_prop(self, property, value):
@@ -585,10 +596,7 @@ class SubDevice:
         try:
             return self.gw.send("set_device_prop", {"sid": self.sid, property: value})
         except Exception as ex:
-            _LOGGER.debug(
-                "Got an exception while setting propertie %s to value %s: %s", property, str(value), ex, exc_info=True
-            )
-            return None
+            raise GatewayException("Got an exception while setting propertie %s to value %s: %s" % (property, str(value), ex))
 
     @command()
     def unpair(self):
@@ -629,12 +637,9 @@ class AqaraHT(SubDevice):
     def update(self):
         """Update all device properties"""
         values = self.get_subdevice_prop_exp(["temperature", "humidity", "pressure"])
-        if len(values) == 3:
-            self._temperature = values[0]/100
-            self._humidity = values[1]/100
-            self._pressure = values[2]/100
-        return
-
+        self._temperature = values[0]/100
+        self._humidity = values[1]/100
+        self._pressure = values[2]/100
 
 class SensorHT(SubDevice):
     accessor = "get_prop_sensor_ht"
@@ -652,11 +657,22 @@ class SensorHT(SubDevice):
     
     @command()
     def update(self):
-        values = self.get_subdevice_prop_exp(properties)
-        if len(values) == 2:
-            self._temperature = values[0]
-            self._humidity = values[1]
-        return
+        values = self.get_subdevice_prop_exp(self.properties)
+        self._temperature = values[0]
+        self._humidity = values[1]
+
+class AqaraMagnet(SubDevice):
+    _status = None
+
+    @property
+    def status(self):
+        """Returns 'open' or 'closed'"""
+        return self._status
+
+    @command()
+    def update(self):
+        values = self.get_subdevice_prop_exp(["unkown"])
+        self._status = values[0]
 
 class Plug(SubDevice):
     accessor = "get_prop_plug"
@@ -674,11 +690,9 @@ class Plug(SubDevice):
     
     @command()
     def update(self):
-        values = self.get_subdevice_prop_exp(properties)
-        if len(values) == 2:
-            self._power = values[0]
-            self._status = values[1]
-        return
+        values = self.get_subdevice_prop_exp(self.properties)
+        self._power = values[0]
+        self._status = values[1]
 
 class AqaraRelayTwoChannels(SubDevice):
     _status_ch0 = None
@@ -713,11 +727,9 @@ class AqaraRelayTwoChannels(SubDevice):
     def update(self):
         """Update all device properties"""
         values = self.get_subdevice_prop_exp(["load_power", "channel_0", "channel_1"])
-        if len(values) == 3:
-            self._load_power = values[0]
-            self._status_ch0 = values[1]
-            self._status_ch1 = values[2]
-        return
+        self._load_power = values[0]
+        self._status_ch0 = values[1]
+        self._status_ch1 = values[2]
 
     @command(
         click.argument("channel", type=EnumType(AqaraRelayChannel)),
@@ -726,3 +738,4 @@ class AqaraRelayTwoChannels(SubDevice):
     def toggle(self, sid, channel, value):
         """Toggle Aqara Wireless Relay 2ch"""
         return self.subdevice_send_arg("toggle_ctrl_neutral", [channel.value, value.value]).pop()
+
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -159,7 +159,7 @@ class Gateway(Device):
             
             # Obtain the correct subdevice class, ignoring the gateway itself
             subdevice_cls = device_type_mapping.get(device_type)
-            if subdevice_cls is None and device_type != DeviceType(0):
+            if subdevice_cls is None and device_type != DeviceType.Gateway:
                 subdevice_cls = SubDevice
                 _LOGGER.info(
                     "Gateway device type '%s' does not have device specific methods defined, only basic default methods will be available",
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 from datetime import datetime
 from enum import Enum, IntEnum
 from typing import Optional
@@ -25,6 +26,7 @@ color_map = {
 
 
 class DeviceType(IntEnum):
+    Unknown = -1
     Gateway = 0
     Switch = 1
     Motion = 2
@@ -43,18 +45,6 @@ class DeviceType(IntEnum):
     AqaraMagnet = 53
     AqaraRelayTwoChannels = 54
     AqaraSquareButton = 62
-
-
-class AqaraRelayToggleValue(Enum):
-    toggle = "toggle"
-    on = "on"
-    off = "off"
-
-
-class AqaraRelayChannel(Enum):
-    first = "channel_0"
-    second = "channel_1"
-
 
 class Gateway(Device):
     """Main class representing the Xiaomi Gateway.
@@ -104,6 +94,7 @@ class Gateway(Device):
         self._radio = GatewayRadio(self)
         self._zigbee = GatewayZigbee(self)
         self._light = GatewayLight(self)
+        self._devices = []
 
     @property
     def alarm(self) -> "GatewayAlarm":
@@ -126,58 +117,48 @@ class Gateway(Device):
         """Return light control interface."""
         return self._light
 
-    @command()
+    @property
     def devices(self):
-        """Return list of devices."""
+        """Return a list of the already discovered devices."""
+        return self._devices
+
+    @command()
+    def discover_devices(self):
+        """Discovers SubDevices and returns a list of the discovered devices."""
         # from https://github.com/aholstenson/miio/issues/26
-        devices_raw = self.send("get_device_prop", ["lumi.0", "device_list"])
-        devices = [
-            SubDevice(self, *devices_raw[x : x + 5])  # noqa: E203
-            for x in range(0, len(devices_raw), 5)
-        ]
+        devices_raw = self.get_gateway_prop("device_list")
+        self._devices = []
+        thismodule = sys.modules[__name__]
+        
+        for x in range(0, len(devices_raw), 5):
+            try:
+                Device_name = DeviceType(devices_raw[x+1]).name
+            except ValueError:
+                _LOGGER.warning("Unknown subdevice type '%i' discovered, of Xiaomi gateway with ip: %s", devices_raw[x+1], self.ip)
+                Device_name = DeviceType(-1).name
+            
+            SubDeviceClass = getattr(thismodule, Device_name, SubDevice)
+            if SubDeviceClass == SubDevice and Device_name != DeviceType(-1).name:
+                _LOGGER.warning("Gateway device type '%s' does not have device specific methods defined, only basic default methods will be available", Device_name)
+            
+            self._devices.append(SubDeviceClass(self, *devices_raw[x : x + 5]))
+        
+        return self._devices
 
-        return devices
-
-    @command(click.argument("sid"), click.argument("property"))
-    def get_device_prop(self, sid, property):
+    @command(click.argument("property"))
+    def get_gateway_prop(self, property):
         """Get the value of a property for given sid."""
-        return self.send("get_device_prop", [sid, property])
+        return self.send("get_device_prop", ["lumi.0", property])
 
-    @command(click.argument("sid"), click.argument("properties", nargs=-1))
-    def get_device_prop_exp(self, sid, properties):
+    @command(click.argument("properties", nargs=-1))
+    def get_gateway_prop_exp(self, properties):
         """Get the value of a bunch of properties for given sid."""
-        return self.send("get_device_prop_exp", [[sid] + list(properties)])
+        return self.send("get_device_prop_exp", [["lumi.0"] + list(properties)])
 
-    @command(click.argument("sid"), click.argument("property"), click.argument("value"))
-    def set_device_prop(self, sid, property, value):
+    @command(click.argument("property"), click.argument("value"))
+    def set_gateway_prop(self, property, value):
         """Set the device property."""
-        return self.send("set_device_prop", {"sid": sid, property: value})
-
-    @command(
-        click.argument("sid"),
-        click.argument("channel", type=EnumType(AqaraRelayChannel)),
-        click.argument("value", type=EnumType(AqaraRelayToggleValue)),
-    )
-    def relay_toggle(self, sid, channel, value):
-        """Toggle Aqara Wireless Relay 2ch"""
-        return self.send(
-            "toggle_ctrl_neutral",
-            [channel.value, value.value],
-            extra_parameters={"sid": sid},
-        )[0]
-
-    @command(
-        click.argument("sid"),
-        click.argument("channel", type=EnumType(AqaraRelayChannel)),
-    )
-    def relay_get_state(self, sid, channel):
-        """Get the state of Aqara Wireless Relay 2ch for given sid"""
-        return self.send("get_device_prop_exp", [[sid, channel.value]])[0][0]
-
-    @command(click.argument("sid"))
-    def relay_get_load_power(self, sid):
-        """Get the the load power of Aqara Wireless Relay 2ch for given sid"""
-        return self.send("get_device_prop_exp", [[sid, "load_power"]])[0][0]
+        return self.send("set_device_prop", {"sid": "lumi.0", property: value})
 
     @command()
     def clock(self):
@@ -202,12 +183,12 @@ class Gateway(Device):
     @command()
     def timezone(self):
         """Get current timezone."""
-        return self.send("get_device_prop", ["lumi.0", "tzone_sec"])
+        return self.get_gateway_prop("tzone_sec")
 
     @command()
     def get_illumination(self):
         """Get illumination. In lux?"""
-        return self.send("get_illumination")[0]
+        return self.send("get_illumination").pop()
 
 
 class GatewayAlarm(Device):
@@ -247,28 +228,24 @@ class GatewayAlarm(Device):
     def triggering_time(self) -> int:
         """Return the time in seconds the alarm is going off when triggered"""
         # Response: 30, 60, etc.
-        return self._device.send("get_device_prop", ["lumi.0", "alarm_time_len"]).pop()
+        return self._device.get_gateway_prop("alarm_time_len").pop()
 
     @command(click.argument("seconds"))
     def set_triggering_time(self, seconds):
         """Set the time in seconds the alarm is going off when triggered"""
-        return self._device.send(
-            "set_device_prop", {"sid": "lumi.0", "alarm_time_len": seconds}
-        )
+        return self._device.set_gateway_prop("alarm_time_len", seconds)
 
     @command()
     def triggering_light(self) -> int:
         """Return the time the gateway light blinks when the alarm is triggerd"""
         # Response: 0=do not blink, 1=always blink, x>1=blink for x seconds
-        return self._device.send("get_device_prop", ["lumi.0", "en_alarm_light"]).pop()
+        return self._device.get_gateway_prop("en_alarm_light").pop()
 
     @command(click.argument("seconds"))
     def set_triggering_light(self, seconds):
         """Set the time the gateway light blinks when the alarm is triggerd"""
         # values: 0=do not blink, 1=always blink, x>1=blink for x seconds
-        return self._device.send(
-            "set_device_prop", {"sid": "lumi.0", "en_alarm_light": seconds}
-        )
+        return self._device.set_gateway_prop("en_alarm_light", seconds)
 
     @command()
     def triggering_volume(self) -> int:
@@ -536,38 +513,216 @@ class SubDevice:
     def __init__(self, gw, sid, type_, _, __, ___):
         self.gw = gw
         self.sid = sid
-        self.type = DeviceType(type_)
-
-    def unpair(self):
-        return self.gw.send("remove_device", [self.sid])
-
-    def battery(self):
-        return self.gw.send("get_battery", [self.sid])[0]
-
-    def get_firmware_version(self) -> Optional[int]:
-        """Returns firmware version"""
+        self._battery = None
         try:
-            return self.gw.send("get_device_prop", [self.sid, "fw_ver"])[0]
-        except Exception as ex:
-            _LOGGER.debug(
-                "Got an exception while fetching fw_ver: %s", ex, exc_info=True
-            )
-            return None
+            self.type = DeviceType(type_)
+        except ValueError:
+            self.type = DeviceType(-1)
 
     def __repr__(self):
         return "<Subdevice %s: %s fw: %s bat: %s>" % (
-            self.type,
+            self.device_type,
             self.sid,
             self.get_firmware_version(),
-            self.battery(),
+            self.get_battery(),
         )
+
+    @property
+    def device_type(self):
+        return self.type.name
+
+    @property
+    def battery(self):
+        return self._battery
+
+    @command()
+    def subdevice_send(self, command):
+        """Send a command/query to the subdevice"""
+        try:
+            return self.gw.send(command, [self.sid])
+        except Exception as ex:
+            _LOGGER.error(
+                "Got an exception while sending command %s: %s", command, ex, exc_info=True
+            )
+            return None
+
+    @command()
+    def subdevice_send_arg(self, command, arguments):
+        """Send a command/query including arguments to the subdevice"""
+        try:
+            return self.gw.send(command, arguments, extra_parameters={"sid": self.sid})
+        except Exception as ex:
+            _LOGGER.error(
+                "Got an exception while sending command '%s' with arguments '%s': %s", command, str(arguments), ex, exc_info=True
+            )
+            return None
+
+    @command(click.argument("property"))
+    def get_subdevice_prop(self, property):
+        """Get the value of a property of the subdevice."""
+        try:
+            return self.gw.send("get_device_prop", [self.sid, property])
+        except Exception as ex:
+            _LOGGER.debug(
+                "Got an exception while fetching property %s: %s", property, ex, exc_info=True
+            )
+            return None
+
+    @command(click.argument("properties", nargs=-1))
+    def get_subdevice_prop_exp(self, properties):
+        """Get the value of a bunch of properties of the subdevice."""
+        try:
+            return self.gw.send("get_device_prop_exp", [[self.sid] + list(properties)]).pop()
+        except Exception as ex:
+            _LOGGER.debug(
+                "Got an exception while fetching properties %s: %s", properties, ex, exc_info=True
+            )
+            return None
+
+    @command(click.argument("property"), click.argument("value"))
+    def set_subdevice_prop(self, property, value):
+        """Set a device property of the subdevice."""
+        try:
+            return self.gw.send("set_device_prop", {"sid": self.sid, property: value})
+        except Exception as ex:
+            _LOGGER.debug(
+                "Got an exception while setting propertie %s to value %s: %s", property, str(value), ex, exc_info=True
+            )
+            return None
+
+    @command()
+    def unpair(self):
+        return self.subdevice_send("remove_device")
+
+    @command()
+    def get_battery(self):
+        self._battery = self.subdevice_send("get_battery").pop()
+        return self._battery
+
+    @command()
+    def get_firmware_version(self) -> Optional[int]:
+        """Returns firmware version"""
+        return self.get_subdevice_prop("fw_ver").pop()
+
+
+class AqaraHT(SubDevice):
+    _temperature = None
+    _humidity = None
+    _pressure = None
+
+    @property
+    def temperature(self):
+        """Return the temperature in degrees celsius"""
+        return self._temperature
+
+    @property
+    def humidity(self):
+        """Return the humidity in %"""
+        return self._humidity
+
+    @property
+    def pressure(self):
+        """Return the pressure in hPa"""
+        return self._pressure
+
+    @command()
+    def update(self):
+        """Update all device properties"""
+        values = self.get_subdevice_prop_exp(["temperature", "humidity", "pressure"])
+        if len(values) == 3:
+            self._temperature = values[0]/100
+            self._humidity = values[1]/100
+            self._pressure = values[2]/100
+        return
 
 
 class SensorHT(SubDevice):
     accessor = "get_prop_sensor_ht"
     properties = ["temperature", "humidity"]
+    _temperature = None
+    _humidity = None
 
+    @property
+    def temperature(self):
+        return self._temperature
+
+    @property
+    def humidity(self):
+        return self._humidity
+    
+    @command()
+    def update(self):
+        values = self.get_subdevice_prop_exp(properties)
+        if len(values) == 2:
+            self._temperature = values[0]
+            self._humidity = values[1]
+        return
 
 class Plug(SubDevice):
     accessor = "get_prop_plug"
     properties = ["power", "neutral_0"]
+    _power = None
+    _status = None
+
+    @property
+    def power(self):
+        return self._power
+
+    @property
+    def status(self):
+        return self._status
+    
+    @command()
+    def update(self):
+        values = self.get_subdevice_prop_exp(properties)
+        if len(values) == 2:
+            self._power = values[0]
+            self._status = values[1]
+        return
+
+class AqaraRelayTwoChannels(SubDevice):
+    _status_ch0 = None
+    _status_ch1 = None
+    _load_power = None
+
+    class AqaraRelayToggleValue(Enum):
+        toggle = "toggle"
+        on = "on"
+        off = "off"
+
+    class AqaraRelayChannel(Enum):
+        first = "channel_0"
+        second = "channel_1"
+
+    @property
+    def status_ch0(self):
+        """Return the state of channel 0"""
+        return self._status_ch0
+
+    @property
+    def status_ch1(self):
+        """Return the state of channel 1"""
+        return self._status_ch1
+
+    @property
+    def load_power(self):
+        """Return the load power"""
+        return self._load_power
+
+    @command()
+    def update(self):
+        """Update all device properties"""
+        values = self.get_subdevice_prop_exp(["load_power", "channel_0", "channel_1"])
+        if len(values) == 3:
+            self._load_power = values[0]
+            self._status_ch0 = values[1]
+            self._status_ch1 = values[2]
+        return
+
+    @command(
+        click.argument("channel", type=EnumType(AqaraRelayChannel)),
+        click.argument("value", type=EnumType(AqaraRelayToggleValue)),
+    )
+    def toggle(self, sid, channel, value):
+        """Toggle Aqara Wireless Relay 2ch"""
+        return self.subdevice_send_arg("toggle_ctrl_neutral", [channel.value, value.value]).pop()
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -584,7 +584,7 @@ class GatewayLight(GatewayDevice):
         return self._gateway.send("set_rgb", [brightness_and_color])
 
 
-class SubDevice(Device):
+class SubDevice():
     """
     Base class for all subdevices of the gateway
     these devices are connected through zigbee.
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -135,18 +135,18 @@ class Gateway(Device):
         """Discovers SubDevices and returns a list of the discovered devices."""
         # from https://github.com/aholstenson/miio/issues/26
         device_type_mapping = {
-            "AqaraRelayTwoChannels": AqaraRelayTwoChannels,
-            "Plug": AqaraPlug,
-            "SensorHT": SensorHT,
-            "AqaraHT": AqaraHT,
-            "AqaraMagnet": AqaraMagnet,
+            DeviceType.AqaraRelayTwoChannels: AqaraRelayTwoChannels,
+            DeviceType.Plug: AqaraPlug,
+            DeviceType.SensorHT: SensorHT,
+            DeviceType.AqaraHT: AqaraHT,
+            DeviceType.AqaraMagnet: AqaraMagnet,
         }
         devices_raw = self.get_prop("device_list")
         self._devices = []
 
         for x in range(0, len(devices_raw), 5):
             try:
-                device_name = DeviceType(devices_raw[x + 1]).name
+                device_type = DeviceType(devices_raw[x + 1])
             except ValueError:
                 _LOGGER.warning(
                     "Unknown subdevice type '%i': %s discovered, of Xiaomi gateway with ip: %s",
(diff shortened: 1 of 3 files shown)


bskaplou:
@starkillerOG 
;( 

```
$ miiocli gatewayzigbee --ip 192.168.2.109 --token XXXXX get_zigbee_channel
  File "/Users/bskaplou/Library/Caches/pypoetry/virtualenvs/python-miio-NRKm9xRU-py3.7/bin/miiocli", line 11, in <module>
    load_entry_point('python-miio', 'console_scripts', 'miiocli')()
  File "/Users/bskaplou/github/sk/python-miio/miio/cli.py", line 44, in create_cli
    return cli(auto_envvar_prefix="MIIO")
  File "/Users/bskaplou/github/sk/python-miio/miio/click_common.py", line 59, in __call__
  File "/Users/bskaplou/Library/Caches/pypoetry/virtualenvs/python-miio-NRKm9xRU-py3.7/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/bskaplou/Library/Caches/pypoetry/virtualenvs/python-miio-NRKm9xRU-py3.7/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
  File "/Users/bskaplou/Library/Caches/pypoetry/virtualenvs/python-miio-NRKm9xRU-py3.7/lib/python3.7/site-packages/click/core.py", line 1256, in invoke
    Command.invoke(self, ctx)
  File "/Users/bskaplou/Library/Caches/pypoetry/virtualenvs/python-miio-NRKm9xRU-py3.7/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
  File "/Users/bskaplou/Library/Caches/pypoetry/virtualenvs/python-miio-NRKm9xRU-py3.7/lib/python3.7/site-packages/click/core.py", line 610, in invoke
  File "/Users/bskaplou/Library/Caches/pypoetry/virtualenvs/python-miio-NRKm9xRU-py3.7/lib/python3.7/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/bskaplou/github/sk/python-miio/miio/click_common.py", line 242, in group_callback
    ctx.obj = self.device_class(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'ip'
```

Does it work with CLI?

bskaplou:
Too dirty but works


$ git diff && miiocli gatewayzigbee --ip 192.168.2.109 --token XXXXX get_zigbee_channel 
diff --git a/miio/gateway.py b/miio/gateway.py
index 8f024a4..27dc9bd 100644
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -269,6 +269,17 @@ class GatewayZigbee(Device):
     def __init__(self, parent) -> None:
         self._device = parent

+    def __init__(
+        self,
+        ip: str = None,
+        token: str = None,
+        start_id: int = 0,
+        debug: int = 0,
+    ) -> None:
+        self._device = Device(ip, token, start_id, debug, lazy_discover)
+
+
     @command()
     def get_zigbee_version(self):
         """timeouts on device"""
Running command get_zigbee_channel
15
```

**Comment by syssi** [comment]
Same applies to all classes using the @command decorator.

File: miio/gateway.py [hunk 1]
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+    def humidity(self):
+        return self._humidity
+    
+    @command()
+    def update(self):
+        values = self.get_subdevice_prop_exp(properties)
+        if len(values) == 2:
+            self._temperature = values[0]
+            self._humidity = values[1]
+        return
```
rytilahti, line 659: I'd say this should raise an exception if the response is unexpected, to avoid failing quietly.

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+    def status(self):
+        return self._status
+    
+    @command()
+    def update(self):
+        values = self.get_subdevice_prop_exp(properties)
+        if len(values) == 2:
+            self._power = values[0]
+            self._status = values[1]
+        return
```
rytilahti, line 681: Maybe something like this? Adding new exception deriving from `SmartDeviceException`, and then using it to inform the downstream users when something goes wrong.
```
if len(values) != 2:
     raise GatewayException("unexpected result for prop_exp: %s % values)
```

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+
+    @command()
+    def update(self):
+        """Update all device properties"""
+        values = self.get_subdevice_prop_exp(["load_power", "channel_0", "channel_1"])
+        if len(values) == 3:
+            self._load_power = values[0]
+            self._status_ch0 = values[1]
+            self._status_ch1 = values[2]
+        return
```
rytilahti, line 720: See the above comment about error handling.

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+            )
+            return None
+
+    @command()
+    def unpair(self):
+        return self.subdevice_send("remove_device")
+
+    @command()
+    def get_battery(self):
+        self._battery = self.subdevice_send("get_battery").pop()
```
rytilahti, line 599: Is there some specific reason for caching this?

miio/gateway.py
```diff
@@ -126,58 +117,48 @@ def light(self) -> "GatewayLight":
...
-    def get_device_prop(self, sid, property):
+    @command()
+    def discover_devices(self):
+        """Discovers SubDevices and returns a list of the discovered devices."""
+        # from https://github.com/aholstenson/miio/issues/26
+        devices_raw = self.get_gateway_prop("device_list")
+        self._devices = []
+        thismodule = sys.modules[__name__]
+        
+        for x in range(0, len(devices_raw), 5):
```
rytilahti, line 133: This is very convoluted way compared to the previous version...

The better way would be to create a mapping between subdevice types and their implementation classes. Accessing `sys.modules` and `getattr` are usually code smells better to be avoided wherever possible.

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+        return self._battery
 
-    def get_firmware_version(self) -> Optional[int]:
-        """Returns firmware version"""
+    @command()
+    def subdevice_send(self, command):
+        """Send a command/query to the subdevice"""
+        try:
+            return self.gw.send(command, [self.sid])
+        except Exception as ex:
```
rytilahti, line 543: It is better to let this raise an exception and let the downstream to handle it however they prefer.

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+    @property
+    def pressure(self):
+        """Return the pressure in hPa"""
+        return self._pressure
+
+    @command()
+    def update(self):
+        """Update all device properties"""
+        values = self.get_subdevice_prop_exp(["temperature", "humidity", "pressure"])
+        if len(values) == 3:
```
rytilahti, line 632: Error handling. Perhaps `get_subdevice_prop_exp` should check if the number of returned values matches to the requests properties, and raise the exception in there?

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+            self._temperature = values[0]/100
+            self._humidity = values[1]/100
+            self._pressure = values[2]/100
+        return
```
rytilahti, line 636: No need for explicit returns. Error cases should be raised as exceptions instead of silently failing.

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
     def __init__(self, gw, sid, type_, _, __, ___):
         self.gw = gw
         self.sid = sid
-        self.type = DeviceType(type_)
+        self._battery = None
+        try:
+            self.type = DeviceType(type_)
+        except ValueError:
+            self.type = DeviceType(-1)
```
rytilahti, line 520: The type should be set in the inheriting class' init.

miio/gateway.py
```diff
@@ -536,38 +513,216 @@ class SubDevice:
...
+        """Return the humidity in %"""
+        return self._humidity
+
+    @property
+    def pressure(self):
+        """Return the pressure in hPa"""
+        return self._pressure
+
+    @command()
+    def update(self):
```
rytilahti, line 629: Wrt. `@command` inside subdevices, have you thought how are these supposed to be exposed to the cli?

starkillerOG:
@bskaplou I only work with python3, I know almost nothing about the cli implementation, so honestly I have no idea if it works with cli, probably not yet...

miio/gateway.py
starkillerOG: My idea was that this function can be called in an update command of the device_specific class that updates all properties in one command.
The HomeAssistant properties can then be directly mapped on the properties of this class.
So in case of the AqaraTH sensor you call update once and it updates pressure, temperature, humidity and battery level in one go.

miio/gateway.py
starkillerOG: I will improve on this but will take a few days since I don't have much time now

miio/gateway.py
starkillerOG: Why?
In this way you only need to write the code ones, otherwise this needs to be copied into all device_type specific classes

miio/gateway.py
starkillerOG: Honestly I don't know how the cli works...

In python it would work like this:

from miio import (gateway)
gateway = gateway.Gateway("192.168.1.IP", 'token')
devices = gateway.discover_devices()

Now select the sub device you want to control, for instance the first discovered device:

Aqara1 = devices[0]
Aqara1.update()
print(Aqara1.temperature)

starkillerOG:
@bskaplou could you test using python3?

In python3 this schould work:

from miio import (gateway)
gateway = gateway.Gateway("192.168.1.IP", 'token')
devices = gateway.discover_devices()

Now select the sub device you want to control, for instance the first discovered device:

Aqara1 = devices[0]
Aqara1.update()
print(Aqara1.temperature)

bskaplou:
@starkillerOG Yep it works... But CLI is important because it allows to play with lib and it's well documented... 

Right now CLI doesn't work because click is trying to create object with ip/token/etc passed to __init__

you can resurrect CLI by adding following method to each subdevice
```
def __init__(
     self,
     ip: str = None,
     token: str = None,
     start_id: int = 0,
     debug: int = 0,
     lazy_discover: bool = True,
) -> None:
     self._device = Device(ip, token, start_id, debug, lazy_discover)
```

Click will call this __init__ with cli arguments and everything works well

bskaplou:
@rytilahti yes sure :) Waiting for https://github.com/rytilahti/python-miio/pull/700

About letting me in ... I'm just a bit frustrated with a bunch of pending PRs all around ...
Some of them are accepted https://github.com/home-assistant/core/commit/37f9b24a8567accdc349fab04f0606697c52809a https://github.com/home-assistant/core/commit/37cabe556289e31c95cc7eb39d9b97f00e7c9f1c https://github.com/rytilahti/python-miio/pull/696

Some other are too sloooowwwww this one, https://github.com/kennedyshead/aioasuswrt/pull/39 https://github.com/home-assistant/core/pull/35778

starkillerOG:
> 
> 
> Too dirty but works
> 
> ```
> $ git diff && miiocli gatewayzigbee --ip 192.168.2.109 --token XXXXX get_zigbee_channel 
> diff --git a/miio/gateway.py b/miio/gateway.py
> index 8f024a4..27dc9bd 100644
> --- a/miio/gateway.py
> +++ b/miio/gateway.py
> @@ -269,6 +269,17 @@ class GatewayZigbee(Device):
>      def __init__(self, parent) -> None:
>          self._device = parent
> 
> +    def __init__(
> +        self,
> +        ip: str = None,
> +        token: str = None,
> +        start_id: int = 0,
> +        debug: int = 0,
> +        lazy_discover: bool = True,
> +    ) -> None:
> +        self._device = Device(ip, token, start_id, debug, lazy_discover)
> +
> +
>      @command()
>      def get_zigbee_version(self):
>          """timeouts on device"""
> Running command get_zigbee_channel
> 15
> ```

This will probably work, but if I understand correctly this will create a new device including connection, handshake etc, for each feature of the Gateway + for each subdevice. That seems to be very dirty and will probably give us lots of trouble in the future.

I know almost nothing about the cli and don't use it.....

@bskaplou could you try to figure out a way that works with the cli but only uses one instance of the device class that is beeing re-used to only have one connection and handshake etc?

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: I now use a dictionary lookup instead, unfortunately I cannot define the dictionary at the top of the file since the class defenitions need to be executed first before this dict can be defined.

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: I added the GatewayException class

miio/gateway.py
starkillerOG: done

miio/gateway.py
rytilahti: Looks much better! Hmm, maybe it will make sense to refactor the gateway to be contained in its own directory, and defining the separate devices in separate files:
```
miio/gateway/gateway.py
miio/gateway/aqaramagnet.py
miio/gateway/light.py
miio/gateway/zigbee.py
```
etc. Or at least have the subdevice classes in their own file?

miio/gateway.py
```diff
@@ -126,58 +122,49 @@ def light(self) -> "GatewayLight":
...
+    def discover_devices(self):
+        """Discovers SubDevices and returns a list of the discovered devices."""
+        # from https://github.com/aholstenson/miio/issues/26
+        device_type_mapping = {"AqaraRelayTwoChannels": AqaraRelayTwoChannels, "Plug": Plug, "SensorHT": SensorHT, "AqaraHT": AqaraHT, "AqaraMagnet": AqaraMagnet}
+        devices_raw = self.get_gateway_prop("device_list")
+        self._devices = []
+        
+        for x in range(0, len(devices_raw), 5):
+            try:
+                Device_name = DeviceType(devices_raw[x+1]).name
```
rytilahti, line 140:
```suggestion
                device_name = DeviceType(devices_raw[x+1]).name
```

miio/gateway.py
```diff
@@ -126,58 +122,49 @@ def light(self) -> "GatewayLight":
...
+            except ValueError:
+                _LOGGER.warning("Unknown subdevice type '%i': %s discovered, of Xiaomi gateway with ip: %s", devices_raw[x+1], devices_raw[x], self.ip)
+                Device_name = DeviceType(-1).name
```
rytilahti, line 143:
```suggestion
                device_name = DeviceType(-1).name
```

miio/gateway.py
```diff
@@ -126,58 +122,49 @@ def light(self) -> "GatewayLight":
...
+            
+            SubDeviceClass = device_type_mapping.get(Device_name, SubDevice)
```
rytilahti, line 145: How about simply checking for Noneness?
```suggestion
            subdevice_cls = device_type_mapping.get(Device_name)
```

miio/gateway.py
```diff
@@ -126,58 +122,49 @@ def light(self) -> "GatewayLight":
...
+            if SubDeviceClass == SubDevice and Device_name != DeviceType(-1).name and Device_name != DeviceType(0).name:
```
rytilahti, line 146:
```suggestion
            if subdevice_cls is None:
```

miio/gateway.py
```diff
@@ -202,140 +189,160 @@ def set_developer_key(self, key):
...
 
 class GatewayAlarm(Device):
     """Class representing the Xiaomi Gateway Alarm."""
-
-    def __init__(self, parent) -> None:
-        self._device = parent
+    
+    def __init__(
+        self,
+        parent: Gateway = None,
```
rytilahti, line 205: `parent` should be the last parameter (this way the base signature doesn't change).

miio/gateway.py
```diff
@@ -202,140 +189,160 @@ def set_developer_key(self, key):
...
         """Set time the alarm stays at 'oning' before transitioning to 'on'"""
-        return self._device.send("set_arm_wait_time", [seconds])
+        return self._gateway.send("set_arm_wait_time", [seconds])
 
     @command()
     def triggering_time(self) -> int:
         """Return the time in seconds the alarm is going off when triggered"""
         # Response: 30, 60, etc.
-        return self._device.send("get_device_prop", ["lumi.0", "alarm_time_len"]).pop()
+        return self._gateway.get_gateway_prop("alarm_time_len").pop()
```
rytilahti, line 249: Do you think it would make sense to have all those getters be without gateway? The object itself knows its did, so `self._gateway.get_prop()` and `self.gateway.light.get_prop()` should work just fine.

miio/gateway.py
```diff
@@ -452,8 +471,20 @@ def set_default_music(self):
...
+        token: str = None,
+        start_id: int = 0,
+        debug: int = 0,
+        lazy_discover: bool = True,
+    ) -> None:
+        if parent is not None:
+            self._gateway = parent
+        else:
+            self._gateway = Device(ip, token, start_id, debug, lazy_discover)
+            _LOGGER.debug("Creating new device instance, only use this for cli interface")
```
rytilahti, line 487: Considering this gets repeated for every "control" interface, maybe add a new `GatewayDevice` class which has the modified `__init__`. Avoids copy&paste.

miio/gateway.py
rytilahti: The idea with having this in a class-specific init is a commonly used pattern and initialized class will/should always know its type by its own.

However, this is not necessary anymore at all as the type is defined already at the initialization?

miio/gateway.py
```diff
@@ -529,45 +560,230 @@ def set_light(self, color_name, brightness):
...
-            self.get_firmware_version(),
-            self.battery(),
-        )
+    @command()
+    def subdevice_send_arg(self, command, arguments):
+        """Send a command/query including arguments to the subdevice"""
+        try:
+            return self._gw.send(command, arguments, extra_parameters={"sid": self.sid})
+        except Exception as ex:
+            raise GatewayException("Got an exception while sending command '%s' with arguments '%s': %s" % (command, str(arguments), ex))
```
rytilahti, line 609: Instead of passing ex as a string, use `raise GatewayException() from ex`, this way the downstream user will receive the full stacktrace.

miio/gateway.py
```diff
@@ -529,45 +560,230 @@ def set_light(self, color_name, brightness):
...
+
+    @command(click.argument("property"))
+    def get_subdevice_prop(self, property):
+        """Get the value of a property of the subdevice."""
+        try:
+            response = self._gw.send("get_device_prop", [self.sid, property])
+        except Exception as ex:
+            raise GatewayException("Got an exception while fetching property %s: %s" % (property, ex))
```
rytilahti, line 617: `raise GatewayException(...) from ex`. I won't repeat this for for similar cases below, but please change them also to be consistent.

miio/gateway.py
rytilahti: The cli collects all methods of `Device`-derived class marked with `@command`, and adds them to `click.group` for that sub command. 

Unfortunately I don't have a clean solution either for this at the moment, but the aim would be to gain access to the group object of the parent class, which can be extended by adding new sub-groups (`zigbee`, `light`, ..) with their own commands.

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: That is indeed better

miio/gateway.py
starkillerOG: yes that is better, although I still need to filter out the gateway device itself.
Does not make sense to make a subdevice for the gateway itself, but it is discovered.

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: thanks, have implemented it now

miio/gateway.py
```diff
@@ -126,58 +125,64 @@ def light(self) -> "GatewayLight":
...
+                _LOGGER.warning(
+                    "Unknown subdevice type '%i': %s discovered, of Xiaomi gateway with ip: %s",
+                    devices_raw[x + 1],
+                    devices_raw[x],
+                    self.ip,
+                )
+                device_name = DeviceType(-1).name
+
+            subdevice_cls = device_type_mapping.get(device_name)
+            if subdevice_cls is None and device_name != DeviceType(0).name:
```
rytilahti, line 160: So using enum and skipping the name checking, the complete sub-device initialization logic would be following:
1. Construct DeviceType
2. If it fails (because of missing type, print out the warning like right now)
3. Check `if dev_type == DeviceType.Gateway`, ignore gateways (just `continue` the loop)
4. Initialize the device class.

miio/gateway.py
```diff
@@ -126,58 +125,64 @@ def light(self) -> "GatewayLight":
...
+            "SensorHT": SensorHT,
+            "AqaraHT": AqaraHT,
+            "AqaraMagnet": AqaraMagnet,
+        }
+        devices_raw = self.get_gateway_prop("device_list")
+        self._devices = []
+
+        for x in range(0, len(devices_raw), 5):
+            try:
+                device_name = DeviceType(devices_raw[x + 1]).name
```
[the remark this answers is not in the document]
starkillerOG: in principle we could just map the number directly to the class, but that gets confusing realy fast since you would constantly need to look up which number represents which device...

What do you prefer, schould I change it to a direct number mapping and still have the enum only for debug messages etc?

miio/gateway.py
starkillerOG: it is nice to check if the dev_type (int) is in the enum and print a warning if it is not, this allows us to discover more dev_types from users reporting those warnings.

miio/gateway.py
starkillerOG: I will remove the gateway

miio/gateway.py
rytilahti: To add to that, the sub devices should not derive from `Device`, but we need to find a nicer way to enable cli access for those. As you mentioned already at https://github.com/rytilahti/python-miio/pull/700#issuecomment-632600444, having these initialized separately as Devices will cause problems in the long run..

So until we find a solution, I think we will simply have to leave these unexposed (or add some sort of simple interface to the gateway class to print out the details of known sub-devices. Maybe it will make sense to add a property to get the available information as dict for downstream users? Something like:
```
@property
def properties():
    return {"temperature": self._temperature, "humidity": self._humidity, "pressure": self._pressure}
```

miio/gateway.py
starkillerOG: done

miio/gateway.py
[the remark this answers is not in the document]
rytilahti: Hmm, I don't understand what do you mean by checking the number? Wouldn't the type information be enough?

miio/gateway.py
rytilahti: We can still print the value when the conversion fails, and ask the users to create an issue :-)

miio/gateway.py
starkillerOG: I now have the Gateway, GatewayAlarm, GatewayRadio, GatewayZigbee and GatewayLight working in cli, although not in the most beautifull way, but for the python implementation nothing changed and still only one device is initialized.

For subdevices things get dificult very soon......
I would like to leave that for someone else better experianced with cli and click

miio/gateway.py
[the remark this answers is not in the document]
starkillerOG: ah right, I misunderstood your first message, I think I now know what you mean
will try implementing in a minute, hold on

miio/gateway.py
```diff
@@ -126,58 +125,64 @@ def light(self) -> "GatewayLight":
...
+            "AqaraRelayTwoChannels": AqaraRelayTwoChannels,
+            "Plug": AqaraPlug,
+            "SensorHT": SensorHT,
+            "AqaraHT": AqaraHT,
+            "AqaraMagnet": AqaraMagnet,
+        }
+        devices_raw = self.get_prop("device_list")
+        self._devices = []
+
+        for x in range(0, len(devices_raw), 5):
```
rytilahti, line 147: Related to the dev_type number conversation below, I think the first line in this loop should be:
```dev_type_id = devices_raw[x + 1]```

Maybe even extract rest of the known elements here, too, instead of passing `*devices_raw[x : x + 5]` later on? This would make the code much more understandable.

File: miio/gateway.py (hunk 21 above)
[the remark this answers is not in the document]
Comment (starkillerOG) [reply miio/gateway.py]: Is my latest commit what you meant?

File: miio/gateway.py [hunk 23]
```diff
@@ -135,33 +135,33 @@ def discover_devices(self):
...
                     devices_raw[x],
                     self.ip,
                 )
-                device_name = DeviceType(-1).name
+                device_type = DeviceType(-1)
 
-            subdevice_cls = device_type_mapping.get(device_name)
-            if subdevice_cls is None and device_name != DeviceType(0).name:
+            subdevice_cls = device_type_mapping.get(device_type)
+            if subdevice_cls is None and device_type != DeviceType(0):
```
rytilahti, line 160: This is more readable:
```suggestion
            if subdevice_cls is None and device_type != DeviceType.Gateway:
```

starkillerOG: nice

miio/gateway.py
```diff
@@ -155,15 +156,17 @@ def discover_devices(self):
...
             if subdevice_cls is None and device_type != DeviceType(0):
                 subdevice_cls = SubDevice
                 _LOGGER.info(
                     "Gateway device type '%s' does not have device specific methods defined, only basic default methods will be available",
                     device_type.name,
                 )
-
+            
+            # Initialize and save the subdevice, ignoring the gateway itself
             if devices_raw[x] != "lumi.0":
```
rytilahti, line 170: We have the device type already here, so simply use `if device_type != DeviceType.Gateway`.

starkillerOG: true, will change

starkillerOG: done

miio/gateway.py
starkillerOG: sounds good, I know the unknown3 == firmware version number of the subdevice

miio/gateway.py
rytilahti: Okay, I can fetch my gateway and a test device and try at least to expose the property values for the cli (after we get this PR merged).

miio/gateway.py
starkillerOG: at least `self.get_subdevice_prop("fw_ver").pop()` results in 3x '3' and 2x '9' for my 5 devices and those numbers match the unknown3 parameter

miio/gateway.py
starkillerOG: That would be very nice

starkillerOG: will clean up those parms now

miio/gateway.py
starkillerOG: I have implemented a dataclass and made it a bit nicer

miio/gateway.py
```diff
@@ -145,14 +155,22 @@ def discover_devices(self):
         self._devices = []
 
         for x in range(0, len(devices_raw), 5):
+            # Extract discovered information
+            dev_info = SubDeviceInfo
```
rytilahti, line 159: Simply use `SubDeviceInfo(*devices_raw[x : x + 5])` here, it should work just fine :-)

miio/gateway.py
rytilahti: Great! :100:

miio/gateway.py
```diff
@@ -145,14 +155,22 @@ def discover_devices(self):
...
             try:
-                device_type = DeviceType(devices_raw[x + 1])
+                device_type = DeviceType(dev_info.type_id)
             except ValueError:
                 _LOGGER.warning(
                     "Unknown subdevice type '%i': %s discovered, of Xiaomi gateway with ip: %s",
-                    devices_raw[x + 1],
-                    devices_raw[x],
+                    dev_info.type_id,
+                    dev_info.sid,
```
rytilahti, line 173: Print out the whole dev_info object out here, dataclasses print out nicely (and maybe someone, sometime, somewhere will see such a warning and figures out what those unknowns are)

miio/gateway.py
```diff
@@ -529,45 +561,291 @@ def set_light(self, color_name, brightness):
...
+    def subdevice_send(self, command):
+        """Send a command/query to the subdevice"""
+        try:
+            return self._gw.send(command, [self.sid])
+        except Exception as ex:
+            raise GatewayException(
+                    "Got an exception while sending command %s"
+                    % (command)
+                ) from ex
+            return None
```
rytilahti, line 649: Remove this unreachable return.

miio/gateway.py
starkillerOG: you are right, just tested and looks very good

miio/gateway.py
starkillerOG: nice

starkillerOG:
I think this can now be merged

miio/gateway.py
```diff
@@ -529,45 +581,295 @@ def set_light(self, color_name, brightness):
...
+            self.type = DeviceType(dev_info.type_id)
+        except ValueError:
+            self.type = DeviceType(-1)
 
     def __repr__(self):
         return "<Subdevice %s: %s fw: %s bat: %s>" % (
-            self.type,
+            self.device_type,
             self.sid,
             self.get_firmware_version(),
```
rytilahti, line 626: I just tested this PR on my gateway (lumi.gateway.v3, firmware 1.4.1_150) does not seem to support this command which makes it crash when doing `discover_devices` via the cli:
```
DEBUG:miio.miioprotocol:192.168.xx.xx:54321 >>: {'id': 5, 'method': 'get_device_prop', 'params': ['lumi.158d00015694a4', 'fw_ver']}
DEBUG:miio.miioprotocol:192.168.xx.xx:54321 (ts: 1970-01-01 00:08:52, id: 5) << {'id': 5, 'error': {'code': -30001, 'message': 'Resp invalid json.'}}
ERROR:miio.miioprotocol:Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
DEBUG:miio.click_common:Exception: Got an exception while fetching property fw_ver
```
with all devices (SensorHT, Plug, Motion, Magnet).

rytilahti:
test script:
```
from miio import Gateway

gw = Gateway("192.168.xx.xx", "91c52a27eff00b95xxxx7")
devs = gw.discover_devices()
for dev in devs:
    try:
        dev.update()
    except Exception as ex:
        print("got exception during update for %s: %s" % (dev.type, ex))
    print(dev)
```

produces the following output:

```
got exception during update for DeviceType.Plug: One or more empty results while fetching properties ['power', 'neutral_0']: ['', 'unknown']
Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
<Subdevice Plug: lumi.158d00015694a4 fw: <version unavailable> bat: 60>
got exception during update for DeviceType.Magnet: update() missing 2 required positional arguments: 'url' and 'md5'
Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
<Subdevice Magnet: lumi.158d000129252c fw: <version unavailable> bat: 60>
got exception during update for DeviceType.Magnet: update() missing 2 required positional arguments: 'url' and 'md5'
Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
<Subdevice Magnet: lumi.158d00012ae837 fw: <version unavailable> bat: 60>
got exception during update for DeviceType.SensorHT: One or more empty results while fetching properties ['temperature', 'humidity']: [10000, 0]
Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
<Subdevice SensorHT: lumi.158d0001581018 fw: <version unavailable> bat: 5>
got exception during update for DeviceType.Motion: update() missing 2 required positional arguments: 'url' and 'md5'
Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
<Subdevice Motion: lumi.158d00012f5446 fw: <version unavailable> bat: 60>
```

the url, md5 errors are due to SubDevice not having `update` so it inherits that from `Device`...

miio/gateway.py
```diff
@@ -529,45 +581,295 @@ def set_light(self, color_name, brightness):
...
+        try:
+            return self._gw.send(command, arguments, extra_parameters={"sid": self.sid})
+        except Exception as ex:
+            raise GatewayException(
+                "Got an exception while sending "
+                "command '%s' with arguments '%s'" % (command, str(arguments))
+            ) from ex
+
+    @command(click.argument("property"))
+    def get_subdevice_prop(self, property):
```
rytilahti, line 663: `get_prop` and `get_prop_*` just like with the main device to be consistent all around?

miio/gateway.py
```diff
@@ -529,45 +581,295 @@ def set_light(self, color_name, brightness):
...
+            if not item:
+                raise GatewayException(
+                    "One or more empty results while "
+                    "fetching properties %s: %s" % (properties, response)
+                )
+
+        return response
+
+    @command(click.argument("property"), click.argument("value"))
+    def set_subdevice_prop(self, property, value):
```
rytilahti, line 707: `set_prop`? Or maybe all getters/setters should be using non-shortened form for prop actually... `set_property`, `get_property` are not that much longer..

miio/gateway.py
```diff
@@ -529,45 +581,295 @@ def set_light(self, color_name, brightness):
...
+    @command()
+    def get_firmware_version(self) -> Optional[int]:
+        """Returns firmware version"""
+        return self.get_subdevice_prop("fw_ver").pop()
+
+
+class AqaraHT(SubDevice):
+    _temperature = None
+    _humidity = None
+    _pressure = None
```
rytilahti, line 737: These need to be declared inside `__init__`, otherwise they are class attributes and are shared between objects (if you have multiple subdevices using the same class).

miio/gateway.py
```diff
@@ -529,45 +581,295 @@ def set_light(self, color_name, brightness):
...
+
+    @property
+    def load_power(self):
+        """Return the load power"""
+        return self._load_power
+
+    @command()
+    def update(self):
+        """Update all device properties"""
+        values = self.get_subdevice_prop_exp(["load_power", "channel_0", "channel_1"])
```
rytilahti, line 862: Define the property names as class attributes (like other subdevices are already doing).

miio/gateway.py
rytilahti: `get_firmware_version()` should probably catch the error, and return None if it's not available as the return value is already defined as `Optional[int]`?

miio/gateway.py
```diff
@@ -529,45 +581,295 @@ def set_light(self, color_name, brightness):
...
+            ) from ex
+
+        if len(list(properties)) != len(response):
+            raise GatewayException(
+                "unexpected result while fetching properties %s: %s"
+                % (properties, response)
+            )
+
+        for item in response:
+            if not item:
```
rytilahti, line 698: I think this should be handled by the sub-device class as the interpretation of a non-true value can be device-specific, see my comment on the devices I tested against.

miio/gateway.py
rytilahti: If you have test devices for this, you could give dataclasses a try here. So simply convert this to:
```
@dataclass
class AqaraHT(SubDevice):
    temperature: int = None
    humidity: int = None
    pressure: int = None
```
remove the properties and set these values inside the update :-)

miio/gateway.py
starkillerOG: I have firmware version 1.4.1_175 for my gateway, with model lumi.gateway.v3-MW300.
There it does work, regardless I will fix this to use the value from the discovery if this command fails

miio/gateway.py
starkillerOG: I made it more consistant

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: done

starkillerOG:
> 
> 
> test script:
> 
> ```
> from miio import Gateway
> 
> gw = Gateway("192.168.xx.xx", "91c52a27eff00b95xxxx7")
> devs = gw.discover_devices()
> for dev in devs:
>     try:
>         dev.update()
>     except Exception as ex:
>         print("got exception during update for %s: %s" % (dev.type, ex))
>     print(dev)
> ```
> 
> produces the following output:
> 
> ```
> got exception during update for DeviceType.Plug: One or more empty results while fetching properties ['power', 'neutral_0']: ['', 'unknown']
> Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
> <Subdevice Plug: lumi.158d00015694a4 fw: <version unavailable> bat: 60>
> got exception during update for DeviceType.Magnet: update() missing 2 required positional arguments: 'url' and 'md5'
> Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
> <Subdevice Magnet: lumi.158d000129252c fw: <version unavailable> bat: 60>
> got exception during update for DeviceType.Magnet: update() missing 2 required positional arguments: 'url' and 'md5'
> Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
> <Subdevice Magnet: lumi.158d00012ae837 fw: <version unavailable> bat: 60>
> got exception during update for DeviceType.SensorHT: One or more empty results while fetching properties ['temperature', 'humidity']: [10000, 0]
> Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
> <Subdevice SensorHT: lumi.158d0001581018 fw: <version unavailable> bat: 5>
> got exception during update for DeviceType.Motion: update() missing 2 required positional arguments: 'url' and 'md5'
> Got error when receiving: {'code': -30001, 'message': 'Resp invalid json.'}
> <Subdevice Motion: lumi.158d00012f5446 fw: <version unavailable> bat: 60>
> ```
> 
> the url, md5 errors are due to SubDevice not having `update` so it inherits that from `Device`, so that should be changed.
> 
> I'll try to figure out the properties for those devices, and think about how to expose the properties in a generic way.

Do you know why the 'Resp invalid json.' error occurs?
Have you tryed upgrading the gateway firmware?

miio/gateway.py
starkillerOG: I made it more consistant

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: I made it more specific, I now check if `item==''` which is the response if the property is not known by the gateway.

rytilahti: Item can be empty also otherwise as shown in that output I gave, or I think that the values are what they are as the plug is not connnected. I don't exactly know where the plug is located, but if I find it tomorrow I'll plug it in and see what it returns after that :-)

rytilahti:
I haven't tried updating the gateway as I'm not using it at all. I think that error just indicates that the device doesn't have that command.

miio/gateway.py
```diff
@@ -202,267 +230,289 @@ def set_developer_key(self, key):
...
     def set_triggering_volume(self, volume):
         """Set the volume level at which alarms go off [0-100]"""
-        return self._device.send("set_alarming_volume", [volume])
+        return self._gateway.send("set_alarming_volume", [volume])
 
     @command()
     def last_status_change_time(self):
-        """Return the last time the alarm changed status, type datetime.datetime"""
-        return datetime.fromtimestamp(self._device.send("get_arming_time").pop())
+        """
```
rytilahti, line 336: Could you make this (and other similar cases, if any) one-liners by removing datetime.timetime and defining it as return value type hint.

miio/gateway.py
```diff
@@ -529,45 +581,302 @@ def set_light(self, color_name, brightness):
...
-class SubDevice:
-    def __init__(self, gw, sid, type_, _, __, ___):
-        self.gw = gw
-        self.sid = sid
-        self.type = DeviceType(type_)
+class SubDevice(Device):
+    """
+    Base class for all subdevices of the gateway
+    these devices are connected through zigbee.
+    """
```
rytilahti, line 591: Didn't we discuss this already, that subdevices must not derive from `Device` due to cli issues? Please refactor this only to take gateway and dev_info as parameters.

miio/gateway.py
rytilahti: bump.

miio/gateway.py
starkillerOG: I just commited an improved version, could you check if this is the way you imagened?

miio/gateway.py
starkillerOG: sorry I forgot, done now

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: But if the device is not connected I think it is fine to raise an exception messaging that an empty response is received.
I just want to prevent that for instance the empty string '' is beeing devided by a number and raises an exception that is hard to understand (such as is done in the AqaraHT device update function where the raw response is devided by 100 to convert to degrees celsius).

Schould I remove the empty check?

starkillerOG:
@rytilahti I see you have a "SensorHT" device, could you check for me if that device supports the "pressure" propertie like the "AqaraHT" device? 

I also saw that your "SensorHT" device reported a temperature of 10000 and humidity of 0, for me with "AqaraHT" I need to devide by 100 to get the temperature in degrees celcius and humidity in %, however for you that would mean 100 degrees celcius and a humidity of 0%, seems to me like strange values.....
Do you know why those values are like this?

starkillerOG:
I have the following devices:
- AqaraHT
- AqaraMagnet
- RemoteSwitchSingle
- RemoteSwitchDouble

But I cannot figure out the device properties of the AqaraMagnet or RemoteSwitches.......

@rytilahti @bskaplou How did you guys sniff the Xiaomi Home app to figure out the device properties?

starkillerOG:
I think everything is done, what else schould I do before this can get merged?

starkillerOG:
my test script:
```
from miio import Gateway

gateway = Gateway("192.168.1.ip", "token")

gateway.discover_devices()
devices = gateway.devices

for dev in devices:
    if hasattr(dev, 'update'):
        try:
            dev.update()
        except Exception as ex:
            print("got exception during update for %s: %s" % (dev.device_type, ex))
    print(dev)
```

output:
```
<Subdevice AqaraHT: lumi.xxx fw: 3 bat: 60 props: {'temperature': 25.55, 'humidity': 37.9, 'pressure': 1027.63}>
<Subdevice AqaraHT: lumi.xxx fw: 3 bat: 60 props: {'temperature': 26.71, 'humidity': 35.83, 'pressure': 1027.62}>
<Subdevice AqaraHT: lumi.xxx fw: 3 bat: 60 props: {'temperature': 23.87, 'humidity': 41.41, 'pressure': 1027.72}>
got exception during update for AqaraMagnet: unexpected result while fetching properties ['unkown']: []
<Subdevice AqaraMagnet: lumi.xxx fw: 3 bat: 60 props: {'status': None}>
<Subdevice RemoteSwitchSingle: lumi.xxx fw: 9 bat: 60 props: {}>
<Subdevice RemoteSwitchDouble: lumi.xxx fw: 9 bat: 60 props: {}>
```

rytilahti:
Looks great! I'm not completely happy with the props thing (as I realized we lose the docs), but I think it's good that there is now an interface to get the props.

One more change (besides fixing linting) to be done before this is ready to ship: add `update()` to `SubDevice`:
```
def update():
    """Update the device-specific properties."""
```

so there is no need to check if the update exist.

rytilahti:
> @rytilahti I see you have a "SensorHT" device, could you check for me if that device supports the "pressure" propertie like the "AqaraHT" device?

It looks like it supports that. Here's the output with `pressure`:
```
[10000, 0, 0]
```
and here with an incorrect property:
```
[10000, 0, '']
```

> I also saw that your "SensorHT" device reported a temperature of 10000 and humidity of 0, for me with "AqaraHT" I need to devide by 100 to get the temperature in degrees celcius and humidity in %, however for you that would mean 100 degrees celcius and a humidity of 0%, seems to me like strange values.....
> Do you know why those values are like this?

Yes, those are sensor-specific "invalid" values, if I recall correctly. I'm not sure why it's reporting invalids, maybe the gateway cannot access the sensor (as it's bound to my zigbee2mqtt setup)?

miio/gateway.py
```diff
@@ -529,45 +580,261 @@ def set_light(self, color_name, brightness):
...
+                "unexpected result while fetching properties %s: %s"
+                % (properties, response)
+            )
+
+        for item in response:
+            if item == '':
+                raise GatewayException(
+                    "One or more empty results while "
+                    "fetching properties %s: %s" % (properties, response)
+                )
```
rytilahti, line 699: Please remove this. The empty value can very well be a valid value, or it may just a placeholder for invalid state or whatever depending on the device. It is up to `SubDevice` subdevices to handle it accordingly, if raising an exception is wanted behavior.

rytilahti: Also, please update the PR description to describe briefly the changes, so that it will be easier for anyone reading the changelog/merged PRs to get an idea.

miio/gateway.py
```diff
@@ -529,45 +580,261 @@ def set_light(self, color_name, brightness):
...
+        self._battery = None
+        self._fw_ver = dev_info.fw_ver
+        try:
+            self.type = DeviceType(dev_info.type_id)
+        except ValueError:
+            self.type = DeviceType(-1)
+        
+        # if dataclass props is defined, initialize an instance
+        if hasattr(self, 'props'):
+            self._props = self.props()
```
rytilahti, line 608: One more comment for this. You should simply initialize it always (and add an empty props for the main class), this way there is no need to do any sort of checking and the API is the same no matter what sort of device it is and what it exposes.

miio/gateway.py
rytilahti: Yes, please remove this. You can add a check for the values prior the conversion (and raise an exception if the values are unexpected), or simply wrap the conversions inside try-catch block and raise a `GatewayException` with a friendly message when bad things happen.

miio/gateway.py
starkillerOG: I have removed it

miio/gateway.py
starkillerOG: done

miio/gateway.py
```diff
@@ -151,7 +151,7 @@ def discover_devices(self):
         device_type_mapping = {
             DeviceType.AqaraRelayTwoChannels: AqaraRelayTwoChannels,
             DeviceType.Plug: AqaraPlug,
-            DeviceType.SensorHT: SensorHT,
+            DeviceType.SensorHT: AqaraHT,
```
rytilahti, line 154: I would keep the old class in place, as maybe the conversion is different or there are other differences.. Better just let it be as it was.

miio/gateway.py
```diff
@@ -666,11 +668,6 @@ def get_property(self, property):
                 "Got an exception while fetching property %s" % (property)
             ) from ex
 
-        if response == []:
-            raise GatewayException(
-                "Empty response while fetching property '%s': %s" % (property, response)
-            )
```
rytilahti: I think this can stay, but check with `if not response:`. Or am I mistaken, that this command should always be responded with some value (even if it's `''`)?

miio/gateway.py
starkillerOG: Alright I will duplicate those classes again, just a moment and then it is ready to merge

starkillerOG: done

miio/gateway.py
starkillerOG: You are right will add it back

starkillerOG: done

rytilahti:
Btw, if you are developing locally, the simplest way to make sure the checks are passing is to execute `tox -e lint` and fix the issues before pushing in the changes!

starkillerOG:
@rytilahti thanks for the tox -e lint hint, however I get a gazillion errors mostly about the doc files ending with .rst
How do you use tox to only check one specific file, in this case gateway.py?

rytilahti:
Can you paste an error or two? But looks like it's now passing, the only problem is that python3.6 does not support dataclasses, it seems... Please add dataclasses to the required packages (either by manually modifying pyproject.toml, or simply commanding `poetry add dataclasses`) and making a commit.

I think we are bound to have that until homeassistant drops python 3.6 support..
I think it's maybe time to drop 3.6 support in the next release.

starkillerOG:
@rytilahti do you know how to fix the tests failing?
somthing about the "dataclasses" module not beeing found???

starkillerOG:
```
_______________ ERROR collecting miio/tests/test_wifirepeater.py _______________
ImportError while importing test module '/Users/runner/runners/2.169.0/work/1/s/miio/tests/test_wifirepeater.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
miio/__init__.py:25: in <module>
    from miio.gateway import Gateway
miio/gateway.py:4: in <module>
    from dataclasses import asdict as dataclasses_asdict
E   ModuleNotFoundError: No module named 'dataclasses'
_________________ ERROR collecting miio/tests/test_yeelight.py _________________
ImportError while importing test module '/Users/runner/runners/2.169.0/work/1/s/miio/tests/test_yeelight.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
miio/__init__.py:25: in <module>
    from miio.gateway import Gateway
miio/gateway.py:4: in <module>
    from dataclasses import asdict as dataclasses_asdict
E   ModuleNotFoundError: No module named 'dataclasses'
=============================== warnings summary ===============================
/Users/runner/Library/Caches/pypoetry/virtualenvs/python-miio-5D3ggvah-py3.6/lib/python3.6/site-packages/construct/core.py:3
  /Users/runner/Library/Caches/pypoetry/virtualenvs/python-miio-5D3ggvah-py3.6/lib/python3.6/site-packages/construct/core.py:3: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import struct, io, binascii, itertools, collections, pickle, sys, os, tempfile, hashlib, importlib, imp
```

rytilahti:
Please commit the `poetry.lock` file, too, then it should work!

starkillerOG:
running poetry add dataclasses

```
Using version ^0.7 for dataclasses

Updating dependencies
Resolving dependencies...

[SolverProblemError]
The current project's Python requirement (^3.6.5) is not compatible with some of the required packages Python requirement:
  - dataclasses requires Python >=3.6, <3.7

Because dataclasses (0.7) requires Python >=3.6, <3.7
 and no versions of dataclasses match >0.7,<0.8, dataclasses is forbidden.
So, because python-miio depends on dataclasses (^0.7), version solving failed.
```

I think we have a problem....

rytilahti:
Ah, after reading this https://github.com/python-poetry/poetry/issues/1413#issuecomment-621178566 the simplest fix is to modify the required python version... So please replace this:
```
python = "^3.6.5"
```
with
```
python = ">=3.6.5,<4.0"
```

starkillerOG:
nope does not work:

```
Using version ^0.7 for dataclasses

Updating dependencies
Resolving dependencies...

[SolverProblemError]
The current project's Python requirement (>=3.6.5,<4.0) is not compatible with some of the required packages Python requirement:
  - dataclasses requires Python >=3.6, <3.7

Because dataclasses (0.7) requires Python >=3.6, <3.7
 and no versions of dataclasses match >0.7,<0.8, dataclasses is forbidden.
So, because python-miio depends on dataclasses (^0.7), version solving failed.
```

rytilahti:
Argh, okay.. Well, I think it's better get rid of `dataclass` and use `attrs` (which we are already using anyway in this projcet) instead.

The conversion is pretty simple:
* Import `attr`
* Use `@attr.s(auto_attribs=True)` instead of `@dataclass` for the props classes. The auto attributes will allow avoiding defining the properties like `humidity = attr.ib()`
* Use `attr.asdict()` instead of the dataclass variant

The good thing with this is also (which I think we should have converted them anyway) that this will allow use converters (maybe we want to convert, e.g., the open/close or on/off to booleans)?

starkillerOG:
@rytilahti I think I found a solution

rytilahti:
@starkillerOG ah, I saw the commit. Could you still do the attrs conversion, or should I do that? It's better in the long run to make it properly anyway..

starkillerOG:
Not today, I am going to bed, but I will do the conversion tommorow, or you can merge it now and do the conversion...
what you like

rytilahti:
Okay, no need to hurry with that, let's merge it then tomorrow (or whenever you find time to do that). Thanks for the great work already! :1st_place_medal:

starkillerOG:
Thanks for all your help @rytilahti

PR #709, opened 2020-05-28
Zigbee suqare button and cube support via fake_device

bskaplou:
Add support for square button and cube actions via script installation on gate and events receiving with fake_device.

Also added zigbee_command to make functionality avialable through CLI

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -4,6 +4,7 @@ import logging
 from datetime import datetime
 from enum import Enum, IntEnum
 from typing import Optional
+from random import randrange
 
 import attr
 import click
@@ -24,7 +25,6 @@ from .gateway_scripts import (
     build_shakeair,
     build_singlepress,
     build_taptap,
-    tokens,
 )
 from .utils import brightness_and_color_to_int, int_to_brightness, int_to_rgb
 
@@ -222,8 +222,8 @@ class Gateway(Device):
         """Delete script by id."""
         return self.send("miIO.xdel", [script_id])
 
-    @command(click.argument("sid"), click.argument("command"))
-    def subdevice_command(self, sid, command):
+    @command(click.argument("sid"), click.argument("command"), click.argument("encoded_token"))
+    def subdevice_command(self, sid, command, encoded_token):
         """Send command to subdevice."""
         self.discover_devices()
         target = list(filter(lambda subdevice: subdevice.sid == sid, self.devices))
@@ -232,17 +232,18 @@ class Gateway(Device):
         elif not hasattr(target[0], command):
             return f"Device with sid={sid} has no method {command}"
         else:
-            return getattr(target[0], command)()
+            return getattr(target[0], command)(encoded_token)
 
-    def install_script(self, sid, builder, ip=None):
+    def install_script(self, sid, builder, encoded_token, ip=None):
         """Install script for by building script source and sending it with miio method. You need to run fake or real device to capture script execution results."""
         if ip is None:
             addresses = ipv4_nonloop_ips()
             my_ip = addresses[0]  # Taking first public IP
         else:
             my_ip = ip
-        data_tkn = tokens["data_tkn"]
-        source = builder(sid, my_ip)
+
+        data_tkn = randrange(5000, 10000)
+        source = builder(sid, my_ip, encoded_token)
         return self.send(
             "send_data_frame",
             {
@@ -792,7 +793,7 @@ class SubDevice:
         return self._fw_ver
 
     @command()
-    def uninstall_scripts(self):
+    def uninstall_scripts(self, encoded_token):
         return dict(
             map(
                 lambda action: (
@@ -997,34 +998,34 @@ class Cube(SubDevice):
     properties = []
 
     @command()
-    def install_move_script(self):
+    def install_move_script(self, encoded_token):
         """Generate and install script which captures move event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_move)
+        return self._gw.install_script(self.sid, build_move, encoded_token)
 
     @command()
-    def install_rotate_script(self):
+    def install_rotate_script(self, encoded_token):
         """Generate and install script which captures rotate event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_rotate)
+        return self._gw.install_script(self.sid, build_rotate, encoded_token)
 
     @command()
-    def install_shake_script(self):
+    def install_shake_script(self, encoded_token):
         """Generate and install script which captures shake in air event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_shakeair)
+        return self._gw.install_script(self.sid, build_shakeair, encoded_token)
 
     @command()
-    def install_flip90_script(self):
+    def install_flip90_script(self, encoded_token):
         """Generate and install script which captures horizontal 90 flip and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_flip90)
+        return self._gw.install_script(self.sid, build_flip90, encoded_token)
 
     @command()
-    def install_taptap_script(self):
+    def install_taptap_script(self, encoded_token):
         """Generate and install script which captures double tap on surface event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_taptap)
+        return self._gw.install_script(self.sid, build_taptap, encoded_token)
 
     @command()
-    def install_flip180_script(self):
+    def install_flip180_script(self, encoded_token):
         """Generate and install script which captures horizontal 180 flip and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_flip180)
+        return self._gw.install_script(self.sid, build_flip180, encoded_token)
 
 
 class AqaraSquareButton(SubDevice):
@@ -1033,21 +1034,21 @@ class AqaraSquareButton(SubDevice):
     properties = []
 
     @command()
-    def install_singlepress_script(self):
+    def install_singlepress_script(self, encoded_token):
         """Generate and install script which captures single press event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_singlepress)
+        return self._gw.install_script(self.sid, build_singlepress, encoded_token)
 
     @command()
-    def install_doublepress_script(self):
+    def install_doublepress_script(self, encoded_token):
         """Generate and install script which captures double press event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_doublepress)
+        return self._gw.install_script(self.sid, build_doublepress, encoded_token)
 
     @command()
-    def install_longpress_script(self):
+    def install_longpress_script(self, encoded_token):
         """Generate and install script which captures loooong press event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_longpress)
+        return self._gw.install_script(self.sid, build_longpress, encoded_token)
 
     @command()
-    def install_shake_script(self):
+    def install_shake_script(self, encoded_token):
         """Generate and install script which captures shake in air event and sends miio package to device"""
-        return self._gw.install_script(self.sid, build_shake)
+        return self._gw.install_script(self.sid, build_shake, encoded_token)
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -224,6 +224,7 @@ class Gateway(Device):
 
     @command(click.argument("sid"), click.argument("command"))
     def subdevice_command(self, sid, command):
+        """Send command to subdevice."""
         self.discover_devices()
         target = list(filter(lambda subdevice: subdevice.sid == sid, self.devices))
         if len(target) < 1:
@@ -233,9 +234,13 @@ class Gateway(Device):
         else:
             return getattr(target[0], command)()
 
-    def install_script(self, sid, builder):
-        addresses = ipv4_nonloop_ips()
-        my_ip = addresses[0]  # Taking first public IP ;(
+    def install_script(self, sid, builder, ip=None):
+        """Install script for by building script source and sending it with miio method. You need to run fake or real device to capture script execution results."""
+        if ip is None:
+            addresses = ipv4_nonloop_ips()
+            my_ip = addresses[0]  # Taking first public IP
+        else:
+            my_ip = ip
         data_tkn = tokens["data_tkn"]
         source = builder(sid, my_ip)
         return self.send(
@@ -993,26 +998,32 @@ class Cube(SubDevice):
 
     @command()
     def install_move_script(self):
+        """Generate and install script which captures move event and sends miio package to device"""
         return self._gw.install_script(self.sid, build_move)
 
     @command()
     def install_rotate_script(self):
+        """Generate and install script which captures rotate event and sends miio package to device"""
         return self._gw.install_script(self.sid, build_rotate)
 
     @command()
     def install_shake_script(self):
+        """Generate and install script which captures shake in air event and sends miio package to device"""
         return self._gw.install_script(self.sid, build_shakeair)
 
     @command()
     def install_flip90_script(self):
+        """Generate and install script which captures horizontal 90 flip and sends miio package to device"""
         return self._gw.install_script(self.sid, build_flip90)
 
     @command()
     def install_taptap_script(self):
+        """Generate and install script which captures double tap on surface event and sends miio package to device"""
         return self._gw.install_script(self.sid, build_taptap)
 
     @command()
     def install_flip180_script(self):
+        """Generate and install script which captures horizontal 180 flip and sends miio package to device"""
         return self._gw.install_script(self.sid, build_flip180)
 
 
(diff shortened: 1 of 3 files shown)


miio/gateway.py
```diff
@@ -921,3 +985,55 @@ def update(self):
...
+
+
+class Cube(SubDevice):
+    """Subdevice Cube specific properties and methods"""
+
+    properties = []
+
+    @command()
+    def install_move_script(self):
+        return self._gw.install_script(self.sid, "move", build_move)
```
starkillerOG, line 997: I would rename all of these to "subscribe_to_move" etc.
And make an aditional function "subscribe_to_all_events" that calls all these functions in one go.

maybe consider removing all these individual functions and just make one list "events" like the "properties" that contains all events that can be subscribed to ["move", "rotate", "shakeair", "flip90", "taptap", "flip180"] and then just make one function "subscribe_to_all_events" that subscibes to all events in a for loop.

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+    "longpress": lambda sid: action_prefix + "6" + sid_to_num(sid),
+    "flip90": lambda sid: action_prefix + "7" + sid_to_num(sid),
+    "flip180": lambda sid: action_prefix + "8" + sid_to_num(sid),
+    "shakeair": lambda sid: action_prefix + "9" + sid_to_num(sid),
+    "taptap": lambda sid: action_prefix + "10" + sid_to_num(sid),
+}
+
+
+def build_move(
+    source_sid,
```
starkillerOG, line 47: all of these build functions are just copies of each-other, I would combine all of them in one build_function.
(see other comments for the things that change per build function)

consider renaming to "build_subscription_info"

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+    target_ip,
+    target_model=fake_device_model,
+    target_id=fake_device_id,
+    source_model="lumi.sensor_cube.v1",
```
starkillerOG, line 51: the source_model will be a required parameter in the general function

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+    message_id=0,
+):
```
starkillerOG, line 53: add one parameter: "event" which equals "move" for this case and in other cases is "flip90" etc...

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+
+    lumi, source_id = source_sid.split(".")
+    method_name = f"move_{source_id}"
```
starkillerOG, line 56: move --> {event}

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+
+    move = [
```
starkillerOG, line 58: move -> subscription_info

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+        [
+            action_id["move"](source_sid),
```
starkillerOG, line 60: move --> event

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+            [
+                "1.0",
+                randint(1590161094, 1590162094),
+                [
+                    "0",
+                    {
+                        "did": source_sid,
+                        "extra": "[1,18,2,85,[6,256],0,0]",
```
starkillerOG, line 68: The 256 is specific for the move event, so make a mapping dict at the top of this file that maps events to these specific values:

extra_parameter_mapping = {"move": 256, "flip90": 64} etc.

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+                        "key": "event." + source_model + ".move",
```
starkillerOG, line 69: "event." + source_model + ".move"-> f"event.{source_model}.{event}"

miio/gateway_scripts.py
```diff
@@ -0,0 +1,542 @@
...
+                    }
+                ],
+            ],
+        ]
+    ]
+
+    return dumps(move)
+
+
+def build_flip90(
```
starkillerOG, line 95: The rest of these functions are then not needed anymore

miio/gateway.py
```diff
@@ -201,6 +218,38 @@ def discover_devices(self):
...
+        self.discover_devices()
+        target = list(filter(lambda subdevice: subdevice.sid == sid, self.devices))
+        if len(target) < 1:
+            return f"Device with sid={sid} not found"
+        elif not hasattr(target[0], command):
+            return f"Device with sid={sid} has no method {command}"
+        else:
+            return getattr(target[0], command)()
+
+    def install_script(self, sid, action, builder):
```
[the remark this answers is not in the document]
bskaplou: This function is actually install script to gateway. It doesn't subscribe to events... It can even install script which will callback to other device...

miio/gateway.py
```diff
@@ -738,6 +787,21 @@ def get_firmware_version(self) -> Optional[int]:
             )
         return self._fw_ver
 
+    @command()
+    def uninstall_scripts(self):
```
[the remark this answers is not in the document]
bskaplou: this is actual installation of program into gateway memory.... installation and deletion...

miio/gateway.py
```diff
@@ -201,6 +218,38 @@ def discover_devices(self):
 
         return self._devices
 
+    def x_del(self, script_id):
```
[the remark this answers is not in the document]
bskaplou: x_del is name of method... python method name mimics MIMO method... Less conversions => more comprehension... x_del is not executed directly, it's low-level method which is called from unsubscribe ...

starkillerOG:
@bskaplou I have not yet been able to sniff local direct traffic to or from my gateway.
I do see the UDP packets on my desktop using wireshark but that is a diffrent protecoll and since I use my phone for the MiHome app it does not capture direct packets I think.

On my phone I also have an capture app "Packet Capture", that worked greath for another application, but with the MiHome app it just captures hundreds of packets to outside servers (I pressume the Xioami Cloud servers, but I do not see local packets).

What app are you capturing, is it the MiHome app?
How are you capturing, which program do you use and do you do it on your phone or do u use an android/ios emulater on your PC and then capture with your PC?

miio/gateway_scripts.py
starkillerOG: looks much better to me!

miio/gateway.py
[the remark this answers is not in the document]
starkillerOG: alright I understand, you are probably right if this function is only called by other functions

miio/gateway.py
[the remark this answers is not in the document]
starkillerOG: Could you explain what you can do with loading such a "program" into gateway memory?
Can you only get callbacks when an event happens/property changes or can it do more?

It would be really nice if you could include docstrings """explanation"""" for all new functions that briefly explain what the function can do and how it works.

miio/gateway_scripts.py
```diff
@@ -0,0 +1,332 @@
...
+        )
+    )
+
+
+def build_shake(
+    source_sid,
+    target_ip,
+    target_model=fake_device_model,
+    target_id=fake_device_id,
+    source_model="lumi.sensor_switch.aq3",
```
starkillerOG, line 317: maybe consider adding "lumi.sensor_switch.aq3" as a property of the subdevice class:
```
    @property
    def model(self):
        """Return the model of the device."""
        return "lumi.sensor_switch.aq3"
```
also include a default for the base class:
```
    @property
    def model(self):
        """Return the model of the device."""
        return "lumi.unknown"
```

I think this can be usefull for more functions in the future and this can be used to show in HomeAssistant as the model in the device info in the GUI.

That property can than be passed to these functions.

starkillerOG: Probably someone will eventually figure out how to get the model directly from the gateway, I have tryed but not suceeded yet since I cannot yet capture packets.
I know the models are shown in the MiHome app if you go to the gateway device --> click ... --> about --> Hub info.
That shows a list containing dicts that have model, sid, and name, would be very usefull to get that list since we would then also have nice default names for inside HomeAsssitant while setting up the integration

miio/gateway.py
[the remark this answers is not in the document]
bskaplou: `Could you explain what you can do with loading such a "program" into gateway memory?`
Gateway is able to store limited amount of such "scripts". Scripts are permanent upon restarts. External device call parameters are also encoded with script (i.e. rotate)...
For each devices receiving messages from scripts gateway maintains state... It means gateway is sending ping requests and will not send actual packages in pings are faulty..

`Can you only get callbacks when an event happens/property changes or can it do more?`
Unfortunately we have no callbacks... Script allows to send package to other miio device as a result of action....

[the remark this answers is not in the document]
bskaplou: Subscription mechanism can be implement by scripts if you use target with ip is an ip of python-miio running device... But in xiaomi scripts are not expected to be used as callbacks. It was designed to make gateway implement complete automation listen=>process=>send data to other devices... Such behavior is still possbile. You can create script which turns on wi-fi power plug in cube rotation without processing of callback in hass.

bskaplou:
Significantly updated...

starkillerOG:
How do you sniff miio trafic?

starkillerOG:
I think this looks very good!

Could you possibly make an script for the gateway alarm?
I saw that there is an automation to trigger stuff if the Alarm goes off

bskaplou:
@starkillerOG 
I use rt-ac68u asus router to sniff traffic with asuswrt-merlin firmware. 
Steps:
 1. Install asuswrt-merlin into router
 2. Install Entware in SD card of router
 3. Install tcpdump in router
 4. Sniff traffic on router via ssh: tcpdump -i eth1 -w dump1 '(src 192.168.2.109 or dst 192.168.2.109)' . 109 is my router
 5. Copy file to laptop with scp
 6. Convert dump into json: tshark -r dump1 -T json >dump1.json
 7. Decrypt data with nodejs miio tool miio protocol json-dump dump1.json --token XXX
 8. All done :)

bskaplou:
@rytilahti is everything fine with PR?

starkillerOG:
@bskaplou great work with this code!

In the mean time I have been able to sniff traffic and test the code.
I was able to sniff trafic using BlueStack (an android emulator for windows), install MiHome App and then capture packets using wireshark --> export as json and use the miio tool https://github.com/aholstenson/miio to decode the packets.

Your code worked mostly nicely and I was able to capture the magnet sensor open event.

The two major things that I think still need to improve in this code are:
- the tokens dict in the gateway_scripts.py file
  - the 'real' schould use the token of the gateway class
  - there realy needs to be a method to obtain/use encoded token and data_tkn, otherwise this code will be unusable for HomeAssistant (otherwise everyone would need to first sniff traffic to figure out these codes, which I think is too much to ask for HomeAssistant users)
- the fake_device.py file schould be implemented such that it can be called by HomeAssistant or inside a simple python script with a non blocking method to call the start server and stop server.
  - maybe consider a asyncio server (see code example below)
  - you could also look at https://github.com/Danielhiversen/PyXiaomiGateway/blob/master/xiaomi_gateway/__init__.py for inspiration with the "listen" and "stop_listen" functions inside the XiaomiGatewayDiscovery class


```
import asyncio

async def handle_message(reader, writer):
    data = await reader.read(1000)
    message = data.decode()
    if message:
        addr = writer.get_extra_info('peername')
        print("Received %r from %r" % (message, addr))

    #writer.write(data)
    #await writer.drain()

    #print("Close the client socket")
    #writer.close()

loop = asyncio.get_event_loop()
coro = asyncio.start_server(handle_message, IP, port, loop=loop)
server = loop.run_until_complete(coro)

# Serve requests until Ctrl+C is pressed
print('Serving on {}'.format(server.sockets[0].getsockname()))
try:
    loop.run_forever()
except KeyboardInterrupt:
    pass

# Close the server
server.close()
loop.run_until_complete(server.wait_closed())
loop.close()
```

bskaplou:
@starkillerOG removed token from code... passed as cli args now... 
FakeDevice is not expected to be directly called from hass. Fake device is simple example which gives the idea.  .... hass will have it's own async cycle for sure...

starkillerOG:
Homeassistant will almost certanly require the server/fake_device code to be implemented in miio library instead of in HASS itself, since they always want to push as much as code possible to the libs.

Besides it is convenient to have the server code in miio lib because then things like open/closed state can be tracked internally in the miio lib.
There will just need to be a way to register an external callback function such that HASS can fire events when for instance a remote button is pressed.

@rytilahti what do you think?

miio/gateway.py
```diff
@@ -928,6 +1001,67 @@ def update(self):
         self._props.status_ch1 = values[1]
         self._props.load_power = values[2]
 
+class Cube(SubDevice):
+    """Subdevice Cube specific properties and methods"""
+
+    properties = []
+
+    @command()
+    def install_move_script(self, encoded_token):
```
rytilahti, line 1010: All these installs should probably be just a single method in `SubDevice`: `install_event_handler` or something like that, that would take 1) device, 2) event, 3) event receiver, and 4) method to be called.

The subdevices themselves should have a class attribute defining the available events. So in the case of the cube, the Cube class should define that it has events:
```
@property
def events():
    return ["tap_twice", "shake"]
```
The users of the lib can then request available events, and create a script for those.

miio/gateway.py
```diff
@@ -744,6 +802,21 @@ def get_firmware_version(self) -> Optional[int]:
...
+                lambda action: (
+                    action,
+                    (
+                        action_id[action](self.sid),
+                        self._gw.x_del(action_id[action](self.sid)),
+                    ),
+                ),
+                action_id.keys(),
+            )
+        )
```
rytilahti, line 818: This is way coo complicated and rather hard to interpret.

Is the goal to remove all scripts? Is there a way to get a list of current events, that could be iterated and then removed? How does removal/overwriting of a single script look like?

How is the encoded token used?

miio/gateway.py
```diff
@@ -207,6 +224,47 @@ def discover_devices(self):
...
+        if len(target) < 1:
+            return f"Device with sid={sid} not found"
+        elif not hasattr(target[0], command):
+            return f"Device with sid={sid} has no method {command}"
+        else:
+            return getattr(target[0], command)(encoded_token)
+
+    def install_script(self, sid, builder, encoded_token, ip=None):
+        """Install script for by building script source and sending it with miio method. You need to run fake or real device to capture script execution results."""
+        if ip is None:
```
rytilahti, line 249: I'd like to see this being much simpler, in my opinion the API should be more something like:
```
gw = Gateway()
cube = gw.get_device_by_id("cube id")
some_other_device = gw.get_device_by_id("my light")
cube.install_event_handler(cube, 'shake', some_other_device, 'some action')
```

Likewise, if I have a cube instance, I would like to query the list of available events and already installed scripts:
```
print("available events: %s" % cube.events)  # I mention this way of exposing the available events in some other comment

# if there is only one script possible, otherwise we need a way to get a list of already installed scripts to make it possible for the user to decide which one to remove
cube.remove_script("shake")  

# the other case
handlers = cube.installed_event_handlers
for handler in handers:
    print("handler %s")
```

miio/gateway.py
```diff
@@ -928,6 +1001,67 @@ def update(self):
...
+        return self._gw.install_script(self.sid, build_flip180, encoded_token)
+
+
+class AqaraSquareButton(SubDevice):
+    """Subdevice AqaraSquareButton specific properties and methods"""
+
+    properties = []
+
+    @command()
+    def install_singlepress_script(self, encoded_token):
```
rytilahti, line 1046: See my other comment about how to expose events to users.

miio/gateway_scripts.py
```diff
@@ -0,0 +1,324 @@
...
+    ]
+
+
+def build_move(
+    source_sid,
+    target_ip,
+    target_encoded_token,
+    target_model=fake_device_model,
+    target_id=fake_device_id,
+    source_model="lumi.sensor_cube.v1",
```
rytilahti, line 96: Considering that these seem to be passed everywher, I think the best approach is to create a new class, `EventHandler` could be a possible name. This would contain all the necessary information for the handler, so just to give you an idea:
```
@attr.s
class EventHandler:
    source_sid = attr.ib()
    target_ip = attr.ib()
    target_token = attr.ib()
    target_model = attr.ib()  # is this necessary?!
    target_did = attr.ib()  # should this be `target_sid`?
    target_model = attr.ib() # is this necessary?
```
Potentially a lot of these are already known by the target/source devices, so potentially passing the source "device" and target "device" should be enough for making the connections.

miio/gateway_scripts.py
```diff
@@ -0,0 +1,324 @@
...
+                        "did": source_sid,
+                        "extra": extra,
+                        "key": "event." + source_model + "." + event,
+                        "model": source_model,
+                        "src": "device",
+                        "timespan": ["0 0 * * 0,1,2,3,4,5,6", "0 0 * * 0,1,2,3,4,5,6"],
+                        "token": "",
+                    },
+                ],
+                [
```
rytilahti, line 73: This needs to be done using a container class, otherwise it is way too complicated and hard to interpret. See my comment above, I think some of this data can already be stored in the devices what we have, the event handler (or eventscript? or script? I'm not sure) needs to simply have the potential extra fields (like the `timespan`).

miio/gateway.py
bskaplou: yes... goal is to remove all scripts at least potentially could be installed in device... As far as i get list of installed scripts is not exposed through miio protocol... Here I xdel all scripts for all possible scripts to avoid memory leak on gw

Encoded token is used during script installation and somehow linked with real token. Unfortunately algo is unknown...

rytilahti: Actually, I think there were some sort of list functionality related to scenes in the gateway implementation at some point, that could be related to this?

Anyway, I pointed out the encoded_token here as it is given as a parameter but never used in the method itself?

rytilahti:
As #1446 #1459 are merged now, I think this can be closed. Thanks @bskaplou and @starkillerOG!

PR #712, opened 2020-05-29
Add next_schedule to vacuum timers

MarBra:


miio/vacuum.py
```diff
@@ -357,7 +357,7 @@ def timer(self) -> List[Timer]:
         """Return a list of timers."""
         timers = list()
         for rec in self.send("get_timer", [""]):
-            timers.append(Timer(rec))
+            timers.append(Timer(rec, timezone=self.timezone()))
```
rytilahti, line 360: Move the `self.timezone()` call out from the loop, that will otherwise cause unnecessary I/O.

miio/vacuumcontainers.py
```diff
@@ -434,6 +438,13 @@ def action(self) -> str:
         Note, this seems to be always 'start'."""
         return str(self.data[2][1])
 
+    @property
+    def next_schedule(self) -> datetime:
+        """Next schedule for the timer"""
```
rytilahti, line 443:
```suggestion
        """Next schedule for the timer."""
```

miio/vacuumcontainers.py
```diff
@@ -434,6 +438,13 @@ def action(self) -> str:
...
+        local_tz = timezone(self.timezone)
+        cron = croniter(self.cron, start_time=local_tz.localize(datetime.now()))
+        return cron.get_next(ret_type=datetime)
```
rytilahti, line 446:
```suggestion

        return cron.get_next(ret_type=datetime)
```

PR #713, opened 2020-05-30
Add new device type mappings, add note about 'used_for_public'

starkillerOG:
Not entirely sure what this command does, but I think it has to do with enabeling/disabeling developer mode.
In any case, I came accross this command while sniffing gateway trafic

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -84,7 +84,7 @@ class Gateway(Device):
     * toggle_plug
     * remove_all_bind
     * list_bind [0]
- 
+
     * self.get_prop("used_for_public") # Return the 'used_for_public' status, probably this has to do with developer mode.
     * self.set_prop("used_for_public", state) # Set the 'used_for_public' state, probably this has to do with developer mode.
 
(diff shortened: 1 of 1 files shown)


rytilahti:
I'm a bit hesitant on adding functionality that is unclear, as that can simply be consuming to both API and cli users. How about adding these to the list of known, but unknown commands in the docstring instead?

However, if you really want to have this in, then it should be a single command `used_for_public` which takes an optional value to set.

starkillerOG:
I understand
Was not planning on using it, just wanted it documented for the future.
Docstring seems like a good idea, moved it to there in the latest commit

rytilahti:
Great, agreed! One more thing, did you test those? It would also be worth describing the return values for the future.

starkillerOG:
I looked at the sniffing and saw the set and get values [1] and 1, so presumably the other value will be 0 for off, but do not dare to test because I don't know what it is going to mess up.....

anyway I added the return values to the docstring and also added 2 more device types from issue https://github.com/rytilahti/python-miio/issues/650

rytilahti:
I updated the title, this can be merged as soon as the CI passes.

starkillerOG:
@rytilahti CI has passed, so this is now ready to be merged

PR #717, opened 2020-06-04
add AqaraWallOutlet support

starkillerOG:
see issue https://github.com/rytilahti/python-miio/issues/650

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -942,3 +942,18 @@ class AqaraWallOutlet(SubDevice):
         values = self.get_property_exp(self.properties)
         self._props.status = values[0]
         self._props.load_power = values[1]
+
+    @command()
+    def toggle(self):
+        """Toggle Aqara Wall Outlet"""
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti can you also merge this?

rytilahti:
Looks good to me, but if it's fine for you, we could wait until we also know the control commands so that they can be added in a single PR?

starkillerOG:
yea that is fine, was just pushing fast to see if I could get it in 0.5.1, but since that is already out, lets indeed wait for the controll commands.

rytilahti:
Yeah, when the prepare release PR hits the master, the release is already "done" at that point. I could have waited out for a bit, but as I had some free time I thought it'd be a good idea to make a quick release to unblock some downstream PRs.

starkillerOG:
@rytilahti this has been comfirmed to work by @miway80, So this can now be merged.
Both getting status and controlling the device is implemented.

PR #719, opened 2020-06-07
add firmware property + convert device list to dict

starkillerOG:
This adds a firmware_version property to the subdevice class since HomeAssistant needs to acces the firmware_version property for the device registry.
If directly accesed like `subdevice._fw_ver` it results in a pylint error: `W0212: Access to a protected member _fw_ver of a client class (protected-access)`

I made a base implementation for subdevices in HomeAssistant that also registers all subdevices in the device registry, the PR is here: https://github.com/home-assistant/core/pull/36539/files

It also converts the self._devices from a list to a dict with as key the sid of the subdevice.
This is done because this will probably be needed once a callback meganism is implemented, the server will then need an easy way to get the correct subdevice when it knows the sid, instead of having to loop over the whole list and check the sid in this way direct key acces can be used.
This is related to https://github.com/rytilahti/python-miio/pull/709

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -640,6 +640,11 @@ class SubDevice:
         return self.type.name
 
     @property
+    def firmware_version(self):
+        """Return the firmware version."""
+        return self._fw_ver
+
+    @property
     def battery(self):
         """Return the battery level."""
         return self._battery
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti Could you please merge this and push a new version of python-miio?
I ran into the `Access to a protected member _fw_ver of a client class (protected-access)` problem while implementing HomeAssistant support.

Sorry for this, schould have tought about this when I made the restructure of the Gateway class.

PR #722, opened 2020-06-10
gateway: add model property & implement SwitchOneChannel

starkillerOG:
implements the SwitchOneChannel device with status and control thanks to @miway80

adds a model property to the subdevice base class.
This could be used for the model to display in HomeAssistant device registry (could also use device_type_name, not sure what is better)
But more important this is needed in the future to easily add subscriptions to properties using the install_scripts of @bskaplou of PR https://github.com/rytilahti/python-miio/pull/709

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -170,6 +170,12 @@ class Gateway(Device):
             DeviceType.AqaraSwitchOneChannel: AqaraSwitchOneChannel,
             DeviceType.AqaraSwitchTwoChannels: AqaraSwitchTwoChannels,
             DeviceType.AqaraWallOutlet: AqaraWallOutlet,
+            DeviceType.Cube: Cube,
+            DeviceType.AqaraSquareButton: AqaraSquareButton,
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti all checks now pass, can this be merged?

miio/gateway.py
```diff
@@ -755,6 +768,7 @@ class AqaraHT(SubDevice):
 
     accessor = "get_prop_sensor_ht"
     properties = ["temperature", "humidity", "pressure"]
+    _model = "lumi.weather.v1"
```
rytilahti, line 771: It looks odd when the model is the only one with underscore in the beginning. Should this also be simply `model`?

starkillerOG: well the property getter (line 650) is already called model so that is why I included the underscore.
Besides underscore indicates it is an internall variable in python.
But I could also simply remove the property getter and remove the underscore to directly acces the variable....

Whatever you prefer...

rytilahti: Ah, sorry, I skipped that part somehow.. No, this is good to go!

PR #724, opened 2020-06-11
Gateway: add name + model property to subdevice & add loads of subdevices

starkillerOG:
In this way we can easily define how the name of subdevices schould be set in HomeAssistant. Ideally we would at some point figure out how to get the names set in the Mi Home app...
But for now use the device_type + sid as a name

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -46,19 +46,21 @@ class DeviceType(IntEnum):
     SwitchOneChannel = 9  # lumi.ctrl_neutral1.v1
     SensorHT = 10  # lumi.sensor_ht
     Plug = 11  # lumi.plug
-    RemoteSwitchDoubleV1 = 12  # lumi.sensor_86sw2
+    RemoteSwitchDoubleV1 = 12  # lumi.sensor_86sw2.v1
+    Curtain = 13  # lumi.curtain
     RemoteSwitchSingleV1 = 14  # lumi.sensor_86sw1.v1
     SensorSmoke = 15  # lumi.sensor_smoke
-    AqaraHT = 19  # lumi.weather.v1
     AqaraWallOutletV1 = 17  # lumi.ctrl_86plug.v1
+    SensorNatgas = 18  # lumi.sensor_natgas
+    AqaraHT = 19  # lumi.weather.v1
     SwitchLiveOneChannel = 20  # lumi.ctrl_ln1
     SwitchLiveTwoChannels = 21  # lumi.ctrl_ln2
     AqaraSwitch = 51  # lumi.sensor_switch.aq2
     AqaraMotion = 52  # lumi.sensor_motion.aq2
     AqaraMagnet = 53  # lumi.sensor_magnet.aq2
+    AqaraRelayTwoChannels = 54  # lumi.relay.c2acn01
     AqaraWaterLeak = 55  # lumi.sensor_wleak.aq1
     AqaraVibration = 56  # lumi.vibration.aq1
-    AqaraRelayTwoChannels = 54  # lumi.relay.c2acn01
     AqaraSquareButtonV3 = 62  # lumi.sensor_switch.aq3
     AqaraSwitchOneChannel = 63  # lumi.ctrl_ln1.aq1
     AqaraSwitchTwoChannels = 64  # lumi.ctrl_ln2.aq1
@@ -68,8 +70,6 @@ class DeviceType(IntEnum):
     RemoteSwitchDouble = 135  # lumi.remote.b286acn01
 
 
-# 13 - lumi.curtain
-# 18 - lumi.sensor_natgas
 # 59 - lumi.lock.aq1
 # 66 - lumi.light.aqcn02
 # 68 - lumi.sensor_cube.aqgl01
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti ready to be merged

miio/gateway.py
```diff
@@ -647,6 +647,11 @@ def device_type(self):
         """Return the device type name."""
         return self.type.name
 
+    @property
+    def name(self):
+        """Return the name of the device."""
+        return f"{self.device_type}-{self.sid}"
```
rytilahti, line 653: So, I was thinking the the "name" should be something user-friendly, like a real description of the device (and maybe its component number). Something like what is shown at https://www.zigbee2mqtt.io/information/supported_devices.html#xiaomi

starkillerOG: I am not sure how we would then implement that:
Don't you think "double key wireless wall switch" and "temperature, humidity and pressure sensor" are a bit long for a name, espacially if we also include the sid...

starkillerOG: I could just make a `_name` variable for each subdevice that is implemented and we could then discuss the individual names, does that sound good?

furthermore I see the models are here also specified.
Schould I rename the current _model variable to _zigbee_id and add those models
So for the magnet it would be:
_model = "MCCGQ11LM"
_zigbee_id = "lumi.sensor_magnet.aq2" (see https://zigbee.blakadder.com/Xiaomi_MCCGQ11LM.html)

rytilahti: That's what zigbee2mqtt is using, so I don't think it's that bad:

image

There is no unique id here though, but I'm not sure that is something that belongs to the user interface anyway.

rytilahti: Separating the zigbee_id from the name/model sounds good to me. I'm not sure how much information the "model number" itself does, except that it makes it easier to google for it.

For the sensors in homeassistant we want to have some sort of human-readable string, in my opinion at least. The rest can easily be adapted/added later on, that's why I kind of like the "nice descriptive name (model number)" format zigbee2mqtt uses.

starkillerOG: I will start on this tommorow.
However having the sid in the name is very usefull if you have multiple of the same devices.
In my case I have 3 temperature sensors and I need to figure out which one is which. I can do that using the sid because the sid can be found in the MiHome app where I originally linked the devices and gave them proper names so I know which one is which.

Besides these are only the default names, 95% of the users will change the name anyway once they set up the integration.

rytilahti: Ah, indeed, I can see definitely how that can be useful. Another idea, could the sid be exposed as a separate attribute in homeassistant instead of embedding it into the name?

I don't really care that much as I'm not personally using this, but I think having "Aqara Temperature sensor ... (sid)" is friendlier than "AqaraHT-123123123213". I agree, that the sid should somehow be exposed especially as it can be useful with the app as you mentioned.

starkillerOG: that would be ideal but currently the only info that can be shown is:
manfacturer
model
fw_ver
connected_via_device

I would defenitly like to expose more attributes but that is currently not possible.
What I would like:
serial_number (I would use sid for xiaomi_miio)
protocol_version (not for xiaomi_miio)
zigbee_model
mac_adress (not for xiaomi_miio)
ip_adress

All of those would be nice to have under device_info...
But that is not going to happen any time soon

starkillerOG:
@rytilahti Is this alright now?

miio/gateway.py
```diff
@@ -647,11 +689,23 @@ def device_type(self):
         """Return the device type name."""
         return self.type.name
 
+    @property
+    def name(self):
+        """Return the name of the device."""
+        if self._name == "unknown":
+            self._name = self.device_type
```
rytilahti, line 696: I don't think it's worth to have this check here in place. When someone notices that there is an "Unknown (some sid)", I hope they'll report back so that can be added to the library.

rytilahti:
One comment about the most recent commit, I think we need to add some sort of warning log indicating that the device has been detected but is not implemented (based on missing properties, or somehow else in case the device has no properties at all?). Otherwise this will give us bug reports of correctly detected but not supported devices.

miio/gateway.py
starkillerOG: I have removed that check

starkillerOG:
> 
> 
> One comment about the most recent commit, I think we need to add some sort of warning log indicating that the device has been detected but is not implemented (based on missing properties, or somehow else in case the device has no properties at all?). Otherwise this will cause bug reports of correctly detected but not yet supported devices.

Well those devices will log an debug message when the update method is called if that has not been specified by the device. And these devices will not have enteties so I think we will be fine.

In princeple we also want people that have a device that is not supported yet to report here so they can figure out which properties the device has and then implement those.
I am thinking of maybe having a general step by step guide how to implement support for a subdevice that we can reffer people to?

starkillerOG:
I reduced the not implemented warning to log level because there are tons of devices like motion sensor, door sensor, remote switches, buttons and cube that do not have any properties, they only provide events. And since event support requires subscriptions that is still quite far away.

So I don't think it is fair to bombard people with warnings of not implemented devices since we don't have a way to implement them yet before the subscription PR (https://github.com/rytilahti/python-miio/pull/709) is done.

starkillerOG:
@rytilahti I think this is now ready to be merged.
Could you have a look at it?
(at least I am done with adding subdevices...)

rytilahti:
> I am thinking of maybe having a general step by step guide how to implement support for a subdevice that we can reffer people to?

Yes, having a guide how to add support for new devices would be great!

> I reduced the not implemented warning to log level because there are tons of devices like motion sensor, door sensor, remote switches, buttons and cube that do not have any properties, they only provide events.

We need a way to differentiate between "no known properties" and "no properties" to avoid useless warnings, but this can be also done separately from this PR.

miio/gateway.py
```diff
@@ -906,154 +1175,325 @@ def update(self):
...
+class LockV1(SubDevice):
+    """Subdevice LockV1 specific properties and methods."""
+
+    properties = []
+    _zigbee_model = "lumi.lock.v1"
+    _model = "A6121"
+    _name = "Vima cylinder lock"
+
+
+class IkeaBulb82(SubDevice):
```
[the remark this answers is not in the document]
starkillerOG: Yea the gateway.py file grew a lot and is getting way too big.
Was already thinking the same thing, at least all subdevices schould be moved to a gateway_subdevices.py file or something....

for the IkeaBulbs and simular classes, I think it would be best to define a new class `IkeaBulb` that inherits properties from `SubDevice` class, then all individual model classes like `IkeaBulb82` can inherit all common functions from the `IkeaBulb` class.

Not sure if that is what you were trying to say....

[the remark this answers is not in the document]
starkillerOG: Maybe split up all subdevices in files which have one common class that they inherit from. so all IkeaBulbs in one file, all cubes in one, all buttons in one etc.

starkillerOG:
@rytilahti could you please push to a new version of python-miio?
I could then finish the HomeAssistant subdevice implementation

miio/gateway.py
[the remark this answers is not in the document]
rytilahti: Yes, having a common base class would make it much cleaner already. I don't think having a `gateway_subdevices.py` is a good idea, it's much better to have a structure like:
```
miio/gateway/
             gateway.py
             gateway_zigbee.py
             gateway_light.py
             gateway_radio.py
```
and so on. To extend that example, here's another idea that would make it much nicer:
```
miio/gateway/
            gateway.py
            subdevice.py
miio/gateway/aqara/
                  cube.py
                  switch.py
miio/gateway/ikea/
                  bulb.py
```

rytilahti:
I'll see if I can find some time tomorrow to prepare a release, but I can't make any promises on that.

Issue #295, opened 2018-04-02
Extend the Air Conditioning Companion support


syssi:
```
get_scan_result []
-> [0, 0]

get_clock
-> {'Min': 30, 'Day': 31, 'Enable': 0, 'Hour': 7, 'MusicTime': 60, 'ClockVol': 60, 'Music': 20}

get_default_music '[0]'
['0']
get_default_music '[1]'
['10']
get_default_music '[2]'
['20']
get_default_music '[3]'
['30']
get_default_music '[4]'
['40']
get_default_music '[5]'
['0']
get_default_music '[6]'
['0']
get_default_music '[7]'
['0']
get_default_music '[8]'
['0']
...

get_ir_learn_result []
-> ['(null)']

get_music_info [0]
-> {'list': [{'index': 0, 'mid': '0', 'time': 1}, {'index': 1, 'mid': '1', 'time': 1}, {'index': 2, 'mid': '2', 'time': 1}, {'index': 3, 'mid': '3', 'time': 1}, {'index': 4, 'mid': '4', 'time': 1}, {'index': 5, 'mid': '5', 'time': 1}, {'index': 6, 'mid': '6', 'time': 1}, {'index': 7, 'mid': '7', 'time': 1}, {'index': 8, 'mid': '8', 'time': 1}], 'total': 9, 'default': '0'}

get_music_info '[1]'
-> {'total': 4, 'list': [{'mid': '10', 'time': 1, 'index': 10}, {'mid': '11', 'time': 1, 'index': 11}, {'mid': '12', 'time': 1, 'index': 12}, {'mid': '13', 'time': 1, 'index': 13}], 'default': '10'}

get_music_info '[2]'
-> {'default': '20', 'total': 10, 'list': [{'time': 1, 'index': 20, 'mid': '20'}, {'time': 1, 'index': 21, 'mid': '21'}, {'time': 1, 'index': 22, 'mid': '22'}, {'time': 1, 'index': 23, 'mid': '23'}, {'time': 1, 'index': 24, 'mid': '24'}, {'time': 1, 'index': 25, 'mid': '25'}, {'time': 1, 'index': 26, 'mid': '26'}, {'time': 1, 'index': 27, 'mid': '27'}, {'time': 1, 'index': 28, 'mid': '28'}, {'time': 1, 'index': 29, 'mid': '29'}]}

get_music_info '[3]'
-> {'total': 0, 'list': [], 'default': -1}

get_music_info '[4]'
[...]

PR #729, opened 2020-06-15
Add AqaraSmartBulbE27 support

starkillerOG:
Also add untested bind commands to docstring from https://github.com/rytilahti/python-miio/issues/295

Add AqaraSmartBulbE27 support from https://github.com/igzero/xiaomigateway

Note that the `self._props.color_temp = self.send("get_ct").pop()` is untested but that seemed to be a very good guess as:
`set_ct`
`set_bright`
`get_bright`
are implemented in https://github.com/igzero/xiaomigateway

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -1330,9 +1330,9 @@ class AqaraSmartBulbE27(SubDevice):
         self._props.brightness = self.send("get_bright").pop()
         self._props.color_temp = self.send("get_ct").pop()
         if self._props.brightness > 0 and self._props.brightness <= 100:
-            self._props.status = 'on'
+            self._props.status = "on"
         else:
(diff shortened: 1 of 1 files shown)


starkillerOG:
@igzero Could you test this code?
If so it will be straigt forward to add the bulb to HomeAssistant in the official xiaomi_miio component.

starkillerOG:
@rytilahti can you merge this?
All tests pass (except coverage)

starkillerOG:
@mouth4war did you test this code?

rytilahti:
Sure thing! I think we should start adding some tests for the gateway to avoid breaking it inadverently in the future. That's probably best done in sync with refactoring the file into its own package as discussed earlier.

PR #731, opened 2020-06-16
Moved access to discover message attribute inside 'if message is not None' statement

jthure:
If `m` is None, which happens if discovery fails, then `header = m.header.value` is failing. It should be inside the check that `m` is not None


rytilahti:
Indeed, well spotted, thanks for the PR! :+1:

Issue #728, opened 2020-06-13
Add support for `lumi.gateway.mieu01` subdevices


javicalle:
The EU version of the gateway is supported in the current versión of the library and  can be intergated in HA.

Some tests had been made to evaluate subdevices resolution with no success:
```
(venv)$ python /tmp/test_miio_gateway/gateway_devices.py
Traceback (most recent call last):
  File "/tmp/test_miio_gateway/gateway_devices.py", line 5, in <module>
    gateway.discover_devices()
  File "/opt/DevelopmentSpace/hassio/home-assistant/venv/lib/python3.7/site-packages/miio/gateway.py", line 168, in discover_devices
    devices_raw = self.get_prop("device_list")
  File "/opt/DevelopmentSpace/hassio/home-assistant/venv/lib/python3.7/site-packages/miio/gateway.py", line 207, in get_prop
    return self.send("get_device_prop", ["lumi.0", property])
  File "/opt/DevelopmentSpace/hassio/home-assistant/venv/lib/python3.7/site-packages/miio/device.py", line 147, in send
    command, parameters, retry_count, extra_parameters=extra_parameters
  File "/opt/DevelopmentSpace/hassio/home-assistant/venv/lib/python3.7/site-packages/miio/miioprotocol.py", line 203, in send
    self._handle_error(payload["error"])
  File "/opt/DevelopmentSpace/hassio/home-assistant/venv/lib/python3.7/site-packages/miio/miioprotocol.py", line 263, in _handle_error
    raise DeviceError(error)
miio.exceptions.DeviceError: {'code': -5005, 'message': 'params error'}
```

where `gateway_devices.py`:
```python
from miio import Gateway

gateway = Gateway("192.168.1.IP", "tokentokentoken")

gateway.discover_devices()
devices = gateway.devices

for dev in devices:
    dev.update()
    print(dev)
```
@starkillerOG sugest (home-assistant/core#22078) a new approach with the `get_device_list` command.

starkillerOG:
@javicalle alright lets see if we can get this working:

1) try testing this code as a start:
```
from miio import Gateway
gateway = Gateway("192.168.1.IP", "tokentokentoken")
print(gateway.info())
```
That will print basic information about your gateway and that will confirm that the communication is okay. (although since it works with the HomeAssistant integration this schould not be a problem)

[...]

PR #732, opened 2020-06-16
prevent errors on "lumi.gateway.mieu01"

starkillerOG:
see discussion in https://github.com/rytilahti/python-miio/issues/728

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -1,8 +1,8 @@
 """Xiaomi Aqara Gateway implementation using Miio protecol."""
 
+import logging
 from datetime import datetime
 from enum import Enum, IntEnum
-import logging
 from typing import Optional
 
 import attr
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -16,7 +16,7 @@ from .utils import brightness_and_color_to_int, int_to_brightness, int_to_rgb
 _LOGGER = logging.getLogger(__name__)
 
 GATEWAY_MODEL_CHINA = "lumi.gateway.v3"
-GATEWAY_MODEL_EU ="lumi.gateway.mieu01"
+GATEWAY_MODEL_EU = "lumi.gateway.mieu01"
 GATEWAY_MODEL_ZIG3 = "lumi.gateway.mgl03"
 GATEWAY_MODEL_AQARA = "lumi.gateway.aqhm01"
 
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -265,6 +265,9 @@ class Gateway(Device):
 
         # Skip the models which do not support getting the device list
         if self.model == "lumi.gateway.mieu01":
+            _LOGGER.info(
+                "Gateway model '%s' does not (yet) support getting the device list", self.model,
+            )
             return self._devices
 
         devices_raw = self.get_prop("device_list")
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti can you also merge this?
All tests pass (except coverage)

miio/gateway.py
```diff
@@ -252,9 +261,14 @@ def discover_devices(self):
             DeviceType.D1WallSwitchTripleNN: D1WallSwitchTripleNN,
             DeviceType.ThermostatS2: ThermostatS2,
         }
-        devices_raw = self.get_prop("device_list")
         self._devices = {}
 
+        # Check if this gateway does not supports getting the device_list
+        if self.model == "lumi.gateway.mieu01":
+            return self._devices
```
rytilahti, line 268: Should this also log a warning?

miio/gateway.py
```diff
@@ -252,9 +261,14 @@ def discover_devices(self):
             DeviceType.D1WallSwitchTripleNN: D1WallSwitchTripleNN,
             DeviceType.ThermostatS2: ThermostatS2,
         }
-        devices_raw = self.get_prop("device_list")
         self._devices = {}
 
+        # Check if this gateway does not supports getting the device_list
```
rytilahti, line 266:
```suggestion
        # Skip the models which do not support getting the device list
```
Maybe?

miio/gateway.py
starkillerOG: I added the warning

starkillerOG:
@rytilahti thanks for your feedback!
I think this is now ready to merge.

miio/gateway.py
```diff
@@ -847,9 +872,25 @@ def unpair(self):
     @command()
     def get_battery(self):
         """Update the battery level and return the new value."""
-        self._battery = self.send("get_battery").pop()
+        if self._gw.model != "lumi.gateway.mieu01":
```
rytilahti, line 875: The docstring seems to be incorrect, I think it should be something like "Return the battery level, if available". It is also better not to check against the model number (or at least create a constant for that), but try to fetch the state and return None (and log a warning) if it fails. Or that's my opinion :-)

starkillerOG: @rytilahti I have updated the docstrings and made constants for the gateway models I know.

I would like to check this against the gateway_model because the battery level and voltage are used in the `__repr__` function, since the model is stored you only need 1x communication with the gateway to get the model but for all subsequent subdevices the model is already stored in memory such that you don't have aditional communication with the gateway while we already know that it is going to fail...

Could also make a dedicated variable like _supports_battery that will be set to falls after the first try fails. In that way we can catch the error withouth checking for the gateway_model and still limit the gateway communication to 1x.

What do you prefer?

miio/gateway.py
```diff
@@ -252,9 +265,18 @@ def discover_devices(self):
             DeviceType.D1WallSwitchTripleNN: D1WallSwitchTripleNN,
             DeviceType.ThermostatS2: ThermostatS2,
         }
-        devices_raw = self.get_prop("device_list")
         self._devices = {}
 
+        # Skip the models which do not support getting the device list
+        if self.model == GATEWAY_MODEL_EU:
+            _LOGGER.info(
```
rytilahti, line 272: Use `warning` instead of `info` for these warnings, this is otherwise good to go.

starkillerOG: The internall method will not help much becuase you would still need to check for the return value of that method and then have some specific error message to specify which function is not supported.
So do not think adding such method will actually reduce the amount of code...

rytilahti: Fair enough, let's merge this then!

PR #735, opened 2020-06-24
gateway: cleanup SensorHT and Plug class

starkillerOG:
Remove properties that do not apply to these devices


rytilahti:
LGTM :+1:

PR #737, opened 2020-06-26
Gateway: Add control commands to Plug

starkillerOG:
see: https://github.com/rytilahti/python-miio/issues/650#issuecomment-649782979 from @wuid24


rytilahti:
LGTM :+1:

PR #741, opened 2020-06-30
fan: Ability to disable delayed turn off functionality

insajd:
When 0 is passed, delay turn off is disabled. Before this, if delay off was set, it could not be unset.


rytilahti:
LGTM after the tests pass. What do you think @syssi?

syssi:
The tests should be updated, too.

insajd:
Thanks guys!

PR #751, opened 2020-07-04
chuangmi_camera: Improve home monitoring support

impankratov:
General:
- fix error with `rotate` command (`click.argument` name wasn't matching method's argument name).

`chuangmi.camera.ipc009` specific (maybe will work for other models too):
- add motion sensitivity
- add monitoring configuration
- add methods for NAS configuration (get/set config, clear directory)

### DIFF
--- a/miio/chuangmi_camera.py
+++ b/miio/chuangmi_camera.py
@@ -34,6 +34,14 @@ class MotionDetectionSensitivity(enum.Enum):
     Low = 1
 
 
+class HomeMonitoringMode(enum.IntEnum):
+    """Home monitoring mode."""
+
+    Off = 0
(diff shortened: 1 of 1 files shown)


miio/chuangmi_camera.py
```diff
@@ -12,6 +12,76 @@
...
+    3,
+    3,
+    3,
+    3,
+    3,
+    3,
+    3,
+    3,
+    3,
+]
```
rytilahti, line 48: This is already defined as an enum below, so better simply use something like `[enumvalue] * times` to create the array when needed.

impankratov: Thanks for the advice, fixed it.

PR #753, opened 2020-07-06
AirPurifier MIoT: round temperature 

petrkotek:
AirPurifier MIoT: round temperature to one decimal point to return value such as `20.8 °C` instead of `20.799999 °C`.

image

This is either encountered only for certain temperatures or was caused by an update to new firmware (2.0.5.0009)


rytilahti:
Thanks! :+1:

PR #763, opened 2020-07-18
Resume zoned clean from error state

r4nd0mbr1ck:
Resume zoned clean if currently in zoned clean and in error state. 

Right now, zoned clean resume is only supported when in paused state. This should also occur in the error state.


rytilahti:
LGTM, thanks :+1:

PR #769, opened 2020-07-19
add "lumi.acpartner.v3" since it also works with this code

starkillerOG:


starkillerOG:
@rytilahti I was also supprised that the AC partner supported the same API.
But according to @shenxn it supports both the alarm feature and connected zigbee subdevices (see discussion here: https://github.com/home-assistant/core/pull/37959)

starkillerOG:
@rytilahti not sure what the best way is for the AC partner......
Maybe that needs to be done after a refactor of the gateway.py file in which we split out that file as discussed earlier into a seperate folder with seperate files for each subdevice.....
Maybe all of the code of miio/airconditioningcompanion.py could then be placed as a part of the gateway just like the alarm, light etc.

shenxn:
Since the `AirConditioningCompanion` and `Gateway` are reading different properties, keep them in different classes should only double the handshake (discovery), which should only happen once. So I think it is acceptable for now. It is better to do some refactor, though. But I haven't came up with a good way to do this.

rytilahti:
Okay, fine by me. Maybe due time the specific gateway features are contained inside their own implementation files (e.g., alarm, light) that can be re-used on other gateways. I remember reading long time ago from somewhere that the gateways and ac partners share quite a bit of implementation which may not be that surprising as they are both developed by "lumi".

PR #770, opened 2020-07-19
improve gateway light class

starkillerOG:
cleanup the gateway light class and add missing functionality needed for HomeAssistint

- added docstrings to explain all functionality.
- now contains these functions:
  - rgb_status
  - night_light_status
  - set_rgb
  - set_night_light
  - set_rgb_brightness
  - set_night_light_brightness
  - set_rgb_color
  - set_night_light_color
  - set_rgb_using_name
  - set_night_light_using_name

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -698,7 +698,7 @@ class GatewayLight(GatewayDevice):
         click.argument("brightness", type=int),
         click.argument("rgb", type=(int, int, int)),
     )
-    def set_night_light(self, brightness: int, rgb:(int, int, int)):
+    def set_night_light(self, brightness: int, rgb: (int, int, int)):
         """Set gateway night light using brightness and rgb tuple."""
         brightness_and_color = brightness_and_color_to_int(brightness, rgb)
 
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -3,7 +3,7 @@
 import logging
 from datetime import datetime
 from enum import Enum, IntEnum
-from typing import Optional
+from typing import Optional, Tuple
 
 import attr
 import click
@@ -688,7 +688,7 @@ class GatewayLight(GatewayDevice):
         click.argument("brightness", type=int),
         click.argument("rgb", type=(int, int, int)),
     )
-    def set_rgb(self, brightness: int, rgb: (int, int, int)):
+    def set_rgb(self, brightness: int, rgb: Tuple[int, int, int]):
         """Set gateway light using brightness and rgb tuple."""
         brightness_and_color = brightness_and_color_to_int(brightness, rgb)
 
@@ -698,7 +698,7 @@ class GatewayLight(GatewayDevice):
         click.argument("brightness", type=int),
         click.argument("rgb", type=(int, int, int)),
     )
-    def set_night_light(self, brightness: int, rgb: (int, int, int)):
+    def set_night_light(self, brightness: int, rgb: Tuple[int, int, int]):
         """Set gateway night light using brightness and rgb tuple."""
         brightness_and_color = brightness_and_color_to_int(brightness, rgb)
 
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -646,14 +646,79 @@ class GatewayLight(GatewayDevice):
     """Light controls for the gateway."""
 
     @command()
-    def get_night_light_rgb(self):
-        """Unknown."""
-        # Returns 0 when light is off?"""
-        # looks like this is the same as get_rgb
-        # id': 65064, 'method': 'set_night_light_rgb', 'params': [419407616]}
-        # {'method': 'props', 'params':
-        # {'light': 'on', 'from.light': '4,,,'}, 'id': 88457} ?!
-        return self._gateway.send("get_night_light_rgb")
+    def rgb_status(self):
+        """Get current status of the light (always correct)."""
+        # Returns {"is_on": false, "brightness": 0, "rgb": (0, 0, 0)} when light is off
+        state_int = self._gateway.send("get_rgb").pop()
+        brightness = int_to_brightness(state_int)
+        rgb = int_to_rgb(state_int)
+        if brightness > 0:
+            is_on = True
+        else:
+            is_on = False
+        return {"is_on": is_on, "brightness": brightness, "rgb": rgb}
+
+    @command()
+    def night_light_status(self):
+        """Get status of the night light."""
+        # The night light controls the same LEDs as the rgb light, but has a seprate memory for brightness and color.
+        # Changing the rgb light does not effect the stored state of the night light, while changing the night light does effect the state of the rgb light.
+        # Therefore this command only gives the correct status of the LEDs if the last command was a night light command and not a rgb light command, otherwise it gives the stored values of the night light.
+        # Returns {"is_on": false, "brightness": 0, "rgb": (0, 0, 0)} when night light is off
+        state_int = self._gateway.send("get_night_light_rgb").pop()
+        brightness = int_to_brightness(state_int)
+        rgb = int_to_rgb(state_int)
+        if brightness > 0:
+            is_on = True
+        else:
+            is_on = False
+        return {"is_on": is_on, "brightness": brightness, "rgb": rgb}
+
+    @command(
+        click.argument("brightness", type=int),
+        click.argument("rgb", type=(int, int, int)),
+    )
+    def set_rgb(self, brightness, rgb):
+        """Set gateway light using brightness and rgb tuple."""
+        brightness_and_color = brightness_and_color_to_int(brightness, rgb)
+        return self._gateway.send("set_rgb", [brightness_and_color])
+
+    @command(
+        click.argument("brightness", type=int),
+        click.argument("rgb", type=(int, int, int)),
+    )
+    def set_night_light(self, brightness, rgb):
+        """Set gateway night light using brightness and rgb tuple."""
+        brightness_and_color = brightness_and_color_to_int(brightness, rgb)
+        return self._gateway.send("set_night_light_rgb", [brightness_and_color])
+
+    @command(click.argument("brightness", type=int))
+    def set_rgb_brightness(self, brightness):
+        """Set gateway light brightness (0-100)."""
+        if 100 < brightness < 0:
+            raise Exception("Brightness must be between 0 and 100")
+        current_color = self.rgb_status()["rgb"]
+        return self.set_rgb(brightness, current_color)
+
+    @command(click.argument("brightness", type=int))
+    def set_night_light_brightness(self, brightness):
+        """Set night light brightness (0-100)."""
+        if 100 < brightness < 0:
+            raise Exception("Brightness must be between 0 and 100")
+        current_color = self.night_light_status()["rgb"]
+        return self.set_night_light(brightness, current_color)
+
+    @command(click.argument("color_name", type=str))
+    def set_rgb_color(self, color_name):
+        """Set gateway light color using color name (red, green, etc)."""
+        if color_name not in color_map.keys():
+            raise Exception(
+                "Cannot find {color} in {colors}".format(
+                    color=color_name, colors=color_map.keys()
+                )
+            )
+        current_brightness = self.rgb_status()["brightness"]
+        return self.set_rgb(current_brightness, color_map[color_name])
 
     @command(click.argument("color_name", type=str))
     def set_night_light_color(self, color_name):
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti I just fully tested this code and works just fine.
I think this can be merged.

miio/gateway.py
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
-        """Unknown."""
-        # Returns 0 when light is off?"""
-        # looks like this is the same as get_rgb
-        # id': 65064, 'method': 'set_night_light_rgb', 'params': [419407616]}
-        # {'method': 'props', 'params':
-        # {'light': 'on', 'from.light': '4,,,'}, 'id': 88457} ?!
-        return self._gateway.send("get_night_light_rgb")
+    def rgb_status(self):
+        """Get current status of the light (always correct)."""
+        # Returns {"is_on": false, "brightness": 0, "rgb": (0, 0, 0)} when light is off
```
rytilahti, line 651: I think this commentary could go directly to the docstring:
```
"""Get current status of the light.

Example:
   {"is_on": false, "brightness": 0, "rgb": (0, 0, 0)}
"""

It is also unclear what does "always correct" mean in this context, so that should be clarified (or removed, if unnecessary).

File: miio/gateway.py [hunk 2]
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
+        else:
+            is_on = False
+        return {"is_on": is_on, "brightness": brightness, "rgb": rgb}
+
+    @command()
+    def night_light_status(self):
+        """Get status of the night light."""
+        # The night light controls the same LEDs as the rgb light, but has a seprate memory for brightness and color.
+        # Changing the rgb light does not effect the stored state of the night light, while changing the night light does effect the state of the rgb light.
+        # Therefore this command only gives the correct status of the LEDs if the last command was a night light command and not a rgb light command, otherwise it gives the stored values of the night light.
```
rytilahti, line 666: Include this description also in the docstring, that way it is visible in the apidocs.

miio/gateway.py
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
+    def rgb_status(self):
+        """Get current status of the light (always correct)."""
+        # Returns {"is_on": false, "brightness": 0, "rgb": (0, 0, 0)} when light is off
+        state_int = self._gateway.send("get_rgb").pop()
+        brightness = int_to_brightness(state_int)
+        rgb = int_to_rgb(state_int)
+        if brightness > 0:
+            is_on = True
+        else:
+            is_on = False
```
rytilahti, line 658: This can be simplified by using `is_on = brightness > 0`. Same also below for the night light.

miio/gateway.py
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
+        # Therefore this command only gives the correct status of the LEDs if the last command was a night light command and not a rgb light command, otherwise it gives the stored values of the night light.
+        # Returns {"is_on": false, "brightness": 0, "rgb": (0, 0, 0)} when night light is off
+        state_int = self._gateway.send("get_night_light_rgb").pop()
+        brightness = int_to_brightness(state_int)
+        rgb = int_to_rgb(state_int)
+        if brightness > 0:
+            is_on = True
+        else:
+            is_on = False
+        return {"is_on": is_on, "brightness": brightness, "rgb": rgb}
```
rytilahti, line 675: Add a newline before the return to make the code more readable.

miio/gateway.py
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
+
+    @command(
+        click.argument("brightness", type=int),
+        click.argument("rgb", type=(int, int, int)),
+    )
+    def set_rgb(self, brightness, rgb):
```
rytilahti, line 681:
```suggestion
    def set_rgb(self, brightness: int, rgb: Tuple[int, int, int]):
```

rytilahti: Type annotations should also be added to other other functions.

miio/gateway.py
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
+        """Set night light brightness (0-100)."""
+        if 100 < brightness < 0:
+            raise Exception("Brightness must be between 0 and 100")
+        current_color = self.night_light_status()["rgb"]
+        return self.set_night_light(brightness, current_color)
 
     @command(click.argument("color_name", type=str))
-    def set_night_light_color(self, color_name):
-        """Set night light color using color name (red, green, etc)."""
+    def set_rgb_color(self, color_name):
```
rytilahti, line 712: rgb consists of three separate components (r, g, and b), so shouldn't this rather be `set_color`?

miio/gateway.py
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
+        """Set gateway light color using color name (red, green, etc)."""
```
rytilahti, line 713: Is the list of colors static and are the valid values known? Maybe this should be enforced using an enum (or a list)?

miio/gateway.py
```diff
@@ -646,71 +646,112 @@ class GatewayLight(GatewayDevice):
...
-        )
-        return self._gateway.send("set_night_light_rgb", [brightness_and_color])
+        current_brightness = self.rgb_status()["brightness"]
+        return self.set_rgb(current_brightness, color_map[color_name])
 
     @command(click.argument("color_name", type=str))
-    def set_color(self, color_name):
-        """Set gateway lamp color using color name (red, green, etc)."""
+    def set_night_light_color(self, color_name):
+        """Set night light color using color name (red, green, etc)."""
```
rytilahti, line 725: See my comment above wrt color names.

miio/gateway.py
starkillerOG: I wanted to be consistant between "rgb" and "night_light"
All functions begin with `set_rgb` or `set_night_light`

So I think this is the correct name

miio/gateway.py
starkillerOG: done, you are right!

miio/gateway.py
starkillerOG: done

miio/gateway.py
starkillerOG: done, thanks!

miio/gateway.py
starkillerOG: I added newlines to all functions before the retrun's

miio/gateway.py
starkillerOG: thanks, did it for all functions of this PR

miio/gateway.py
starkillerOG: This is a static dict that is defined on line 26:
https://github.com/rytilahti/python-miio/blob/79e61fbba512a79b569fc5cf5d14a2b5e7730c01/miio/gateway.py#L26

I actually did not program this and it was already in the code like this.
This will not be used for HomeAssistant since there you just set RGB values and HomeAssistant has its own color names defined.
I don't think these functions will be used much.
basically you can now specify "red", "green", "blue", "white", "yellow", "orange", "aqua", "olive" or "purple"
I will put it in the docstrings

starkillerOG:
@rytilahti any more suggestions?

miio/gateway.py
rytilahti: Ah, okay. Do not add them to the docstring, but add a comment saying that `color_map` variable in the source holds the valid values. This avoids repeating the list :-)

miio/gateway.py
rytilahti: Fair enough!

miio/gateway.py
```diff
@@ -647,69 +647,79 @@ class GatewayLight(GatewayDevice):
...
-        # The night light controls the same LEDs as the rgb light, but has a seprate memory for brightness and color.
-        # Changing the rgb light does not effect the stored state of the night light, while changing the night light does effect the state of the rgb light.
-        # Therefore this command only gives the correct status of the LEDs if the last command was a night light command and not a rgb light command, otherwise it gives the stored values of the night light.
-        # Returns {"is_on": false, "brightness": 0, "rgb": (0, 0, 0)} when night light is off
+        """
+        Get status of the night light.
+
+        The night light controls the same LEDs as the rgb light, but has a seprate memory for brightness and color.
+        Changing the rgb light does not effect the stored state of the night light, while changing the night light does effect the state of the rgb light.
+        Therefore this command only gives the correct status of the LEDs if the last command was a night light command and not a rgb light command, otherwise it gives the stored values of the night light.
```
rytilahti, line 672: Some sort of explanation to the class' docstring would be great, I think this could (at least partially) belong there as it description itself is not just about the night light but the whole light support.

What do you think?

miio/gateway.py
starkillerOG: Alright, I changed it

miio/gateway.py
starkillerOG: You are right!
I will change it, just a moment

starkillerOG: I made a more clear explanation in the class string, and removed the duplicate info here.

starkillerOG:
@rytilahti are the docstrings more clear now?

starkillerOG:
@rytilahti all test pass again.
Anything else I can improve on this?

miio/gateway.py
```diff
@@ -643,74 +643,132 @@ def set_default_music(self):
 
 
 class GatewayLight(GatewayDevice):
-    """Light controls for the gateway."""
+    """
+    Light controls for the gateway.
+
+    The gateway LEDs can be controlled using 'rgb' or 'night_light' methods.
+    The 'night_light' methods control the same LEDs as the 'rgb' methods, but has a seprate memory for brightness and color.
```
rytilahti, line 650:
```suggestion
    The 'night_light' methods control the same light as the 'rgb' methods, but has a separate memory for brightness and color.
```

miio/gateway.py
```diff
@@ -643,74 +643,132 @@ def set_default_music(self):
...
+    Changing the 'rgb' light does not effect the stored state of the 'night_light', while changing the 'night_light' does effect the state of the 'rgb' light.
```
rytilahti, line 651:
```suggestion
    Changing the 'rgb' light does not affect the stored state of the 'night_light', while changing the 'night_light' does effect the state of the 'rgb' light.
```

miio/gateway.py
```diff
@@ -643,74 +643,132 @@ def set_default_music(self):
...
-        # Returns 0 when light is off?"""
-        # looks like this is the same as get_rgb
-        # id': 65064, 'method': 'set_night_light_rgb', 'params': [419407616]}
-        # {'method': 'props', 'params':
-        # {'light': 'on', 'from.light': '4,,,'}, 'id': 88457} ?!
-        return self._gateway.send("get_night_light_rgb")
+    def rgb_status(self):
+        """
+        Get current status of the light.
+        Always represents the current status of the LEDs as oposed to 'night_light_status'.
```
rytilahti, line 658:
```suggestion
        Always represents the current status of the light as opposed to 'night_light_status'.
```

miio/gateway.py
```diff
@@ -643,74 +643,132 @@ def set_default_music(self):
...
+        rgb = int_to_rgb(state_int)
+        is_on = brightness > 0
+
+        return {"is_on": is_on, "brightness": brightness, "rgb": rgb}
+
+    @command()
+    def night_light_status(self):
+        """
+        Get status of the night light.
+        This command only gives the correct status of the LEDs if the last command was a 'night_light' command and not a 'rgb' light command, otherwise it gives the stored values of the 'night_light'.
```
rytilahti, line 675:
```suggestion
        This command only gives the correct status of the light if the last command was a 'night_light' command and not a 'rgb' light command, otherwise it gives the stored values of the 'night_light'.
```

Am I interpreting this correctly, that if any rgb light commands will overwrite the night light state?

miio/gateway.py
```diff
@@ -643,74 +643,132 @@ def set_default_music(self):
...
+        rgb = int_to_rgb(state_int)
+        is_on = brightness > 0
+
+        return {"is_on": is_on, "brightness": brightness, "rgb": rgb}
+
+    @command(
+        click.argument("brightness", type=int),
+        click.argument("rgb", type=(int, int, int)),
+    )
+    def set_rgb(self, brightness: int, rgb: (int, int, int)):
```
rytilahti, line 691:
```suggestion
    def set_rgb(self, brightness: int, rgb: Tuple[int, int, int]):
```

Also, you need to import `Tuple` from the typing module for this.

miio/gateway.py
```diff
@@ -643,74 +643,132 @@ def set_default_music(self):
...
+        """Set gateway light using brightness and rgb tuple."""
+        brightness_and_color = brightness_and_color_to_int(brightness, rgb)
+
+        return self._gateway.send("set_rgb", [brightness_and_color])
+
+    @command(
+        click.argument("brightness", type=int),
+        click.argument("rgb", type=(int, int, int)),
```
rytilahti, line 699: I'm not sure if such a type is accepted by click, but I think this is acceptable for now considering the cli functionality for gateway is not working anyway at the moment..

miio/gateway.py
```diff
@@ -643,74 +643,132 @@ def set_default_music(self):
...
+    )
+    def set_night_light(self, brightness: int, rgb: (int, int, int)):
```
rytilahti, line 701: See my previous comment about `Tuple`.

miio/gateway.py
starkillerOG: The other way around.
issuing ether a rgb light command or night light command will always change the LEDs to the values in the command.
the rgb_status will then be updated to the actual value of the LEDs

However if a rgb light command is sent, the night_light_status does not change (only the rgb_status changes)
If a night light command is sent, both the nigth_light_status and the rgb_status will change

miio/gateway.py
starkillerOG: Not sure about cli, I actually do not use that, so I try my best to keep everything in style with the rest, but I am not sure about how the cli works with things like this.

starkillerOG:
@rytilahti done, ready to be merged

PR #774, opened 2020-07-27
Remove labeler as it doesn't work as expected

rytilahti:
Related to https://github.com/actions/first-interaction/issues/10

Discussed in https://github.com/rytilahti/python-miio/pull/772#issuecomment-664429591


Toxblh:
As I understand the file https://github.com/rytilahti/python-miio/blob/master/.github/labeler.yml is depend on labler too, possible need to delete as well

rytilahti:
Yeah, I was thinking about leaving it for the time being in case the issue gets fixed, but considering that the file isn't that complicated to redo, let's remove it too.

PR #782, opened 2020-08-01
re-structure gateway devices

starkillerOG:
Restructure the gateway devices to put everything with simular API together.
~The next step is to use common classes for devices with simular API (to be done in this PR)~.

This PR also creates the base classes for switches and light bulbs, so all known models are now fully supported, although some models have not been tested (they are assummed to have simular API calls as devices that are closely related).

This PR also adds subdevice support for the `lumi.gateway.mgl03` and is a first step towards subdevice support for the `lumi.gateway.mieu01`


starkillerOG:
@rytilahti could you look if you agree with this structure for the gateway devices?
We have discussed this re-structuring before.

starkillerOG:
@rytilahti This is now finished, all tests pass and it now also includes all base classes for switches and light bulbs.

I can not actually test is since I don't own switches or light bulbs, but I am pretty sure it will work.
It is tested and does work with the weather sensors.

If this gets merged I can add all the light bulbs and switches to HomeAssistant.

starkillerOG:
The use of python-miio (importing) does not change with this PR, the following script will still work as it used to:
```
from miio import Gateway

gateway = Gateway("192.168.1.IP", "TokenTokenToken")

gateway.discover_devices()
devices = gateway.devices

for dev in devices.values():
    try:
        dev.update()
    except Exception as ex:
        print("got exception during update for %s: %s" % (dev.device_type, ex))
    print(dev)
```

miio/gateway/gateway_device/gateway_device.py
```diff
@@ -0,0 +1,32 @@
+"""Xiaomi Gateway device base class."""
```
rytilahti, line 1: This can be moved to top level, rename to `gatewaydevice.py` to be more consistent with naming. Underscores are rather uncommon in python names.

miio/gateway/sub_device/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+from .door_sensor import AqaraMagnet, Magnet
+from .light import (
+    AqaraSmartBulbE27,
+    IkeaBulb82,
+    IkeaBulb83,
+    IkeaBulb84,
+    IkeaBulb85,
+    IkeaBulb86,
+    IkeaBulb87,
+    IkeaBulb88,
```
rytilahti, line 16: I'm wondering how different all of these bulbs really are, i.e., could it be possible to have a single class which exposes the features specific to the bulb in question?

For the time being, I think it's fine to have it as it is as there are so many unknowns at the moment, but at some point it would greatly reduce the boilerplate code if devices could simply be defined in terms of their functionalities (maybe even in simple ini/json/yaml files)? Some food for thought for some later time..

miio/gateway/sub_device/button.py
```diff
@@ -0,0 +1,39 @@
+"""Xiaomi Zigbee buttons."""
+
+from .sub_device import SubDevice
+
+
+class Switch(SubDevice):
```
rytilahti, line 6: See my previous commentary on switch/button/remote_button division.

miio/gateway/sub_device/curtain.py
```diff
@@ -0,0 +1,30 @@
+"""Xiaomi Zigbee curtain."""
+
+from .sub_device import SubDevice
+
+
+class CurtainV1(SubDevice):
```
rytilahti, line 6: Common curtain base class?

miio/gateway/sub_device/sub_device.py
```diff
@@ -0,0 +1,221 @@
+"""Xiaomi Gateway subdevice base class."""
```
rytilahti, line 1: Maybe following could be a better directory structure for subdevices?
```
gateway/gateway.py
gateway/alarm.py
gateway/devices/subdevice.py (this main class)
gateway/devices/switch.py
```
etc.

miio/gateway/gateway.py
```diff
@@ -0,0 +1,420 @@
...
+GATEWAY_MODEL_AC_V1 = "lumi.acpartner.v1"
+GATEWAY_MODEL_AC_V2 = "lumi.acpartner.v2"
+GATEWAY_MODEL_AC_V3 = "lumi.acpartner.v3"
+
+
+class GatewayException(DeviceException):
+    """Exception for the Xioami Gateway communication."""
+
+
+class DeviceType(IntEnum):
```
rytilahti, line 27: This class should probably go into its own file under the subdevices directory. I think the mapping below also belongs there rather than be contained inside the `discover` method below.

miio/gateway/gateway.py
```diff
@@ -0,0 +1,420 @@
...
+        return self._info.model
+
+    @command()
+    def discover_devices(self):
+        """
+        Discovers SubDevices
+        and returns a list of the discovered devices.
+        """
+
+        from .sub_device import (
```
rytilahti, line 203: See my above comment wrt. location of these imports & the mapping. Maybe the subdevice class should have a static method returning an instance based on the mapping to simplify the code?

So the code below would be something like:
```
for x in range(0, len(devices_raw), 5):
    try:
        dev = SubDevice.device_from_raw(devices_raw[x : x + 5])
    except UnknownDeviceException:
        _LOGGER.warning("Unknown/unsupported device: %s", dev)
```

starkillerOG:
@rytilahti no problem, you don't have to hurry, it is indeed a large restructure.

I won't have any time the next 3 weeks since I will be moving to a new house.
So I don't know if I will be able to finish this PR any time soon, but maybe in the evening hours.... we will see.

miio/gateway/sub_device/lock.py
```diff
@@ -0,0 +1,39 @@
+"""Xiaomi Zigbee Locks."""
+
+from .sub_device import SubDevice
+
+
+class DoorLockS1(SubDevice):
```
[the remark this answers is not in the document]
starkillerOG: @rytilahti true, I will make a base class.
I did not do that yet because we have now idea how to control locks, so I don't have any code to put in the base class.
But I will make the base classes anyway for the future.

miio/gateway/sub_device/sensor.py
```diff
@@ -0,0 +1,39 @@
+"""Xiaomi Zigbee various sensors."""
+
+from .sub_device import SubDevice
+
+
+class SensorSmoke(SubDevice):
```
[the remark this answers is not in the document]
starkillerOG: I don't think so, these 3 sensors were kind of the odd ones that I don't know how to control and probably are all 3 diffrent from eachother...
I will not make a base class for these, we need more info on these devices before we can decide how they schould be implemented.....

miio/gateway/sub_device/switch.py
```diff
@@ -0,0 +1,245 @@
...
+from enum import IntEnum
+
+import attr
+import click
+
+from ...click_common import command
+from .sub_device import SubDevice
+
+
+class OneChannelSwitch(SubDevice):
```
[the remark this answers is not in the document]
starkillerOG: @rytilahti all the switches have shared base classes, the structure is:
OneChannelSwitch = Main base class
TwoChannelSwitch inherits from OneChannelSwitch and is also a base class
OneChannelSwitchLoad inherits from OneChannelSwitch and is also a base class
TwoChannelSwitchLoad inherits from OneChannelSwitch and is also a base class
ThreeChannelSwitchLoad inherits from OneChannelSwitch and is also a base class

The rest of the classes in this file represent individual physicall devices and inherit from one of the base classes above.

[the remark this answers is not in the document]
starkillerOG: `switches` are devices that can actually physically turn something on/off, they can supply power to an appliance such as a lamp.
`remote_switches` are just battery powered things you can stick anywhere on a wall and can only give an signal to the gateway that can trigger an automation to do someting (like turn on/off a smart light), they are supposed to be mounted on a wall.
`buttons` are basically the same thing as `remote_switches` but they are small hand held battery powered devices.

I think we could put `buttons` and `remote_switches` together since they might have a simular protecoll.
`switches` defenitly have diffrent commands, so schould be seperated.

miio/gateway/gateway.py
```diff
@@ -0,0 +1,420 @@
...
+        lazy_discover: bool = True,
+    ) -> None:
+        super().__init__(ip, token, start_id, debug, lazy_discover)
+
+        from .gateway_device import (
+            GatewayAlarm,
+            GatewayRadio,
+            GatewayZigbee,
+            GatewayLight,
+        )
```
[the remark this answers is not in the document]
starkillerOG: I will move them to the main gateway directory.

However they cannot be imported at the top of the file:
`GatewayAlarm` inherits from `GatewayDevice`
`GatewayDevice` needs `Gateway` for its constructor arguments.

Therefore `GatewayAlarm` needs to be imported after the `Gateway` class defenition, otherwhise you get an error that `Gateway` is undifined....

miio/gateway/gateway.py
```diff
@@ -0,0 +1,420 @@
...
+
+        self._alarm = GatewayAlarm(parent=self)
+        self._radio = GatewayRadio(parent=self)
+        self._zigbee = GatewayZigbee(parent=self)
+        self._light = GatewayLight(parent=self)
+        self._devices = {}
+        self._info = None
+
+    @property
+    def alarm(self) -> "GatewayAlarm":  # noqa: F821
```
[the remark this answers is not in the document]
starkillerOG: As explained it is not possible to move the imports to the top...

miio/gateway/gateway_device/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Xiaomi Gateway device base class."""
```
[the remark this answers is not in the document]
starkillerOG: true

miio/gateway/sub_device/__init__.py
starkillerOG: They already inherit form one base class that defines everything of these bulbs except the model and zigbee ID.
I do like keeping those since they will probably be needed to match devices using diffrent discovery methods such as cloud that give zigbee_id and not the ID_number that we currently use for matching

miio/gateway/sub_device/curtain.py
starkillerOG: will do

miio/gateway/sub_device/sub_device.py
starkillerOG: agreed

miio/gateway/sub_device/lock.py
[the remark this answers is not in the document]
starkillerOG: done

miio/gateway/sub_device/motion_sensor.py
```diff
@@ -0,0 +1,21 @@
+"""Xiaomi Zigbee motion."""
+
+from .sub_device import SubDevice
+
+
+class Motion(SubDevice):
```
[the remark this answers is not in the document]
starkillerOG: done

miio/gateway/sub_device/remote_switch.py
```diff
@@ -0,0 +1,57 @@
+"""Xiaomi Zigbee remote switches."""
+
+from .sub_device import SubDevice
+
+
+class RemoteSwitchDoubleV1(SubDevice):
```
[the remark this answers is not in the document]
starkillerOG: done

miio/gateway/sub_device/switch.py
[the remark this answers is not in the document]
starkillerOG: I have put the `buttons` and `remote_switches` together in the remote_switches.py file

miio/gateway/sub_device/button.py
starkillerOG: done

miio/gateway/sub_device/curtain.py
starkillerOG: done

miio/gateway/sub_device/switch.py
```diff
@@ -0,0 +1,245 @@
...
+
+        status: str = None  # 'on' / 'off'
+
+    @command()
+    def update(self):
+        """Update all device properties."""
+        values = self.get_property_exp(self.properties)
+        self._props.status = values[0]
+
+    class Channel_map(IntEnum):
```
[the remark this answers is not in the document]
starkillerOG: done

miio/gateway/gateway_device/__init__.py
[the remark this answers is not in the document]
starkillerOG: it has been removed

miio/gateway/gateway.py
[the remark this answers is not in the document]
starkillerOG: The classes are renamed and the files have been moved to the main directory

miio/gateway/gateway_device/gateway_device.py
starkillerOG: done

miio/gateway/sub_device/sub_device.py
starkillerOG: done

miio/gateway/gateway.py
starkillerOG: I have moved them to the `subdevice.py` file

miio/gateway/gateway.py
starkillerOG: I have moved large portions of the mapping from the gateway.py to the subdevice.py file.
However the exact function device_from_raw as you sugested is very specific to the info returned by the `self.get_prop("device_list")` command, therefore I would like to keep those together as it is now.

Other ways to get the subdevice list such as the Cloud or the `get_device_list` command will give back diffrently formatted info that needs to be mapped slichtly diffrent...

starkillerOG:
@rytilahti I have prosessed all feedback and all tests pass again.

MicroSur:
Is this PR will bring Aqara relay support in HomeAssistant at last?

starkillerOG:
@rytilahti Could you have another look and revieuw/merge this PR?

starkillerOG:
Could this be merged?

starkillerOG:
@rytilahti can you merge this/have another look at it?

rytilahti:
Hi @starkillerOG and sorry for the delay, I have been on a holiday and while I have seen your messages I haven't been able to spend some time to think through how to do a proper, maintainable implementation for sub-devices. Now that I'm back, I think I have a plan (or at least a direction) how I think this should be done.

As a foreword, I'm not very keen on having a new class per sub-device as is done in this PR. There are various reasons for that:

1. Having a new class for each and every new sub-device does not scale (who knows how many supported devices there are out there?). In fact, while preparing this comment, I spent some time to dig into the PR code to understand how different implementations differ from others, which class points to which "zigbee model/id" and so on. What I learned is that a deep class structures makes it laborious.

2. Creating tests for each device is hard and tiresome; no one wants to write more test code. So while I would like to see more tests in the code base, and to make that happen, less code is better.

3. Coming back to the point one, adding support for new device will require the implementer to dig deep to figure out which class is the most sane parent class to use for the base implementation.

So, I'm proposing that we change the approach how to add support for sub devices. This will will require some more work in this PR, but it will make the maintenance and adding support for new devices so much simpler that it's definitely worth it.

My proposal is that we think the sub devices in terms of device types (much like homeassistant does): lights, switches, weather sensors, alarms etc. share common features and should therefore be implemented in in a single class. 
In order to make the implementations of these device types adaptive, we need a new metadata file describing the supported (and unsupported?) zigbee devices. This file should contain information on what properties are available, what commands are used to fetch the information and how to change the values of those properties.

Here are some examples (in YAML, but I'm open for any structured file format) of how the configuration meta data could look like:

# A basic switch with just a toggle

The properties can define setters & getters.
If the getter is not defined, the default getter as defined in the `SubDevice` base class (`get_device_prop_exp` or `get_device_prop`?) will be used.

I'm not sure about the naming of the property type here, maybe it should be `channel` like in the API?

```
- zigbee_id: lumi.ctrl_neutral1.v1
  type: switch
  name: Wall switch no neutral (QBKG04LM)
  properties:
    - property: neutral_0
      type: switch
      index: 0
      getter: get_device_prop_exp
      setter: toggle_ctrl_neutral
```

## Switch with toggle & load power

Getter can also be defined on the device definition level to avoid repetition.
The extra features not implemented by all devices (such as load power) need to be defined separately.

```
- zigbee_id: lumi.plug
  name: Plug
  type: switch
  getter: get_prop_plug
  properties:
    - property: neutral_0
      type: switch
      index: 0
      setter: toggle_plug
    - property: load_power
      type: load
```

## Switch with multiple "switches"/"channels"

Switches with more than one channel/switch can be defined as below.
The API as already in this PR can be used, the `index` is the same as what is used for commanding the switch.

Alternatively, we can expose a list of `Channel` (or `Switch`?) classes based on the property information, which will simply have `toggle`, `turn_on` and `turn_off` methods for more object-oriented access.

```
- zigbee_id: lumi.switch.n3acn3
  name: D1 wall switch triple (QBKG26LM)
  type: switch
  setter: toggle_ctrl_neutral
  properties:
    - property: neutral_0
      type: switch
      index: 0
    - property: neutral_1
      type: switch
      index: 1
    - property: neutral_2
      type: switch
      index: 2
    - property: load_power
      type: load
```

# Weather sensors

Properties could also define extra information, such as used units, that can be useful for downstreams. We can add `properties()` property returning all the available properties as a list of `Property(name, unit, value)`.

```
- zigbee_id: lumi.sensor_ht.v1
  name: Weather sensor (WSDCGQ01LM)
  type: weather_sensor
  getter: get_prop_sensor_ht
  properties:
    - property: humidity
      unit: percent
    - property: temperature
      unit: celsius
    
- zigbee_id: lumi.weather.v1
  name: Weather sensor (WSDCGQ11LM)
  type: weather_sensor
  getter: get_prop_sensor_ht
  properties:
    - property: temperature
      unit: celsius
    - property: humidity
      unit: percent
    - property: pressure
      unit: hpa
```

# Unsupported devices

We likely want to define unsupported devices to allow printing out this information to the user who has such device, and ask for help to implement it (and to avoid issues about devices that are already known but not implemented).

A simple boolean flag should be enough in this case:
```
- zigbee_id: lumi.curtain
  name: Curtain (ZNCLDJ11LM)
  type: curtain
  supported: false
```
I'm not going to go any deeper to other device types, as this is merely written as motivation on how I would approach this. What do you think @starkillerOG?

Shortly put, this would have the following pros:

* Less code duplication and messy class hierarchies, simpler code base => easier to maintain and write tests.
* Easier extendability. Potentially no coding needed to add support for devices similar to existing ones.
* The metadata file can be used to generate documentation about supported devices => less documentation maintenance.

starkillerOG:
@rytilahti yes I definitely agree with you.
Thank you for the clear and nice feedback!
I won't have any time the coming weeks to work on this, but as soon as I have some spare time I will start rewriting to use a YAML file to define all the specific subdevices as you stated in your comment.

Then there will just be one class per type (like switch or wether_sensor) to define the code and all device specific information will be in the YAML.

I think the road ahead is clear, just need to find some time...

starkillerOG:
I picked up this PR again, current progress is in this branch: https://github.com/starkillerOG/python-miio/tree/patch-18
I will update this PR once I get the code back to a working state.
But for those of you that are interested you can have a peak at patch-18

starkillerOG:
Whoops, I messed up the re-base to fix some merge conflicts.
Therefore I will close this PR and make a new one here: https://github.com/rytilahti/python-miio/pull/924

PR #799, opened 2020-08-15
Added some parameters: Error code, Viomimode, Viomibintype

fs79:
added:

error codes
    2103: "Charging",
    2105: "Fully charged"

saw the 4 but don´t know what it means, but so no error message in cli command
class ViomiMode(Enum):
    Unknown = 4

class ViomiBinType(Enum):
    NoBin = 0

### DIFF
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -38,6 +38,8 @@ ERROR_CODES = {
     530: "Mop and water tank missing",
     531: "Water tank is not installed",
     2101: "Unsufficient battery, continuing cleaning after recharge",
+    2103: "Charging",
+    2105: "Fully charged",
 }
 
 
@@ -63,7 +65,7 @@ class ViomiMode(Enum):
     Vacuum = 0  # No Mop, Vacuum only
     VacuumAndMop = 1
     Mop = 2
-
+    Unknown = 4
 
 class ViomiLanguage(Enum):
     CN = 1  # Chinese (default)
@@ -94,7 +96,7 @@ class ViomiBinType(Enum):
     Vacuum = 1
     Water = 2
     VacuumAndWater = 3
-
+    NoBin = 0
 
 class ViomiWaterGrade(Enum):
     Low = 11
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -65,7 +65,8 @@ class ViomiMode(Enum):
     Vacuum = 0  # No Mop, Vacuum only
     VacuumAndMop = 1
     Mop = 2
-    Unknown = 4
+    CleanZone = 3
+    CleanSpot = 4
 
 class ViomiLanguage(Enum):
     CN = 1  # Chinese (default)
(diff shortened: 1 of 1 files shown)


miio/viomivacuum.py
```diff
@@ -63,7 +65,7 @@ class ViomiMode(Enum):
     Vacuum = 0  # No Mop, Vacuum only
     VacuumAndMop = 1
     Mop = 2
-
+    Unknown = 4
```
rytilahti, line 68: I think it would make more sense to add a log message(s) to the place(s) where the unknown mode is encountered, otherwise this looks good to go, thanks for the PR! :+1:

miio/viomivacuum.py
```diff
@@ -38,6 +38,8 @@
     530: "Mop and water tank missing",
     531: "Water tank is not installed",
     2101: "Unsufficient battery, continuing cleaning after recharge",
+    2103: "Charging",
+    2105: "Fully charged",
```
titilambert, line 42: I confirm this message, I have the same 
Thanks !

miio/viomivacuum.py
rytilahti: Ping?

titilambert: I'm not sure I understand your comment, you mean a log message in the `mode`  method ?
like here: https://github.com/rytilahti/python-miio/blob/5a3f2f446ff12210ba0c5fe0ede5016f4ca7c01e/miio/viomivacuum.py#L145 ?

rytilahti: I tried to say that it's more useful to simply log the unknown state on the call site, like done in https://github.com/rytilahti/python-miio/blob/5a3f2f446ff12210ba0c5fe0ede5016f4ca7c01e/miio/viomivacuum.py#L125 --  in case someone is seeing those warnings in their logs and knows what causes them, they can create a PR to fix it.

titilambert: If @fs79 doesn't answer, I can integrate this patch in this PR (https://github.com/rytilahti/python-miio/pull/808)

fs79:
I‘m fine with logging the unknown state. 
It’s important that the unknown state doesn’t destroy the output of the cli command. 
Im using the output and if there’s an issue due to unknown parameters it‘s an issue. 

Would you do it in general for unknown states in all the outputs or only for the above mentioned vacuum/mop/... state?

rytilahti:
Yes, I think it makes sense to log all unknown cases to avoid potential exceptions. Btw, if you want to parse the command line output, you may be interested in checking out the `--output` flag.

fs79:
@rytilahti 
You mean the "-o" flag?
Yes, I´m using it vor type vacuum but for viomivacuum it's not possible to use this flag.

rytilahti:
Yeah, I think `-o` is the shortcut for that flag, which should definitely be available (and working) on all devices, so it must be a bug that can be fixed. Please feel free to create a separate issue / debug that / create a PR to fix the issue.

fs79:
Issue for the -o json_pretty issue with viomivacuum cleaner.
https://github.com/rytilahti/python-miio/issues/816

miio/viomivacuum.py
syssi: Mode `4` is something like `CleanPosition`.

syssi (approved):

PR #803, opened 2020-08-19
Add get_property and get_value to MiotDevice

BenjaminFaal:
Instead of always getting all properties now you can just get one and also get the value directly

### DIFF
--- a/miio/miot_device.py
+++ b/miio/miot_device.py
@@ -29,6 +29,16 @@ class MiotDevice(Device):
         return self.get_properties(
             properties, property_getter="get_properties", max_properties=15
         )
+    
+    def get_property(self, property_key: str):
+        property = {"did": property_key, **self.mapping.get(property_key)}
+
+        return self.get_properties(
+            [property], property_getter="get_properties", max_properties=1
+        )[0]
+
+    def get_value(self, property_key: str):
+        return self.get_property(property_key)["value"]
 
     def set_property(self, property_key: str, value):
         """Sets property value."""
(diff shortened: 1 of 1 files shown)


miio/miot_device.py
```diff
@@ -29,6 +29,16 @@ def get_properties_for_mapping(self) -> list:
...
         )
+    
+    def get_property(self, property_key: str) -> dict:
+        property = {"did": property_key, **self.mapping.get(property_key)}
+
+        return self.get_properties(
+            [property], property_getter="get_properties", max_properties=1
+        )[0]
+
+    def get_value(self, property_key: str):
```
rytilahti, line 40: Would it be better just to have only `get_property` returning directly the value? Please also add docstring(s).

BenjaminFaal: get_property returns the whole property including the `code` property so you can check the code for the status of the property
get_value returns the value directly

miio/miot_device.py
```diff
@@ -29,6 +29,18 @@ def get_properties_for_mapping(self) -> list:
         return self.get_properties(
             properties, property_getter="get_properties", max_properties=15
         )
+    
+    def get_property(self, property_key: str) -> dict:
```
rytilahti, line 33: The return type should be `Dict` (import from typing is necessary) until python 3.9, iirc.

miio/miot_device.py
rytilahti: Okay, fair enough. Could you improve the docstrings to contain example return values (that'll make it easy to see what to expect)?

rytilahti:
Obsoleted by #905

Issue #637, opened 2020-02-22
Add support for lumi.acpartner.mcn02 please?


ihaettypo:
The most affordable Xiaomi AC Companion right now. Please add support thanks.

tcwj2008:
@rytilahti I have this device, is there anything that I can do

wzhd:
Mine just arrived and I did a little packet capturing.

It first did an HTTP request:

> GET /gslb?dm=ots.io.mi.com HTTP/1.1
> Host: dns.io.mi.com
> User-Agent: MIoT

The `User-Agent` is particularly concerning. The reply contains just a few IPs:

> {"info":{"host_list":[{"ip":"\<redacted>","port":443},{"ip":"\<redacted>","port":443},{"ip":"\<redacted>","port":443}],"enable":1},"sign":\<redacted>","timestamp":"1580070784"}

The rest are mostly TLSv1.2 communications with one of the returned IPs. The server name is `Mijia Cloud`.

![image](https://user-images.githubusercontent.com/3471744/82576943-e36dc400-9bbc-11ea-983e-981da5025250.png)

It did send some UDP packets from source port 54321 to dest port 8053.

I wonder whether it's a miot device.

rytilahti:
If you can obtain the token, you could try to execute `miot_info` command to see how is it responding. If it is a miot device, it will (hopefully) respond to that call.

The miot support is still WIP and there are a couple of PRs where the work is being done, but if it's a miot device, we can use the description files to help adding a preliminary support for that device. Otherwise it will require deciphering that udp device communication (which can be useful anycase, as the description files are not always complete).

rezmus:
it's miio device

https://cdn.cnbj1.fds.api.mi-img.com/rn-plugins/2019-12-26/signed_10032_1000741_15_ANDROID_bundle_6f61e13511ab120a80f23e852aba4d8a.zip

tcwj2008:
execute miiocli device info, it respond this

Model: lumi.acpartner.mcn02
Hardware version: esp32
Firmware version: 2.0.8_0002
Network: {'localIp': '192.168.1.111', 'mask': '255.255.255.0', 'gw': '192.168.1.1'}
AP: {'ssid': 'mywifi', 'bssid': '11:22:33:44:55:66', 'rssi': -56, 'primary': 1}

PR #804, opened 2020-08-23
Add support for lumi.acpartner.mcn02

EugeneLiu:
[Add support for lumi.acpartner.mcn02](https://github.com/rytilahti/python-miio/issues/637), product model is KTBL03LM


rytilahti:
Looks good to me from surface, but does this device differ so much from the other acpartners that it couldn't be integrated into the same class?

EugeneLiu:
This acpartners is an inexpensive version and does not have an auto-learning mode.The commands it uses to get status are different from other acpartners, so the status(like FanSpeed,OperationMode) is different too.And start_id use a random value for the command to work.

rytilahti:
Fair enough, please run `tox -e lint` and include the changed files to this PR to make the checks pass, see https://github.com/rytilahti/python-miio/pull/805#issuecomment-684042065 .

PR #805, opened 2020-08-29
Add consumable status to viomi vacuum

titilambert:
Tested with:
Model: viomi.vacuum.v7
Hardware version: Linux
Firmware version: 3.5.3_0047

PR #808, opened 2020-08-31
Improve Viomi support (status reporting, maps)

titilambert:
This PR refactors viomi vacuum (like STYJ02YM)

I tested all the attributes with STYJ02YM.
I re-organised the commands to stick with the android UI.
I added also some commands:
- maps management
- list rooms
- cleaning specific room

### DIFF
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -266,20 +266,6 @@ class ViomiRoutePattern(Enum):
     Y = 1
 
 
-class ViomiVoiceState(Enum):
-    Off = 0
-    Level_10 = 1
-    Level_20 = 2
-    Level_30 = 3
-    Level_40 = 4
-    Level_50 = 5
-    Level_60 = 6
-    Level_70 = 7
-    Level_80 = 8
-    Level_90 = 9
-    Level_100 = 10
-
-
 class ViomiEdgeState(Enum):
     Off = 0
     Unknown = 1
@@ -592,20 +578,20 @@ class ViomiVacuum(Device):
             "mode",
             "mop_route",
             "mop_type",
-            # "order_time",
             "remember_map",
             "repeat_state",
             "run_state",
             "s_area",
             "s_time",
-            # "start_time",
             "suction_grade",
             "v_state",
             "water_grade",
-            # "water_percent",
-            # "zone_data",
             # The following list of properties existing but
             # there are not used in the code
+            # "order_time",
+            # "start_time",
+            # "water_percent",
+            # "zone_data",
             # "sw_info",
             # "main_brush_hours",
             # "main_brush_life",
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -19,10 +19,10 @@ Settings:
 - Cleaning history - MISSING (cleanRecord)
 - Scheduled cleanup - get_ordertime
 - Vacuum along the edges - get_mode/set_mode
-- Secondary cleanup - set_repeat/repeat_state
+- Secondary cleanup - set_repeat/repeat_cleaning
 - Mop or vacuum & mod mode - set_moproute/mop_route
 - DND(DoNotDisturb) - set_notdisturb/get_notdisturb
-- Voice On/Off - set_voice/voice_state
+- Voice On/Off - set_sound_volume/voice_state
 - Remember Map - voice_state
 - Virtual wall/restricted area - MISSING
 - Map list - get_maps/rename_map/delete_map/set_map
@@ -94,7 +94,7 @@ class ViomiVacuumException(DeviceException):
 
 
 class ViomiPositionPoint:
-    """Vacuum position coordonate."""
+    """Vacuum position coordinate."""
 
     def __init__(self, pos_x, pos_y, phi, update, plan_multiplicator=1):
         self._pos_x = pos_x
@@ -110,15 +110,15 @@ class ViomiPositionPoint:
 
     @property
     def pos_y(self):
-        """Y coordonate with multiplicator."""
+        """Y coordinate with multiplicator."""
         return self._pos_y * self._plan_multiplicator
 
     def image_pos_x(self, offset, img_center):
-        """X coordonate on an image."""
+        """X coordinate on an image."""
         return self.pos_x - offset + img_center
 
     def image_pos_y(self, offset, img_center):
-        """Y coordonate on an image."""
+        """Y coordinate on an image."""
         return self.pos_y - offset + img_center
 
     def __repr__(self) -> str:
@@ -716,7 +716,7 @@ class ViomiVacuum(Device):
     def get_positions(self, plan_multiplicator=1) -> List[ViomiPositionPoint]:
         """Return the last positions.
 
-        returns: [x, y, phi, update, x, y, phi, update, x, y, phi, update, ...]
+        plan_multiplicator scale up the coordinates values
         """
         results = self.send("get_curpos", [])
         positions = []
@@ -814,7 +814,7 @@ class ViomiVacuum(Device):
         )
 
     @command(click.argument("state", type=EnumType(ViomiVoiceState)))
-    def set_voice(self, state: ViomiVoiceState):
+    def set_sound_volume(self, state: ViomiVoiceState):
         """Switch the voice on or off."""
         enabled = 1
         if state.value == 0:
@@ -864,7 +864,7 @@ class ViomiVacuum(Device):
         """Rename map."""
         maps = self.get_maps()
         if map_id not in [m["id"] for m in maps]:
-            raise ViomiVacuumException("Map id {} doesn't exists".format(map_id))
+            raise ViomiVacuumException(f"Map id {map_id} doesn't exists")
         return self.send("rename_map", {"mapID": map_id, "name": map_name})
 
     @command(
@@ -881,23 +881,18 @@ class ViomiVacuum(Device):
         if map_name:
             map_id = None
             maps = self.get_maps()
-            for map_ in maps:
-                if map_["name"] == map_name:
-                    map_id = map_["id"]
-            if map_id is None:
+            map_ids = [map_["id"] for map_ in maps if map_["name"] == map_name]
+            if not map_ids:
+                map_names = ", ".join([m["name"] for m in maps])
                 raise ViomiVacuumException(
-                    "Error: Bad map name, should be in {}".format(
-                        ", ".join([m["name"] for m in maps])
-                    )
+                    f"Error: Bad map name, should be in {map_names}"
                 )
+            map_id = map_ids[0]
         elif map_id:
             maps = self.get_maps()
             if map_id not in [m["id"] for m in maps]:
-                raise ViomiVacuumException(
-                    "Error: Bad map id, should be in {}".format(
-                        ", ".join([str(m["id"]) for m in maps])
-                    )
-                )
+                map_ids = ", ".join([str(m["id"]) for m in maps])
+                raise ViomiVacuumException(f"Error: Bad map id, should be in {map_ids}")
         # Get scheduled cleanup
         schedules = self.send("get_ordertime", [])
         scheduled_found, rooms = _get_rooms_from_schedules(schedules)
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -749,7 +749,7 @@ class ViomiVacuum(Device):
         raise NotImplementedError()
 
     @command()
-    def del_scheduled_cleanup(self):
+    def delete_timer(self):
         """Not implemented yet."""
         # Needs to reads and understand:
         # self.send("det_ordertime", [shedule_id])
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -709,7 +709,7 @@ class ViomiVacuum(Device):
         """
         self.send("set_suction", [watergrade.value])
 
-    def get_positions(self, plan_multiplicator=1) -> [ViomiPositionPoint]:
+    def get_positions(self, plan_multiplicator=1) -> List[ViomiPositionPoint]:
         """Return the last positions.
 
         returns: [x, y, phi, update, x, y, phi, update, x, y, phi, update, ...]
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -822,7 +822,7 @@ class ViomiVacuum(Device):
         return self.send("set_voice", [enabled, state.value])
 
     @command(click.argument("state", type=bool))
-    def set_remember(self, state: bool):
+    def set_remember_map(self, state: bool):
         """Set remember map state."""
         return self.send("set_remember", [int(state)])
 
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -475,20 +475,6 @@ def _get_rooms_from_schedules(schedules: List[str]) -> Tuple[bool, Dict]:
     More information:
     * https://github.com/homebridge-xiaomi-roborock-vacuum/homebridge-xiaomi-roborock-vacuum/blob/d73925c0106984a995d290e91a5ba4fcfe0b6444/index.js#L969
     * https://github.com/homebridge-xiaomi-roborock-vacuum/homebridge-xiaomi-roborock-vacuum#semi-automatic
-
-
-    >>> _parse_schedules(['1_0_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2'])
-    (True, {'11': 'room1', '13': 'room2'})
-    >>> _parse_schedules(['1_0_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_0_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
-    (True, {'11': 'room1', '13': 'room2', '10': 'room3', '14': 'room4'})
-    >>> _parse_schedules(['1_1_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_0_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
-    (True, {'10': 'room3', '14': 'room4'})
-    >>> _parse_schedules(['1_0_32_2_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_4_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
-    (False, {})
-    >>> _parse_schedules(['1_1_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2'])
-    (False, {})
-    >>> _parse_schedules([])
-    (False, {})
     """
     rooms = {}
     scheduled_found = False
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -431,10 +431,10 @@ class ViomiVacuumStatus:
         """Pattern mode."""
         return ViomiRoutePattern(self.data["mop_route"])
 
-    @property
-    def order_time(self) -> int:
-        """FIXME: ??? int or bool."""
-        return self.data["order_time"]
+    # @property
+    # def order_time(self) -> int:
+    #    """FIXME: ??? int or bool."""
+    #    return self.data["order_time"]
 
     @property
     def repeat_cleaning(self) -> bool:
@@ -444,25 +444,25 @@ class ViomiVacuumStatus:
         """
         return self.data["repeat_state"]
 
-    @property
-    def start_time(self) -> int:
-        """FIXME: ??? int or bool."""
-        return self.data["start_time"]
+    # @property
+    # def start_time(self) -> int:
+    #    """FIXME: ??? int or bool."""
+    #    return self.data["start_time"]
 
     @property
     def voice_state(self) -> ViomiVoiceState:
         """Voice volume level (from 0 to 100%, 0 means Off)."""
         return ViomiVoiceState(self.data["v_state"])
 
-    @property
-    def water_percent(self) -> int:
-        """FIXME: ??? int or bool."""
-        return self.data["water_percent"]
+    # @property
+    # def water_percent(self) -> int:
+    #    """FIXME: ??? int or bool."""
+    #    return self.data["water_percent"]
 
-    @property
-    def zone_data(self) -> int:
-        """FIXME: ??? int or bool."""
-        return self.data["zone_data"]
+    # @property
+    # def zone_data(self) -> int:
+    #    """FIXME: ??? int or bool."""
+    #    return self.data["zone_data"]
 
 
 def _get_rooms_from_schedules(schedules: List[str]) -> Tuple[bool, Dict]:
@@ -568,10 +568,10 @@ class ViomiVacuum(Device):
             "Unknown properties\n"
             "=================\n\n"
             "Light state: {result.light_state}\n"
-            "Order time: {result.order_time}\n"
-            "Start time: {result.start_time}\n"
-            "water_percent: {result.water_percent}\n"
-            "zone_data: {result.zone_data}\n",
+            # "Order time: {result.order_time}\n"
+            # "Start time: {result.start_time}\n"
+            # "water_percent: {result.water_percent}\n"
+            # "zone_data: {result.zone_data}\n",
         )
     )
     def status(self) -> ViomiVacuumStatus:
@@ -592,18 +592,18 @@ class ViomiVacuum(Device):
             "mode",
             "mop_route",
             "mop_type",
-            "order_time",
+            # "order_time",
             "remember_map",
             "repeat_state",
             "run_state",
             "s_area",
             "s_time",
-            "start_time",
+            # "start_time",
             "suction_grade",
             "v_state",
             "water_grade",
-            "water_percent",
-            "zone_data",
+            # "water_percent",
+            # "zone_data",
             # The following list of properties existing but
             # there are not used in the code
             # "sw_info",
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -742,7 +742,7 @@ class ViomiVacuum(Device):
         raise NotImplementedError()
 
     @command()
-    def set_scheduled_cleanup(self):
+    def add_timer(self):
         """Not implemented yet."""
         # Needs to reads and understand:
         # self.send("set_ordertime", [????])
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -103,7 +103,7 @@ class ViomiPositionPoint:
 
     @property
     def pos_x(self):
-        """X coordonate with multiplicator."""
+        """X coordinate with multiplicator."""
         return self._pos_x * self._plan_multiplicator
 
     @property
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -1,14 +1,59 @@
+"""Viomi Vacuum.
+
+# https://github.com/rytilahti/python-miio/issues/550#issuecomment-552780952
+# https://github.com/homebridge-xiaomi-roborock-vacuum/homebridge-xiaomi-roborock-vacuum/blob/ee10cbb3e98dba75d9c97791a6e1fcafc1281591/miio/lib/devices/vacuum.js
+# https://github.com/homebridge-xiaomi-roborock-vacuum/homebridge-xiaomi-roborock-vacuum/blob/ee10cbb3e98dba75d9c97791a6e1fcafc1281591/miio/lib/devices/viomivacuum.js
+
+Features:
+
+Main:
+- Area/Duration - Missing (get_clean_summary/get_clean_record
+- Battery - battery_life
+- Dock - set_charge
+- Start/Pause - set_mode_withroom
+- Modes (Vacuum/Vacuum&Mop/Mop) - set_mop/id_mop
+- Fan Speed (Silent/Standard/Medium/Turbo) - set_suction/suction_grade
+- Water Level (Low/Medium/High) - set_suction/water_grade
+
+Settings:
+- Cleaning history - MISSING (cleanRecord)
+- Scheduled cleanup - get_ordertime
+- Vacuum along the edges - get_mode/set_mode
+- Secondary cleanup - set_repeat/repeat_state
+- Mop or vacuum & mod mode - set_moproute/mop_route
+- DND(DoNotDisturb) - set_notdisturb/get_notdisturb
+- Voice On/Off - set_voice/voice_state
+- Remember Map - voice_state
+- Virtual wall/restricted area - MISSING
+- Map list - get_maps/rename_map/delete_map/set_map
+- Area editor - MISSING
+- Reset map - MISSING
+- Device leveling - MISSING
+- Looking for the vacuum-mop - MISSING (find_me)
+- Consumables statistics - get_properties
+- Remote Control - MISSING
+
+Misc:
+- Get Properties
+- Language - set_language
+- Led - set_light
+- Rooms - get_ordertime (hack)
+- Clean History Path - MISSING (historyPath)
+- Map plan - MISSING (map_plan)
+"""
+import itertools
 import logging
 import time
 from collections import defaultdict
 from datetime import timedelta
 from enum import Enum
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import click
 
 from .click_common import EnumType, command, format_output
 from .device import Device
+from .exceptions import DeviceException
 from .utils import pretty_seconds
 from .vacuumcontainers import ConsumableStatus, DNDStatus
 
@@ -16,6 +61,7 @@ _LOGGER = logging.getLogger(__name__)
 
 
 ERROR_CODES = {
+    0: "Sleeping and not charging",
     500: "Radar timed out",
     501: "Wheels stuck",
     502: "Low battery",
@@ -41,6 +87,49 @@ ERROR_CODES = {
     2103: "Charging",
     2105: "Fully charged",
 }
+
+
+class ViomiVacuumException(DeviceException):
+    """Exception raised by Viomi Vacuum."""
+
+
+class ViomiPositionPoint:
+    def __init__(self, pos_x, pos_y, phi, update, plan_multiplicator=1):
+        self._pos_x = pos_x
+        self._pos_y = pos_y
+        self.phi = phi
+        self.update = update
+        self._plan_multiplicator = plan_multiplicator
+
+    @property
+    def pos_x(self):
+        """X coordonate with multiplicator."""
+        return self._pos_x * self._plan_multiplicator
+
+    @property
+    def pos_y(self):
+        """Y coordonate with multiplicator."""
+        return self._pos_y * self._plan_multiplicator
+
+    def image_pos_x(self, offset, img_center):
+        """X coordonate on an image."""
+        return self.pos_x - offset + img_center
+
+    def image_pos_y(self, offset, img_center):
+        """Y coordonate on an image."""
+        return self.pos_y - offset + img_center
+
+    def __repr__(self) -> str:
+        return "<ViomiPositionPoint x: {}, y: {}, phi: {}, update {}".format(
+            self.pos_x, self.pos_y, self.phi, self.update
+        )
+
+    def __eq__(self, value) -> bool:
+        return (
+            self.pos_x == value.pos_x
+            and self.pos_y == value.pos_y
+            and self.phi == value.phi
+        )
 
 
 class ViomiConsumableStatus(ConsumableStatus):
@@ -89,6 +178,7 @@ class ViomiConsumableStatus(ConsumableStatus):
 
     @property
     def mop_left(self) -> timedelta:
+        """How long until the mop should be changed."""
         return self.sensor_dirty_total - self.sensor_dirty
 
     def __repr__(self) -> str:
@@ -167,17 +257,39 @@ class ViomiWaterGrade(Enum):
     High = 13
 
 
-class ViomiMopMode(Enum):
+class ViomiRoutePattern(Enum):
     """Mopping pattern."""
 
     S = 0
     Y = 1
 
 
+class ViomiVoiceState(Enum):
+    Off = 0
+    Level_10 = 1
+    Level_20 = 2
+    Level_30 = 3
+    Level_40 = 4
+    Level_50 = 5
+    Level_60 = 6
+    Level_70 = 7
+    Level_80 = 8
+    Level_90 = 9
+    Level_100 = 10
+
+
+class ViomiEdgeState(Enum):
+    Off = 0
+    Unknown = 1
+    On = 2
+    # NOTE: When I got 5, the device was super slow
+    # Shutdown and restart device fixed the issue
+    Unknown2 = 5
+
+
 class ViomiVacuumStatus:
     def __init__(self, data):
         # ["run_state","mode","err_state","battary_life","box_type","mop_type","s_time","s_area",
-        # [ 5,          0,     2103,       85,            3,         1,         0,       0,
         # "suction_grade","water_grade","remember_map","has_map","is_mop","has_newmap"]'
         # 1,               11,           1,            1,         1,       0          ]
         self.data = data
@@ -192,26 +304,21 @@ class ViomiVacuumStatus:
             return ViomiVacuumState.Unknown
 
     @property
-    def is_on(self) -> bool:
-        """True if cleaning."""
-        cleaning_states = [
-            ViomiVacuumState.Cleaning,
-            ViomiVacuumState.VacuumingAndMopping,
-        ]
-        return self.state in cleaning_states
+    def edge_state(self) -> ViomiEdgeState:
+        """Vaccum along the edges
 
-    @property
-    def mode(self):
-        """Active mode.
-
-        TODO: is this same as mop_type property?
+        The settings is valid once
+        0: Off
+        1: Unknown
+        2: On
+        5: Unknown
         """
-        return ViomiMode(self.data["mode"])
+        return ViomiEdgeState(self.data["mode"])
 
     @property
-    def mop_type(self):
-        """Unknown mop_type values."""
-        return self.data["mop_type"]
+    def mop_installed(self) -> bool:
+        """True if the mop is installed."""
+        return bool(self.data["mop_type"])
 
     @property
     def error_code(self) -> int:
@@ -239,7 +346,7 @@ class ViomiVacuumStatus:
     @property
     def clean_time(self) -> timedelta:
         """Cleaning time."""
-        return pretty_seconds(self.data["s_time"])
+        return pretty_seconds(self.data["s_time"] * 60)
 
     @property
     def clean_area(self) -> float:
@@ -250,6 +357,11 @@ class ViomiVacuumStatus:
     def fanspeed(self) -> ViomiVacuumSpeed:
         """Current fan speed."""
         return ViomiVacuumSpeed(self.data["suction_grade"])
+
+    @command()
+    def fan_speed_presets(self) -> Dict[str, int]:
+        """Return dictionary containing supported fanspeeds."""
+        return {x.name: x.value for x in list(ViomiVacuumSpeed)}
 
     @property
     def water_grade(self) -> ViomiWaterGrade:
@@ -268,57 +380,236 @@ class ViomiVacuumStatus:
 
     @property
     def has_new_map(self) -> bool:
-        """TODO: unknown"""
+        """True if the device has scanned a new map (like a new floor)."""
         return bool(self.data["has_newmap"])
 
     @property
     def mop_mode(self) -> ViomiMode:
-        """Whether mopping is enabled and if so which mode.
-
-        TODO: is this really the same as mode?
-        """
+        """Whether mopping is enabled and if so which mode."""
         return ViomiMode(self.data["is_mop"])
+
+    @property
+    def current_map_id(self) -> float:
+        """Current map id."""
+        return self.data["cur_mapid"]
+
+    @property
+    def hw_info(self) -> str:
+        """Hardware info."""
+        return self.data["hw_info"]
+
+    @property
+    def charging(self) -> bool:
+        """True if battery is charging
+
+        Note: When the battery is at 100%, device reports that it is not charging.
+        """
+        return not bool(self.data["is_charge"])
+
+    @property
+    def is_on(self) -> bool:
+        """True if device is working."""
+        return not bool(self.data["is_work"])
+
+    @property
+    def light_state(self) -> bool:
+        """Led state.
+
+        This seems doing nothing on STYJ02YM
+        """
+        return bool(self.data["light_state"])
+
+    @property
+    def map_number(self) -> int:
+        """Number of saved maps."""
+        return self.data["map_num"]
+
+    @property
+    def mop_route(self) -> ViomiRoutePattern:
+        """Pattern mode."""
+        return ViomiRoutePattern(self.data["mop_route"])
+
+    @property
+    def order_time(self) -> int:
+        """FIXME: ??? int or bool."""
+        return self.data["order_time"]
+
+    @property
+    def repeat_state(self) -> bool:
+        """Secondary clean up state."""
+        return self.data["repeat_state"]
+
+    @property
+    def start_time(self) -> int:
+        """FIXME: ??? int or bool."""
+        return self.data["start_time"]
+
+    @property
+    def voice_state(self) -> ViomiVoiceState:
+        """Voice volume level (from 0 to 100%, 0 means Off)."""
+        return ViomiVoiceState(self.data["v_state"])
+
+    @property
+    def water_percent(self) -> int:
+        """FIXME: ??? int or bool."""
+        return self.data["water_percent"]
+
+    @property
+    def zone_data(self) -> int:
+        """FIXME: ??? int or bool."""
+        return self.data["zone_data"]
+
+
+def _get_rooms_from_schedules(schedules: List[str]) -> Tuple[bool, Dict]:
+    """Read the result of "get_ordertime" command to extract room names and ids.
+
+    The `schedules` input needs to follow the following format
+    * ['1_0_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', ...]
+    * [Id_Enabled_Repeatdays_Hour_Minute_?_? _?_?_?_?_NbOfRooms_RoomId_RoomName_RoomId_RoomName_..., ...]
+
+    The function parse get_ordertime output to find room names and ids
+    To use this function you need:
+    1. to create a scheduled cleanup with the following properties:
+    * Hour: 00
+    * Minute: 00
+    * Select all (minus one) the rooms one by one
+    * Set as inactive scheduled cleanup
+    2. then to create an other scheduled cleanup with the room missed at
+       previous step with the following properties:
+    * Hour: 00
+    * Minute: 00
+    * Select only the missed room
+    * Set as inactive scheduled cleanup
+
+    More information:
+    * https://github.com/homebridge-xiaomi-roborock-vacuum/homebridge-xiaomi-roborock-vacuum/blob/d73925c0106984a995d290e91a5ba4fcfe0b6444/index.js#L969
+    * https://github.com/homebridge-xiaomi-roborock-vacuum/homebridge-xiaomi-roborock-vacuum#semi-automatic
+
+
+    >>> _parse_schedules(['1_0_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2'])
+    (True, {'11': 'room1', '13': 'room2'})
+    >>> _parse_schedules(['1_0_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_0_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
+    (True, {'11': 'room1', '13': 'room2', '10': 'room3', '14': 'room4'})
+    >>> _parse_schedules(['1_1_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_0_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
+    (True, {'10': 'room3', '14': 'room4'})
+    >>> _parse_schedules(['1_0_32_2_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_4_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
+    (False, {})
+    >>> _parse_schedules(['1_1_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2'])
+    (False, {})
+    >>> _parse_schedules([])
+    (False, {})
+    """
+    rooms = {}
+    scheduled_found = False
+    for raw_schedule in schedules:
+        schedule = raw_schedule.split("_")
+        # Scheduled cleanup needs to be scheduled for 00:00 and inactive
+        if schedule[1] == "0" and schedule[3] == "0" and schedule[4] == "0":
+            scheduled_found = True
+            raw_rooms = schedule[12:]
+            rooms_iter = iter(raw_rooms)
+            rooms.update(
+                dict(itertools.zip_longest(rooms_iter, rooms_iter, fillvalue=None))
+            )
+    return scheduled_found, rooms
 
 
 class ViomiVacuum(Device):
     """Interface for Viomi vacuums (viomi.vacuum.v7)."""
 
+    _cache = {"edge_state": None, "rooms": {}, "maps": {}}
+    timeout = 5
+    retry_count = 10
+
+    def __init__(
+        self, ip: str, token: str = None, start_id: int = 0, debug: int = 0
+    ) -> None:
+        super().__init__(ip, token, start_id, debug)
+        self.manual_seqnum = -1
+        # self.model = None
+        # self._fanspeeds = FanspeedV1
+
     @command(
         default_output=format_output(
-            "",
+            "\n",
+            "General\n"
+            "=======\n\n"
+            "Hardware version: {result.hw_info}\n"
             "State: {result.state}\n"
-            "Mode: {result.mode}\n"
-            "Error: {result.error}\n"
+            "Working: {result.is_on}\n"
+            "Battery status: {result.error}\n"
             "Battery: {result.battery}\n"
-            "Fan speed: {result.fanspeed}\n"
+            "Charging: {result.charging}\n"
             "Box type: {result.bin_type}\n"
-            "Mop type: {result.mop_type}\n"
-            "Clean time: {result.clean_time}\n"
-            "Clean area: {result.clean_area}\n"
+            "Fan speed: {result.fanspeed}\n"
             "Water grade: {result.water_grade}\n"
+            "Mop mode: {result.mop_mode}\n"
+            "Mop installed: {result.mop_installed}\n"
+            "Vacuum along the edges: {result.edge_state}\n"
+            "Mop route pattern: {result.mop_route}\n"
+            "Secondary Cleanup: {result.repeat_state}\n"
+            "Voice state: {result.voice_state}\n"
+            "Clean time: {result.clean_time}\n"
+            "Clean area: {result.clean_area} m²\n"
+            "\n"
+            "Map\n"
+            "===\n\n"
+            "Current map ID: {result.current_map_id}\n"
             "Remember map: {result.remember_map}\n"
             "Has map: {result.has_map}\n"
             "Has new map: {result.has_new_map}\n"
-            "Mop mode: {result.mop_mode}\n",
+            "Number of maps: {result.map_number}\n"
+            "\n"
+            "Unknown properties\n"
+            "=================\n\n"
+            "Light state: {result.light_state}\n"
+            "Order time: {result.order_time}\n"
+            "Start time: {result.start_time}\n"
+            "water_percent: {result.water_percent}\n"
+            "zone_data: {result.zone_data}\n",
         )
     )
     def status(self) -> ViomiVacuumStatus:
         """Retrieve properties."""
         properties = [
-            "run_state",
-            "mode",
-            "err_state",
             "battary_life",
             "box_type",
-            "mop_type",
-            "s_time",
-            "s_area",
-            "suction_grade",
-            "water_grade",
-            "remember_map",
+            "cur_mapid",
+            "err_state",
             "has_map",
-            "is_mop",
             "has_newmap",
+            "hw_info",
+            "is_charge",
+            "is_mop",
+            "is_work",
+            "light_state",
+            "map_num",
+            "mode",
+            "mop_route",
+            "mop_type",
+            "order_time",
+            "remember_map",
+            "repeat_state",
+            "run_state",
+            "s_area",
+            "s_time",
+            "start_time",
+            "suction_grade",
+            "v_state",
+            "water_grade",
+            "water_percent",
+            "zone_data",
+            # The following list of properties existing but
+            # there are not used in the code
+            # "sw_info",
+            # "main_brush_hours",
+            # "main_brush_life",
+            # "side_brush_hours",
+            # "side_brush_life",
+            # "mop_hours",
+            # "mop_life",
+            # "hypa_hours",
+            # "hypa_life",
         ]
 
         values = self.get_properties(properties)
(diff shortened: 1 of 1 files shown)


titilambert:
@rytilahti ping

miio/viomivacuum.py
```diff
@@ -174,19 +174,15 @@ class ViomiMopMode(Enum):
 
 class ViomiVacuumStatus:
     def __init__(self, data):
-        # ["run_state","mode","err_state","battary_life","box_type","mop_type","s_time","s_area",
```
rytilahti, line 177: Please leave the example query/response in place (simply update it with the new values).

miio/viomivacuum.py
```diff
@@ -174,19 +174,15 @@ class ViomiMopMode(Enum):
...
-        # [ 5,          0,     2103,       85,            3,         1,         0,       0,
-        # "suction_grade","water_grade","remember_map","has_map","is_mop","has_newmap"]'
-        # 1,               11,           1,            1,         1,       0          ]
-        self.data = data
+        self._data = data
```
rytilahti, line 177: Please keep `self.data` naming as this would be otherwise the only status container with a different variable naming.

miio/viomivacuum.py
```diff
@@ -227,96 +223,265 @@ def error(self) -> Optional[str]:
...
             "suction_grade",
+            "sw_info",
+            "v_state",
             "water_grade",
-            "remember_map",
-            "has_map",
-            "is_mop",
-            "has_newmap",
+            "water_percent",
+            "zone_data",
```
rytilahti, line 484: This is quite many variables, are you sure this is working?

miio/viomivacuum.py
```diff
@@ -227,96 +223,265 @@ def error(self) -> Optional[str]:
...
+        return self._data["cur_mapid"]
+
+    @property
+    def hw_info(self) -> str:
+        """Hardware info."""
+        return self._data["hw_info"]
+
+    @property
+    def sw_info(self) -> str:
+        """SoftWare info."""
```
rytilahti, line 288:
```suggestion
        """Software info."""
```

Do you know how this differs from the responses to `miIO.info` call, or is it the same? If the information is the same, I'm wondering if it is worth to have this separately. Same for `hw_info`.

miio/viomivacuum.py
```diff
@@ -227,96 +223,265 @@ def error(self) -> Optional[str]:
...
+        return self._data["sw_info"]
+
+    @property
+    def hypa_hours_left(self) -> int:
+        """HYPA left hours."""
+        return self._data["hypa_hours"]
+
+    @property
+    def hypa_life_left(self) -> int:
+        """HYPA left life percent."""
```
rytilahti, line 298: Is this just a misnamed HEPA filter, or something else?

miio/viomivacuum.py
```diff
@@ -227,96 +223,265 @@ def error(self) -> Optional[str]:
...
+        """Hardware info."""
+        return self._data["hw_info"]
+
+    @property
+    def sw_info(self) -> str:
+        """SoftWare info."""
+        return self._data["sw_info"]
+
+    @property
+    def hypa_hours_left(self) -> int:
```
rytilahti, line 292: Return the values for consumable `_left`s as `timedelta`s (to be consistent with other vacuums).

rytilahti (changes requested):
Thanks for the PR! I did a quick initial review, do you have plans to figure out the unknown variables and are you sure requesting so many variables is not breaking some limits?

edit: some of the consumables are duplicates and already available thanks to #805?

miio/viomivacuum.py
titilambert: Yes, it's the HEPA filter, but the prop is really 'hypa_life'.
What should I do ? keep `hypa_life_left` or change it for `hepa_life_left`

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
rytilahti: It is unfortunately not uncommon that the property names do not describe what the value really is. As this library builds an API for other people to use, I prefer reasonable naming over following the naming schemes used by device manufacturers.

So I think the name should be `hepa_filter_left` to follow the naming scheme from vacuumcontainers.py, even though (based on a grep through the code base) in some cases `life` is there in the name. So that part is not a big deal, pick whichever you think is more useful for any users of the lib.

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: I renamed it to hepa_life_left, is it correct ?

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: Only hardware seems not good with mIO.info

miIO.info
```
Hardware version: Linux
Firmware version: 3.5.3_0047
```

status
```
Hardware version: 1.0.3
Software version: 3.5.3_0047
```

miio/viomivacuum.py
titilambert: So I will rename it `hepa_filter_left`
Thanks !

BTW, do you how can I have the details for each function listed here:
* https://github.com/rytilahti/python-miio/issues/550#issuecomment-552780952
and here
* https://github.com/rytilahti/python-miio/issues/550#issuecomment-570808184

For example, the function, `rename_map` I don't know which parameter it takes ...

miio/viomivacuum.py
titilambert: What do you want me to do ?

rytilahti: I think the extra information is not that useful, so I think simply removing it (or juts keeping the hardware version, if really wanted) is the way to go. I prefer to avoid unnecessary/unhelpful code and keep things like sw/fw information generic for all devices, this will help also if someone wants to use this library to build a generic interface for miio devices.

miio/viomivacuum.py
rytilahti: Alas, capturing the necessary parameters require reverse engineering the plugin for that device, either by reading reverse-engineered code or capturing the traffic between the device and the official app. @rezmus may have an idea of the parameters you are looking for :-)

miio/viomivacuum.py
rytilahti: What was exactly fixed? Did you maybe forgot to push the changes?

miio/viomivacuum.py
titilambert: Thanks ! I'm open the android plugin and I got what I need, I will try to add all the functions that I can :)

miio/viomivacuum.py
titilambert: ok, I'll do that

miio/viomivacuum.py
rytilahti: Please use the new keys & values so that anyone who is looking for the newly introduced properties has an idea how a response looks like.

miio/viomivacuum.py
rytilahti: Please also remove the `hw_info` as that's already available in the info response, right?

miio/viomivacuum.py
```diff
@@ -277,46 +281,216 @@ def mop_mode(self) -> ViomiMode:
...
+        return timedelta(self.data["hypa_hours"])
+
+    @property
+    def hepa_life_left(self) -> int:
+        """HYPA left life percent."""
+        return self.data["hypa_life"]
+
+    @property
+    def charging(self) -> bool:
+        """FIXME: True if device is charging?"""
```
rytilahti, line 311: Could you try to verify the `FIXME` marked items?

miio/viomivacuum.py
```diff
@@ -277,46 +281,216 @@ def mop_mode(self) -> ViomiMode:
...
+        return bool(self.data["is_charge"])
+
+    @property
+    def working(self) -> bool:
+        """FIXME: True if device is working?"""
+        return bool(self.data["is_work"])
```
rytilahti, line 317: Does working mean vacuuming here? That is, do you think `is_on` already provides this information, making this new property unnecessary.

miio/viomivacuum.py
```diff
@@ -277,46 +281,216 @@ def mop_mode(self) -> ViomiMode:
...
+
+    @property
+    def light_state(self) -> bool:
+        """FIXME: True if device ?"""
+        return bool(self.data["light_state"])
```
rytilahti, line 322: Maybe this is related to the indicator led lights, if the device has such?

miio/viomivacuum.py
```diff
@@ -277,46 +281,216 @@ def mop_mode(self) -> ViomiMode:
...
+
+    @property
+    def main_brush_hours_left(self) -> timedelta:
+        """Main brush left hours."""
+        return timedelta(self.data["main_brush_hours"])
```
rytilahti, line 327: The consumable states are already exposed using `ViomiConsumableStatus`, so I'd simply remove the duplicate values (I think hepa is the only one not available through it, currently?)

miio/viomivacuum.py
```diff
@@ -439,3 +613,29 @@ def carpet_mode(self, mode: ViomiCarpetTurbo):
...
         return {x.name: x.value for x in list(ViomiVacuumSpeed)}
+
+    @command(click.argument("state", type=bool))
+    def set_repeat(self, state: bool):
+        """Set or Unset repeat mode."""
+        return self.send("set_repeat", [int(state)])
+
+    @command()
+    def get_maps(self) -> List[Dict[str, Any]]:
+        """Return map list."""
```
rytilahti, line 624: Could you add an example response for this request? If the obtained information could be useful for some other features (e.g., "use this map", or something similar), it may make sense to create a container class out of this that can be re-used.

rytilahti:
@titilambert Ping? Could you also update the description for the PR?

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: doesn't seem having any effect on the device.
Added a comment

titilambert:
> @titilambert Ping? Could you also update the description for the PR?

done

miio/viomivacuum.py
titilambert: Try to test it next Monday or Tuesday

titilambert:
@rytilahti I made a refactoring to have something structured like the android app.
I added a big module docstring to explain what is missing or not.
I still need to test some properties.
After that, I guess, we could merge this PR.
Then, the thing missing will be the management of the scheduled cleanup !

miio/viomivacuum.py
titilambert: I what thinking replacing the `get_consumables` handler byt `get_properties`  with the props:
* "main_brush_hours",
* "main_brush_life",
* ...

It seems simpler to manage (no calculation/timedelta manipulation)
@rytilahti what do you think ?

miio/viomivacuum.py
titilambert: Yes ! few more props to test :)

miio/viomivacuum.py
titilambert: I would like to keep it since it's the actual value.

rytilahti: As you wish :+1:

miio/viomivacuum.py
rytilahti: It is better to use the consumable status container for consistency; at some point someone may want to add support for non-rockrobo vacuums to homeassistant integration and having a common API makes it simpler to achieve.

miio/viomivacuum.py
```diff
@@ -279,113 +344,393 @@ def mop_mode(self) -> ViomiMode:
...
+        self,
+        ip: str = None,
+        token: str = None,
+        start_id: int = 0,
+        debug: int = 0,
+        lazy_discover: bool = True,
+        timeout: int = 5,
+    ) -> None:
+        super().__init__(ip, token, start_id, debug, lazy_discover, timeout=0.5)
+        self._cache = {"edge_state": None, "rooms": {}}
```
rytilahti, line 471: Please avoid overriding the `__init__` if possible. The cache could be initialized during the first use, if it's needed.

miio/viomivacuum.py
```diff
@@ -279,113 +344,393 @@ def mop_mode(self) -> ViomiMode:
...
+        parameters: Any = None,
+        retry_count=20,
+        *,
+        extra_parameters=None,
+    ) -> Any:
+        # For retry_count to 20
+        return super().send(
+            command, parameters, retry_count=20, extra_parameters=extra_parameters
+        )
+
```
rytilahti, line 485:
```suggestion
```

miio/viomivacuum.py
```diff
@@ -279,113 +344,393 @@ def mop_mode(self) -> ViomiMode:
...
+            "water_percent",
+            "zone_data",
         ]
 
         values = self.get_properties(properties)
 
         return ViomiVacuumStatus(defaultdict(lambda: None, zip(properties, values)))
 
+    # The following commands are available on the MAIN page
+    # in the Mi Home Android app
```
rytilahti, line 589:
```suggestion
```

miio/viomivacuum.py
```diff
@@ -279,113 +344,393 @@ def mop_mode(self) -> ViomiMode:
...
+            "-r",
+            multiple=True,
+            help="Rooms name or room id. Can be used multiple times",
+        )
+    )
+    def start_with_room(self, rooms):
+        """Start cleaning specific rooms."""
+        if not self._cache["rooms"]:
+            self.list_rooms()
+            if not self._cache["rooms"]:
```
rytilahti, line 623: It is better to raise an exception (in `list_rooms()`) instead of using this structure.

miio/viomivacuum.py
```diff
@@ -279,113 +344,393 @@ def mop_mode(self) -> ViomiMode:
...
+        [low, medium, high]
+        """
         self.send("set_suction", [watergrade.value])
 
-    @command()
-    def home(self):
-        """Return to home."""
-        self.send("set_charge", [1])
+    # The following commands are available on the SETTINGS page
+    # in the Mi Home Android app
```
rytilahti, line 690:
```suggestion
```

miio/viomivacuum.py
```diff
@@ -279,113 +344,393 @@ def mop_mode(self) -> ViomiMode:
...
-    def clean_mode(self, mode):
-        """Set the cleaning mode."""
-        self.send("set_mop", [mode.value])
+    @command()
+    def get_scheduled_cleanup(self):
+        """Not implemented yet."""
+        # Needs to reads and understand the return of:
+        # self.send("get_ordertime", [])
+        # [id, enabled, repeatdays, hour, minute, ?, ? , ?, ?, ?, ?, nb_of_rooms, room_id, room_name, room_id, room_name, ...]
+        return "Not Implemented yet."
```
rytilahti, line 700: `raise NotImplementedError()` for not-yet implemented functionality.

miio/viomivacuum.py
```diff
@@ -422,22 +767,164 @@ def set_dnd(
...
+    def get_maps(self) -> List[Dict[str, Any]]:
+        """Return map list."""
+        return self.send("get_map")
+
+    @command(click.argument("map_id", type=int))
+    def set_map(self, map_id: int):
+        """Change current map."""
+        maps = self.get_maps()
+        if map_id not in [m["id"] for m in maps]:
+            return "Map id {} doesn't exists".format(map_id)
```
rytilahti, line 792: Raise an exception for such cases.

miio/viomivacuum.py
```diff
@@ -422,22 +767,164 @@ def set_dnd(
...
+        return self.send("set_map", [map_id])
+
+    @command(click.argument("map_id", type=int))
+    def delete_map(self, map_id: int):
+        """Delete map."""
+        maps = self.get_maps()
+        if map_id not in [m["id"] for m in maps]:
+            return "Map id {} doesn't exists".format(map_id)
```
rytilahti, line 800: See my comment above, also prefer f-strings over using `format()`, e.g.:
```
raise ViomiVacuumException(f"Map id {map_id} does not exist")
```

miio/viomivacuum.py
```diff
@@ -422,22 +767,164 @@ def set_dnd(
...
+    def move(self, direction, duration=0.5):
+        """Manual movement."""
+        start = time.time()
+        while time.time() - start < duration:
+            self.send("set_direction", [direction.value])
+            time.sleep(0.1)
+        self.send("set_direction", [ViomiMovementDirection.Stop.value])
+
+    # The following commands don't seem to be available
+    # in the Mi Home Android app
```
rytilahti, line 906:
```suggestion
```

miio/device.py
```diff
@@ -113,10 +113,13 @@ def __init__(
...
         debug: int = 0,
         lazy_discover: bool = True,
+        timeout: int = 5,
     ) -> None:
         self.ip = ip
         self.token = token
-        self._protocol = MiIOProtocol(ip, token, start_id, debug, lazy_discover)
+        self._protocol = MiIOProtocol(
+            ip, token, start_id, debug, lazy_discover, timeout
+        )
```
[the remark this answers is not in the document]
titilambert: done here: https://github.com/rytilahti/python-miio/pull/866

miio/discovery.py
```diff
@@ -253,10 +253,14 @@ def check_and_create_device(self, info, addr) -> Optional[Device]:
...
     def add_service(self, zeroconf, type, name):
         info = zeroconf.get_service_info(type, name)
-        addr = str(ipaddress.ip_address(info.address))
-        if addr not in self.found_devices:
-            dev = self.check_and_create_device(info, addr)
-            self.found_devices[addr] = dev
+        for address in info.addresses:
+            addr = str(ipaddress.ip_address(address))
+            if addr == "192.168.15.110":
+                continue
```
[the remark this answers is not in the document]
titilambert: sorry, fixed

miio/viomivacuum.py
```diff
@@ -102,6 +147,9 @@ def __repr__(self) -> str:
             )
         )
 
+    def __json__(self):
+        return self.data
+
```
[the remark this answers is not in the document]
titilambert: fixed

miio/viomivacuum.py
titilambert: I need to set `retry_count=20`. It improve the stability of the requests.
How can I handle that ?

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: In fact I need to set the timeout to 0.5  seconds. Same question as the send method: is there any other way to change the default values ?

miio/viomivacuum.py
titilambert: @rytilahti I think I found a way...
Please look at the change the in device.py file.

miio/viomivacuum.py
titilambert: @rytilahti I think I found a way...
Please look at the change the in device.py file.

syssi:
@titilambert Take a look at you inbox. I've sent you an e-mail (subject: "Some material to support your viomi PR").

titilambert:
> ")

titilambert:
> @titilambert Take a look at your inbox. I've sent you an e-mail (subject: "Some material to support your viomi PR").

Thanks ! I already had this file but only for viomi.v7
I think I got everything I can (for now)

titilambert:
@rytilahti I found a way to create a map in live using Pillow.
Once this PR is merged, I will create a new one just for this feature
Are you open to add Pillow as dependency in the project ?

titilambert:
@rytilahti I think this PR is now ready !

titilambert:
@rytilahti The PR is now rebased

miio/exceptions.py
```diff
@@ -32,3 +32,7 @@ def __init__(self, error):
 
 class RecoverableError(DeviceError):
     """Exception communicating an recoverable error delivered by the target device."""
+
+
+class ViomiVacuumException(DeviceException):
```
rytilahti, line 37: Please move this to `viomivacuum.py`, this file is only used for exceptions common for the whole project.

miio/viomivacuum.py
```diff
@@ -201,17 +268,24 @@ def is_on(self) -> bool:
...
-    def mode(self):
-        """Active mode.
+    def edge_state(self) -> ViomiEdgeState:
+        """Vaccum along the edges
 
-        TODO: is this same as mop_type property?
+        The settings is valid once
+        0: disabled
+        2: enabled
+        5: unknown
```
rytilahti, line 277: These values are different than what is defined in `ViomiEdgeState`.

miio/viomivacuum.py
```diff
@@ -201,17 +268,24 @@ def is_on(self) -> bool:
...
 
     @property
-    def mop_type(self):
-        """Unknown mop_type values."""
-        return self.data["mop_type"]
+    def mop_installed(self) -> bool:
+        """Mop installed status
+
+        True if the mop is installed
+        False if the mop is NOT installed
```
rytilahti, line 286:
```suggestion
        """True if mop is installed.
```
This delivers the same information with much shorter docstring :-) Less is more sometimes!

miio/viomivacuum.py
```diff
@@ -268,124 +347,340 @@ def has_map(self) -> bool:
...
+    def charging(self) -> bool:
+        """Battery is charging or not.
+
+        is_charge is 1 when the battery is not charging
+        is_charge is 0 when the device is charging
+        Note: When the battery is at 100% is_charge is 1
+
+        Return:
+        - True if the battery is charging
+        - False if the battery is NOT charging
```
rytilahti, line 378:
```suggestion
        """True if battery is charging.

        Note: When the battery is at 100%, device reports that it is not charging.
```

miio/viomivacuum.py
```diff
@@ -268,124 +347,340 @@ def has_map(self) -> bool:
...
         """
-        return ViomiMode(self.data["is_mop"])
+        return not (bool(self.data["is_charge"]))
```
rytilahti, line 380:
```suggestion
        return not bool(self.data["is_charge"])
```

miio/viomivacuum.py
```diff
@@ -268,124 +347,340 @@ def has_map(self) -> bool:
...
+
+    @property
+    def working(self) -> bool:
+        """Device is working or not.
```
rytilahti, line 384: See my suggestion above.

Also, could we rename this to `is_on` for consistency?

miio/viomivacuum.py
```diff
@@ -268,124 +347,340 @@ def has_map(self) -> bool:
...
+    def clean_mode(self, mode: ViomiMode):
+        """Set the cleaning mode.
+
+        [vacuum, vacuumAndMop, mop, cleanzone, cleanspot]
+        """
+        self.send("set_mop", [mode.value])
 
     @command(click.argument("speed", type=EnumType(ViomiVacuumSpeed)))
     def set_fan_speed(self, speed: ViomiVacuumSpeed):
-        """Set fanspeed [silent, standard, medium, turbo]."""
```
rytilahti, line 346: Why change this? I think it is not too long to be shown in the cli tool, which can be helpful?

miio/viomivacuum.py
```diff
@@ -422,22 +717,228 @@ def set_dnd(
...
+        This seems doing nothing on STYJ02YM
+        """
         return self.send("set_carpetturbo", [mode.value])
 
-    @command()
-    def fan_speed_presets(self) -> Dict[str, int]:
-        """Return dictionary containing supported fanspeeds."""
-        return {x.name: x.value for x in list(ViomiVacuumSpeed)}
+    @classmethod
+    def get_device_group(cls):
```
rytilahti, line 895: We don't want to have code duplication for this "hack" from mirobo, rather, we want to have this on the miiocli level for all supported devices. I wrote a bit about the plan here https://github.com/rytilahti/python-miio/issues/728#issuecomment-685967596 .

miio/viomivacuum.py
```diff
@@ -422,22 +717,228 @@ def set_dnd(
...
+        for raw_schedule in schedules:
+            schedule = raw_schedule.split("_")
+            # Scheduled cleanup needs to be scheduled for 00:00 and inactive
+            if schedule[1] == "0" and schedule[3] == "0" and schedule[4] == "0":
+                scheduled_found = True
+                raw_rooms = schedule[12:]
+                rooms_iter = iter(raw_rooms)
+                rooms.update(
+                    dict(itertools.zip_longest(rooms_iter, rooms_iter, fillvalue=None))
+                )
```
rytilahti, line 817: I suggest creating a new method (`parse_schedule`) returning a schedule item in wanted format.

miio/viomivacuum.py
```diff
@@ -167,17 +224,27 @@ class ViomiWaterGrade(Enum):
...
+class ViomiRoutePattern(Enum):
     """Mopping pattern."""
 
     S = 0
     Y = 1
 
 
+class ViomiVoiceState(Enum):
+    Off = 1
+    On = 5
```
rytilahti, line 236: If this has just two states, it'd be better to use a simple boolean.

miio/device.py
```diff
@@ -117,6 +117,8 @@ def __init__(
     ) -> None:
         self.ip = ip
         self.token = token
+        if hasattr(self, "timeout"):
```
rytilahti, line 120: Any suggested modifications to the base class should be done in a PR of its own.

miio/exceptions.py
titilambert: done

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: Like
```
    @property
    def voice_state(self) -> bool:
        """Voice state."""
        return self.data["v_state"] == 5
```
?

titilambert: Fixed

miio/device.py
titilambert: done here: https://github.com/rytilahti/python-miio/pull/884

miio/viomivacuum.py
titilambert: fixed here: https://github.com/rytilahti/python-miio/pull/884

miio/viomivacuum.py
titilambert: fixed there: https://github.com/rytilahti/python-miio/pull/884

miio/viomivacuum.py
titilambert: ho wow, this is a huge change. I'm removing this change for now

miio/viomivacuum.py
titilambert: fixed

titilambert:
@rytilahti need to get PR https://github.com/rytilahti/python-miio/pull/884 merge, then I rebase this one and It will be ready for review

Neuroquila-n8fall:
Great work!
I have tested this PR against my device and got the following findings. These are mostly observations when running specific commands.
> Model: viomi.vacuum.v8
Hardware version: Linux
Firmware version: 3.5.3_0017

- `status` errors out: It requests 27 properties but only 23 are sent back:
> DEBUG:miio.miioprotocol:10.0.0.169:54321 >>: {'id': 203, 'method': 'get_prop', 'params': ['battary_life', 'box_type', 'cur_mapid', 'err_state', 'has_map', 'has_newmap', 'hw_info', 'is_charge', 'is_mop', 'is_work', 'light_state', 'map_num', 'mode', 'mop_route', 'mop_type', 'order_time', 'remember_map', 'repeat_state', 'run_state', 's_area', 's_time', 'start_time', 'suction_grade', 'v_state', 'water_grade', 'water_percent', 'zone_data']}
DEBUG:miio.miioprotocol:10.0.0.169:54321 (ts: 2021-01-16 10:12:29, id: 203) << {'result': [100, 1, 2105, 1, 0, '1.0.3', 0, 0, 1, 1, 0, 0, '0', 1, 0, 5, 0, 0, 0, 2, 10, 13, '0'], 'id': 203}
DEBUG:miio.device:Count (27) of requested properties does not match the count (23) of received values.
ValueError: '0' is not a valid ViomiVacuumSpeed

- `get_rooms` is working like a charm
- `get_current_position` yields `none` when the vacuum is docked and not moving. Also when it is in a state where it is trying to figure out where it is in the room. Not an issue per-se but it's worth noting i think. Sometimes this command is answered very quickly but most of the time it takes more than 30 seconds to arrive.
> DEBUG:miio.miioprotocol:10.0.0.169:54321 >>: {'id': 1, 'method': 'get_curpos', 'params': []}
DEBUG:miio.miioprotocol:10.0.0.169:54321 (ts: 2021-01-16 10:16:22, id: 1) << {'result': [], 'id': 1}

- `carpet_mode` is doing something but i have no idea if this is a bug (or feature) or not that it throttles its speed to standard when on a carpet and goes back to its selected speed when off the carpet
- `consumable_status` is always returning `ConsumableStatus main: 13:00:00, side: 13:00:00, filter: 13:00:00, mop: 3:00:00`
...no matter how long the device has been running. Could be a firmware kink.


I think my device has a strange firmware revision since the official app does not show me the mop pattern selection and carpet mode but I do have the v8.
Because I am very interested to bring this forward I recorded some traffic on the firewall and phone directly when in action with timestamps when I changed something on the app. This includes not only the traffic from mobile to vac but also "mother ship" to vac. I can provide you this pcap if you are interested. Cheers!

jeremysherriff:
The `consumable_status` output doesn't make sense to me. Am I understanding it correctly?
The output of `raw_command get_consumables` returns an array of integers that each increase by 1 for every one hour that the vacuum has been operating - note *increases*.
From what I can tell in the PR, the output of `consumable_status` is supposed to be the time in days+hours left until the consumable(s) need to be changed - but the time keeps increasing?
I see there is some math (https://github.com/rytilahti/python-miio/blob/a3c3b5a178409b0d6d3c846888c655a0af8fe0f5/miio/viomivacuum.py#L139) but the `status` command is throwing errors for me (see next comment) so I can't evaluate whether they are providing the correct output or not.

jeremysherriff:
`status` command is throwing an error;
```
jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx status

ValueError: '0' is not a valid ViomiVacuumSpeed

During handling of the above exception, another exception occurred:

  File "/usr/local/bin/miiocli", line 8, in <module>
    sys.exit(create_cli())
  File "/usr/local/lib/python3.8/dist-packages/miio/cli.py", line 45, in create_cli
    return cli(auto_envvar_prefix="MIIO")
  File "/usr/local/lib/python3.8/dist-packages/miio/click_common.py", line 59, in __call__
  File "/usr/lib/python3/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1137, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 1137, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 956, in invoke
  File "/usr/lib/python3/dist-packages/click/core.py", line 555, in invoke
  File "/usr/local/lib/python3.8/dist-packages/miio/click_common.py", line 285, in wrap
    result_msg = result_msg_fmt.format(**kwargs)
  File "/usr/local/lib/python3.8/dist-packages/miio/viomivacuum.py", line 359, in fanspeed
    return ViomiVacuumSpeed(self.data["suction_grade"])
  File "/usr/lib/python3.8/enum.py", line 309, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib/python3.8/enum.py", line 600, in __new__
    raise exc
  File "/usr/lib/python3.8/enum.py", line 584, in __new__
    result = cls._missing_(value)
  File "/usr/lib/python3.8/enum.py", line 613, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: '0' is not a valid ViomiVacuumSpeed
````
Sorry, my coding skills are not sufficient to understand what is wrong.

**Comment by syssi** [comment]
@jeremysherriff Please provide the model identifier of your device. Did you install/test this feature branch or did you use an official `python-miio` release?

**Comment by jeremysherriff** [comment]
Testing additional functionality. Firmware version 3.5.3_0017.
* get_maps errors with a timeout
* get_rooms instructions were clear (except did not specify a repeat pattern, but I found that any repeat pattern seemed to work). Command then succeeded and produced useful output:
`{'10': 'Bathroom', '12': 'Kitchen', '15': 'Hallway', '13': 'Living Room'}`
* get_current_position returns None when docked, I see this has been mentioned before somewhere. Once the vacuum is somewhere else, i get a good response: `<ViomiPositionPoint x: 1.927604, y: -2.555805, phi: -3.123265, update 1`
* carpet_mode is working well
I'm happy to test anything if someone lets me know what.

**Comment by jeremysherriff** [comment]
> @jeremysherriff Please provide the model identifier of your device. Did you install/test this feature branch or did you use an official `python-miio` release?

@syssi I have probably installed it incorrectly; I simply dropped the viomivacuum.py file from this PR into /usr/local/lib/python3.8/dist-packages/miio/
````
jeremy@tvbox:~$ miiocli --version
miiocli, version 0.5.4
````
````
jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx info
Model: viomi.vacuum.v8
Hardware version: Linux
Firmware version: 3.5.3_0017
````

**Comment by jeremysherriff** [comment]
`get_current_position` output is missing the closing `>`
````
jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx get_current_position
Running command get_current_position
<ViomiPositionPoint x: 1.927604, y: -2.555805, phi: -3.123265, update 1
jeremy@tvbox:~$
````

**Comment by Neuroquila-n8fall** [comment]
> The `consumable_status` output doesn't make sense to me. Am I understanding it correctly?
> The output of `raw_command get_consumables` returns an array of integers that each increase by 1 for every one hour that the vacuum has been operating - note _increases_.
> From what I can tell in the PR, the output of `consumable_status` is supposed to be the time in days+hours left until the consumable(s) need to be changed - but the time keeps increasing?
> I see there is some math (see;)
> 
> https://github.com/rytilahti/python-miio/blob/a3c3b5a178409b0d6d3c846888c655a0af8fe0f5/miio/viomivacuum.py#L139
> 
> but the `status` command is throwing errors for me (see next comment) so I can't evaluate whether they are providing the correct output or not.

We are on the same boat - this firmware version does not support consumable counters yet amongst multi-map support and all the good stuff the chinese version already has. Guess we've to wait until the FW is rolled out that supports it.

The "none" position when docked or when the device is trying to figure out where it is in the room can be ignored at all. Even the Mi Home app does have lags because sometimes the position is unclear. It's safe to assume a position when `state` reached `docked` and just take the last known valid position when `state` equals `pause` but these are things that for the most part the UI should handle and display correctly if you plan to achieve just that.

**Comment by titilambert** [comment]
I made my tests/development with viomi.vacuum.v7, so I guess it's different than the v8 :/
Should we do one file per hardware version !?

**Comment by titilambert** [comment]
> `get_current_position` output is missing the closing `>`
> 
> ```
> jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx get_current_position
> Running command get_current_position
> <ViomiPositionPoint x: 1.927604, y: -2.555805, phi: -3.123265, update 1
> jeremy@tvbox:~$
> ```

I'm fixing that

**Comment by titilambert** [comment]
@Neuroquila-n8fall  @jeremysherriff Let's start by fixing the status issue.
I know how to do it: You have any 23/27 elements. You need to identify which elements are missing.
To do so, edit the `properties` list in the status function, and keep only the first element.
Run the cli status command and see if you have one element. if yes, add a second one, re run the cli status command and see if you have two elements, and so on ... until you know all the missing element.
Then I will patch the code to support your version of the vacuum.
Then we will see for the others commands.

**Comment by titilambert** [comment]
@Neuroquila-n8fall @jeremysherriff One of you could create a new issue where we could track which feature is missing for viomi.vacuum.v8. What do you think ?

**Comment by titilambert** [comment]
@rytilahti I just rebase the branch. Can we merge it ?

**Comment by titilambert** [comment]
```
carpet_mode is doing something but i have no idea if this is a bug (or feature) or not that it throttles its speed to standard when on a carpet and goes back to its selected speed when off the carpet
```
Same for v7 (which doesn't have carpet mode)

**Comment by titilambert** [comment]
@rytilahti I guess the tests are correct. The error seems just during the setup  of windows. 
Is it good for you ?

**Comment by jeremysherriff** [comment]
Carpet mode is working correctly for viomi.v8 even though is is not an option in the Mi Home app.

**Comment by jeremysherriff** [comment]
A few of the values produce no output for me, so the `get_prop` response is parsed incorrectly:
* "mop_route"
* "water_percent"
* "cur_mapid"
* "map_num"

I assume the first two are because I don't have a mop attached (is_mop = 0).
I'm guessing that you will need to break the mop-related stuff out to different status queries, after identifying if a mop is attached. 
Map-related stuff appears firmware-version related.  Perhaps it would be a good idea to include `sw_info` in the `status` to allow for detecting of a multi-map capable version, both inside python-miio and also in external integrations leveraging this library.
```
jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx raw_command get_prop '["mop_route","water_percent","cur_mapid","map_num",]'
Running command raw_command
[]
```
The remaining properties all produce valid output consistently for me.

**Comment by jeremysherriff** [comment]
@titilambert @rytilahti  it occurs to that this will be a breaking change, due to the renaming and removal of a number of previous status fields (Mode, Error, Mop Type), as well as the formatting change of the `status` output (sections with headers, underlines, and white-space between sections).

In addition to this, the 'Clean time:' correction (from seconds to minutes) and the addition of the units on `Clean area:` (adding m²) will also potentially break up-stream users of python-miio.  The area units should be part of the field name (`Clean area m²:`) and potentially shouldn't contain special characters (`Clean area sqm:` might be better?).  I would prefer to see new fields introduced, even for `Clean time:`), the old fields marked as deprecated, and a well-published date for removal of those fields to allow for up-stream development to move to the new values in an orderly way.

**Comment by jeremysherriff** [comment]
I have not been successful with `start_with_room` - it always appears to run a full clean rather than a room-specific clean.  @titilambert  Do I have the format correct?
```
jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx get_rooms
Running command get_rooms
{'10': 'Bathroom', '12': 'Kitchen', '15': 'Hallway', '13': 'Living Room'}

jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx start_with_room -r '10'
Running command start_with_room
None
```
What does the `None` response indicate?

I also tried `-r 10` and `-r 'Bathroom'` with the same result - a full cleanup initiated.

Also, the use of the `-r` parameter flag appears to be different to other commands that take parameters (such as `set_edge`, `set_water_grade`, and `set_fan_speed`).

The `--help` text is incorrect, as it flags the rooms parameters as optional (indicated by [SQUAREBRACKETS]).
Examples of how to provide the room parameter(s) would be useful too, something like this perhaps (and assuming the dropping of the `-r` requirement):
```
Usage: miiocli viomivacuum start_with_room ROOM[[,ROOM],...]

  Start cleaning specific rooms.
  e.g. miiocli viomivacuum --ip <ip> --token <token> start_with_room '2','Bathroom'

Options:
  ROOM              Required. ROOM as name or id, comma separated.  Can be used multiple times.
                    See get_rooms command for valid room names and IDs.
  --help            Show this message and exit.
```

**Comment by Neuroquila-n8fall** [comment]
> @Neuroquila-n8fall @jeremysherriff One of you could create a new issue where we could track which feature is missing for viomi.vacuum.v8. What do you think ?

Was a little absent the last days... Right now it seems the firmware isn't rolling out anytime soon so I'll do the required tests and will open a new issue for the V8 at the current firmware level. I'd be very interested if the issues persist when the new FW is up. Maybe we don't have to care much about the V8 after the fact because the features are the same like on the V7.
_Side note: I'm very annoyed that the V8 has a "dumber" firmware than it's predecessor._

**Comment by titilambert** [comment]
> Carpet mode is working correctly for viomi.v8 even though is is not an option in the Mi Home app.

What do you mean is working ? how do you see it's working ? I mean what does the robot in carpet mode ?

**Comment by titilambert** [comment]
@jeremysherriff First thing to know about viomi, (like @Neuroquila-n8fall said) DON'T TRUST the "name" of the properties and functions. ie, the "carpet_mode" could a switch to turn on/off the led on the robot ...
So, I think we should add an option to specify which version of viomi you have (v7,v8,...) and add some `if` in the code to handle different versions.

That's why I want to merge first this merge request then create a new PR for v7 VS v8 management

**Comment by jeremysherriff** [comment]
I've previously owned a S6 model, which includes carpet mode natively in
the Mi Home app.
Carpet mode uses the sensor eyes to detect that it is not a hard-surface
floor, and then changes the suction/fan speed to the carpet_mode speed that
has been selected.
I have my vacuum set to "Medium" fan speed, and then carpet_mode set to
Turbo.  This means it works harder on the carpets, and conserves battery on
the (easier-to-clean) tiled flooring.
You can hear the vacuum change fan speed as it moved from my kitchen to my
dining room, and then changing again as it moves into the tiled hallway.

Currently, carpet_mode is working perfectly for my v8 kodel using the
viomivacuum.py from this PR.

**Comment by rytilahti** [comment]
> @titilambert @rytilahti it occurs to that this will be a breaking change, due to the renaming and removal of a number of previous status fields (Mode, Error, Mop Type), as well as the formatting change of the `status` output (sections with headers, underlines, and white-space between sections).
>
> In addition to this, the `Clean time:` correction (from seconds to minutes) and the addition of the units on `Clean area:` (adding m²) will also potentially break up-stream users of python-miio. The area units should be part of the field name (`Clean area m²:`) and potentially shouldn't contain special characters (`Clean area sqm:` might be better?). I would prefer to see new fields introduced, even for `Clean time:`), the old fields marked as deprecated, and a well-published date for removal of those fields to allow for up-stream development to move to the new values in an orderly way.

I'm not seeing how this would be a breaking change, the formatting of cli commands is not really part of the API, downstreams should be either using the library directly, or using the output from `--json` to script it.

There is and will be no guarantee that the outputs do not change, it would make it way too complicated to keep track on changes. For the library API, the aim is to keep backwards compatibility as long as it can be easily done.

File: miio/viomivacuum.py [hunk 35]
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+        return bool(self.data["light_state"])
+
+    @property
+    def map_number(self) -> int:
+        """Number of saved maps."""
+        return self.data["map_num"]
+
+    @property
+    def mop_route(self) -> ViomiRoutePattern:
+        """Pattern mode."""
```
rytilahti, line 429: Is this only relevant for mopping? If yes, the docstring should probably be `Mop route pattern`?

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+        return ViomiRoutePattern(self.data["mop_route"])
+
+    @property
+    def order_time(self) -> int:
+        """FIXME: ??? int or bool."""
+        return self.data["order_time"]
```
rytilahti, line 435: Do you know what does this do? If not, it's probably better to remove it for now and re-add it later on to avoid breaking the API in the future. Same applies to other properties marked with `FIXME` below, so I'm not going to comment them separately.

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+
+    @property
+    def repeat_state(self) -> bool:
+        """Secondary clean up state."""
```
rytilahti, line 439: I'm not sure I understand what does this do?

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+    for raw_schedule in schedules:
+        schedule = raw_schedule.split("_")
+        # Scheduled cleanup needs to be scheduled for 00:00 and inactive
+        if schedule[1] == "0" and schedule[3] == "0" and schedule[4] == "0":
+            scheduled_found = True
+            raw_rooms = schedule[12:]
+            rooms_iter = iter(raw_rooms)
+            rooms.update(
+                dict(itertools.zip_longest(rooms_iter, rooms_iter, fillvalue=None))
+            )
```
rytilahti, line 513: Should there be an else branch to inform the user about something being wrong, e.g., by raising an exception or logging it?

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+            self.get_rooms()
+        reverse_rooms = {v: k for k, v in self._cache["rooms"].items()}
+        room_ids = []
+        for room in rooms:
+            if room in self._cache["rooms"]:
+                room_ids.append(room)
+            elif room in reverse_rooms:
+                room_ids.append(reverse_rooms[room])
+            else:
+                return "Rooms {} is unknown, it should be in '{}' " "or in '{}'".format(
```
rytilahti, line 660: Better raise an exception here, makes it easier to handle for downstream users.
Also, use f-strings instead of `"".format()`:
```
room_keys = ", ".join()
room_ids = ", ".join()
raise Exception(f"Room must be in {room_keys} or {room_ids}")

File: miio/viomivacuum.py [hunk 40]
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+        )
 
     @command()
     def pause(self):
         """Pause cleaning."""
-        self.send("set_mode_withroom", [0, 2, 0])
+        # params: [edge_state, 0]
+        # - edge: see ViomiEdgeState
+        # - 2: pause cleaning
+        if not self._cache["edge_state"]:
```
rytilahti, line 677: When can this happen? Should the cache always be updated like in `start_with_room`?

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+            self._cache["edge_state"] = self.get_properties(["mode"])
+        self.send("set_mode", self._cache["edge_state"] + [2])
+
+    @command()
+    def stop(self):
+        """Validate that Stop cleaning."""
+        # params: [edge_state, 0]
+        # - edge: see ViomiEdgeState
+        # - 0: stop cleaning
+        if not self._cache["edge_state"]:
```
rytilahti, line 687: Same as above for cache.

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+    @command(click.argument("watergrade", type=EnumType(ViomiWaterGrade)))
     def set_water_grade(self, watergrade: ViomiWaterGrade):
-        """Set water grade [low, medium, high]."""
+        """Set water grade.
+
+        [low, medium, high]
+        """
         self.send("set_suction", [watergrade.value])
 
+    def get_positions(self, plan_multiplicator=1) -> [ViomiPositionPoint]:
```
rytilahti, line 712:
```suggestion
    def get_positions(self, plan_multiplicator=1) -> List[ViomiPositionPoint]:
```

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+        """Return the last positions.
+
+        returns: [x, y, phi, update, x, y, phi, update, x, y, phi, update, ...]
```
rytilahti, line 715:
```suggestion
```
This can be removed, as the type hint already reveals the return type.

The plan_multiplicator should be documented, though.

miio/viomivacuum.py
```diff
@@ -423,22 +809,165 @@ def set_dnd(
...
+            return self._cache["rooms"]
+        if map_name:
+            map_id = None
+            maps = self.get_maps()
+            for map_ in maps:
+                if map_["name"] == map_name:
+                    map_id = map_["id"]
+            if map_id is None:
+                raise ViomiVacuumException(
+                    "Error: Bad map name, should be in {}".format(
```
rytilahti, line 885: Use f-string instead of format.

miio/viomivacuum.py
```diff
@@ -423,22 +809,165 @@ def set_dnd(
...
+    ):
+        """Return room ids and names."""
+        if self._cache["rooms"] and not refresh:
+            return self._cache["rooms"]
+        if map_name:
+            map_id = None
+            maps = self.get_maps()
+            for map_ in maps:
+                if map_["name"] == map_name:
+                    map_id = map_["id"]
```
rytilahti, line 882: Maybe?
```suggestion
            map = [map["id"] for map in maps if map["name"] == map_name][0]
```

miio/viomivacuum.py
```diff
@@ -423,22 +809,165 @@ def set_dnd(
...
+                    )
+                )
+        # Get scheduled cleanup
+        schedules = self.send("get_ordertime", [])
+        scheduled_found, rooms = _get_rooms_from_schedules(schedules)
+        if not scheduled_found:
+            msg = (
+                "Fake schedule not found. "
+                "Please create a scheduled cleanup with the "
+                "following properties:\n"
```
rytilahti, line 904: Is it always necessary to have this "fake schedule"? If not, maybe just log a warning saying that there are no scheduled cleans? Otherwise `get_rooms` would be unusable due to the exception.

Neuroquila-n8fall:
> > Carpet mode is working correctly for viomi.v8 even though is is not an option in the Mi Home app.
> 
> What do you mean is working ? how do you see it's working ? I mean what does the robot in carpet mode ?
contrary to reports of @jeremysherriff I can confirm, though, it does something but not that it's working as intended:
I can confirm that it does switch on a strange mode where as soon as the robo detects carpet, it will **decrease** suction power, if mode is set to turbo or normal. I've discovered also that this decreases the effectiveness and therefore kept it off. This is almost certainly related to the firmware that is currently installed.


> @jeremysherriff First thing to know about viomi, (like @Neuroquila-n8fall said) DON'T TRUST the "name" of the properties and functions. ie, the "carpet_mode" could a switch to turn on/off the led on the robot ...
> So, I think we should add an option to specify which version of viomi you have (v7,v8,...) and add some if in the code to handle different versions.
> 
> That's why I want to merge first this merge request then create a new PR for v7 VS v8 management

I would rather look on the firmware version. Other than that it makes absolute sense to merge this with a hint to the v7 models and then we can work on the v8 integration - but again i would rather wait for the new firmware. Not that we have all this trouble that might be solved in this future update. _I discovered today that the app received an update at least that lists the past firmware revisions. We might be only a few days away from the actual FW update. _

jeremysherriff:
> contrary to reports of @jeremysherriff I can confirm, though, it does something but not that it's working as intended:
> I can confirm that it does switch on a strange mode where as soon as the robo detects carpet, it will **decrease** suction power, if mode is set to turbo or normal. I've discovered also that this decreases the effectiveness and therefore kept it off. This is almost certainly related to the firmware that is currently installed.

I can confirm that carpet_mode is buggy, for my v8 the suction level drops back to the default mode (currently "normal") when it approaches walls and when it goes under furniture while on carpet.  I'm guessing but I suspect this is due to the way the sensors are checked to see if the surface is hard/shiny.

Neuroquila-n8fall:
Can't really seem to find an option to open an issue on the respective branches so i just put it here.

So we are talking exclusively about this particular combo of HW and FW:
Model: viomi.vacuum.v8
Hardware version: Linux
Firmware version: 3.5.3_0017
**Note** "Hardware Version" is `1.0.3` as stated by the `hw_info` property.

I checked the commands when it sat on the docking station with dust bin inserted. I also verified the enums - they are all correct. 
```
battary_life			Y		
box_type			Y		
cur_mapid			Empty Response []		
err_state			Y		
has_map				Y		
has_newmap			Y		
hw_info				Y		
is_charge			Empty Response []		
is_mop				Y		
is_work				Y		
light_state			Y		
map_num				Empty Response []		
mode				Y		
mop_route			Empty Response []		
mop_type			Y		
order_time			Y		
remember_map			Y		
repeat_state			Empty Response []		
run_state			Y		
s_area				Y		
s_time				Y		
start_time			Y		
suction_grade			Y		
v_state				Y		
water_grade			Y		
water_percent			Empty Response []		
zone_data			Y		
sw_info				Y		
main_brush_hours		Empty Response []		
main_brush_life			Empty Response []		
side_brush_hours		Empty Response []		
side_brush_life			Empty Response []		
mop_hours			Empty Response []		
mop_life			Empty Response []		
hypa_hours			Empty Response []		
hypa_life			Empty Response []		
```

Next step was vacuum 4 zones:
```
is_charge			1 (was empty)
repeat_state			0 (0 repeats = single run, was empty)
```

Next iteration was while a selective room-by-room, 2-in-1 cleaning task was in progress. 
Other than the usual changes in modes and states because of the selected cleaning profile no further commands started to work.


The following commands, which are introduced in this PR, are confirmed working on the v8:
```
get_current_position
get_rooms
```

Again, I would skip the missing properties and wait for the firmware update because you can't make it work when the device doesn't support the extra functions. What I don't get is why the `--rooms` parameter is not working. I think I have to catch a few pcaps of that.....

jeremysherriff:
I've been experimenting with getting specific rooms cleaned, and realised that my Russian hacked Mi Home app is logging the vacuum calls.  Here's a successful call to vacuum the Hallway with a single pass:
```
2021-01-29 15:53:50 -> {"id":163,"method":"set_mode_withroom","params":[0,1,1,15]}
2021-01-29 15:53:50 <- {"code":0,"id":163,"otlocalts":1611888830803511,"result":["ok"]}
```

As per previous, my rooms are:
```
jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx get_rooms
Running command get_rooms
{'10': 'Bathroom', '12': 'Kitchen', '15': 'Hallway', '13': 'Living Room'}
```

jeremysherriff:
In the interest of further understanding this viomi vacuum, I see the following get_prop call repeatedly:
```
2021-01-29 15:53:47 -> {"id":162,"method":"get_prop","params":["sw_info","run_state","mode","err_state","battary_life","box_type","mop_type","s_time","s_area","suction_grade","water_grade","remember_map","has_map","is_mop","has_newmap","timezone"]}
2021-01-29 15:53:47 <- {"code":0,"id":162,"otlocalts":1611888827763392,"result":["3.5.3_0017",5,0,2105,100,1,0,46,37.130000,2,12,1,1,0,0,-46800]}
```

This is the first time I have seen a `timezone` property mentioned, which presumably affects DND and schedule timings.
Anyway, just thought it might be useful.

miio/viomivacuum.py
titilambert: Yes it's mandatory to have schedules in order to see the rooms.
So what do you want me to do ?

miio/viomivacuum.py
titilambert: It's for both mopping activated and not activated

miio/viomivacuum.py
titilambert: It just does a second cleanup (after the first one it's done)

miio/viomivacuum.py
rytilahti: Ok :+1:

miio/viomivacuum.py
rytilahti: Would `repeat_cleaning` be more clear? With a docstring `True if the cleaning is performed twice`?

jeremysherriff:
There is a lot of comments in this PR, and it's very hard to see what is now resolved and what still needs work.
I would like to help, please ignore if the following is not useful;

`edge_state` is a toggle setting in the Mi Home app (so probably a boolean).  The wording in English in the app is
`Vacuum along the edges` with description `This setting is valid once`. It will cause the vacuum to **only** vacuum the edges of the room/zone/map. The setting reverts to off/0/false once the clean-up completes or is cancelled.

`repeat_state` is also a toggle setting in the app. The wording in English in the app is
`Secondary cleanup` with description `After enabling the cleanup will be repeated, only takes effect once`. It makes the vacuum repeat the cleanup room/zone/map twice over (for extra clean!). The setting reverts to off/0/false once the clean-up completes or is cancelled.

Screenshot of these settings in app:
image

jeremysherriff:
The voice volume is set using integers ranging from 1 to 10, represented in the app as 10% to 100%.  I assume this is the same as other Xiaomi models?

Here's logs from the Mi Home app with me setting the volume to 90% and then back to 100%:
```
2021-02-03 19:50:31 -> {"id":1115,"method":"set_voice","params":[1,9]}
2021-02-03 19:50:31 <- {"code":0,"id":1115,"otlocalts":1612335031389261,"result":["ok"]}

2021-02-03 19:50:35 -> {"id":1117,"method":"set_voice","params":[1,10]}
2021-02-03 19:50:35 <- {"code":0,"id":1117,"otlocalts":1612335035817524,"result":["ok"]}
```
When disabling the voice, the parameters become [0,5] (presumably, "off" and "50%") and the app hides the volume setting.  When re-enabling, the app sets the voice volume to 50% regardless of what volume it was originally.  The following is the result of me disabling the voice when it is was 100% volume, and then re-enabling, and manually setting the volume back to 100%:
```
2021-02-03 19:57:40 -> {"id":4146,"method":"set_voice","params":[0,5]}
2021-02-03 19:57:40 <- {"code":0,"id":4146,"otlocalts":1612335461613965,"result":["ok"]}

2021-02-03 19:57:45 -> {"id":4148,"method":"set_voice","params":[1,5]}
2021-02-03 19:57:45 <- {"code":0,"id":4148,"otlocalts":1612335466505960,"result":["ok"]}

2021-02-03 19:58:48 -> {"id":4157,"method":"set_voice","params":[1,10]}
2021-02-03 19:58:48 <- {"code":0,"id":4157,"otlocalts":1612335528958184,"result":["ok"]}
```

miio/viomivacuum.py
rytilahti: Ah, okay. I assume there is no way to such a fake schedule programatically (at least for)? Assuming any schedule (and not just a fake one) is fine, the error message should rather be something like "No schedules found. You can create a fake schedule with the following properties ..", but I'm okay with this for now.

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+        return self.data["repeat_state"]
+
+    @property
+    def start_time(self) -> int:
+        """FIXME: ??? int or bool."""
+        return self.data["start_time"]
+
+    @property
+    def voice_state(self) -> ViomiVoiceState:
+        """Voice volume level (from 0 to 100%, 0 means Off)."""
```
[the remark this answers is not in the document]
titilambert: It's already the case because the `set_voice` use an enum (`off` to `level_100`)

miio/viomivacuum.py
```diff
@@ -43,6 +89,49 @@
 }
 
 
+class ViomiVacuumException(DeviceException):
+    """Exception raised by Viomi Vacuum."""
+
+
+class ViomiPositionPoint:
```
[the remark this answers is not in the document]
titilambert: done

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+            rooms.update(
+                dict(itertools.zip_longest(rooms_iter, rooms_iter, fillvalue=None))
+            )
+    return scheduled_found, rooms
 
 
 class ViomiVacuum(Device):
     """Interface for Viomi vacuums (viomi.vacuum.v7)."""
 
+    _cache = {"edge_state": None, "rooms": {}, "maps": {}}
```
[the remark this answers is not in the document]
titilambert: fixed

miio/viomivacuum.py
```diff
@@ -268,124 +380,398 @@ def has_map(self) -> bool:
...
+    def set_edge(self, state: ViomiEdgeState):
+        """Set or Unset edge mode.
+
+        Vacuum along the edges The settings is valid once
+        """
+        return self.send("set_mode", [state.value])
+
+    @command(click.argument("state", type=bool))
+    def set_repeat(self, state: bool):
+        """Set or Unset repeat mode."""
```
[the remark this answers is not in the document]
titilambert: With repeat mode, the vacuum clean a second time.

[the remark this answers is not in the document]
titilambert: I updated the doc string

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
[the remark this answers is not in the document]
titilambert: is it good for you ?

miio/viomivacuum.py
```diff
@@ -423,22 +809,165 @@ def set_dnd(
             [0 if disable else 1, start_hr, start_min, end_hr, end_min],
         )
 
+    @command(click.argument("state", type=EnumType(ViomiVoiceState)))
+    def set_voice(self, state: ViomiVoiceState):
```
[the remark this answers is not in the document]
titilambert: fixed

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: fixed

miio/viomivacuum.py
titilambert: It happens if you run `pause` before start

miio/viomivacuum.py
titilambert: It happens if you run `pause` before start

miio/viomivacuum.py
titilambert: No, I'm looking for only the "fake schedules" meaning starting at 00:00 and inactive. I'm ignoring the other ones.

miio/viomivacuum.py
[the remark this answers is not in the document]
rytilahti: `set_sound_volume` on `Vacuum` accepts integers, and likewise, `sound_volume` returns integers, so in my opinion this should do the same to be consistent across devices.

miio/viomivacuum.py
```diff
@@ -268,124 +382,400 @@ def has_map(self) -> bool:
...
+    >>> _parse_schedules(['1_0_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_0_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
+    (True, {'11': 'room1', '13': 'room2', '10': 'room3', '14': 'room4'})
+    >>> _parse_schedules(['1_1_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_0_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
+    (True, {'10': 'room3', '14': 'room4'})
+    >>> _parse_schedules(['1_0_32_2_0_0_1_1_11_0_1594139992_2_11_room1_13_room2', '1_0_32_0_4_0_1_1_11_0_1594139992_2_10_room3_14_room4'])
+    (False, {})
+    >>> _parse_schedules(['1_1_32_0_0_0_1_1_11_0_1594139992_2_11_room1_13_room2'])
+    (False, {})
+    >>> _parse_schedules([])
+    (False, {})
```
rytilahti, line 505: Instead of having such in docstrings, you could add some unittests to the parser for these cases. That would keep the code-base cleaner, and making sure no future changes will break the behavior when/if the support will be improved in the future!

miio/viomivacuum.py
```diff
@@ -268,124 +382,400 @@ def has_map(self) -> bool:
...
+            # "start_time",
             "suction_grade",
+            "v_state",
             "water_grade",
-            "remember_map",
-            "has_map",
-            "is_mop",
-            "has_newmap",
+            # "water_percent",
+            # "zone_data",
```
rytilahti, line 606: Should these be below the "The following list of properties .." comment?

miio/viomivacuum.py
```diff
@@ -423,22 +813,160 @@ def set_dnd(
...
+    @command(click.argument("state", type=EnumType(ViomiVoiceState)))
+    def set_sound_volume(self, state: ViomiVoiceState):
+        """Switch the voice on or off."""
+        enabled = 1
+        if state.value == 0:
+            enabled = 0
+        return self.send("set_voice", [enabled, state.value])
+
+    @command(click.argument("state", type=bool))
+    def set_remember(self, state: bool):
```
rytilahti, line 825:
```suggestion
    def set_remember_map(self, state: bool):
```

miio/viomivacuum.py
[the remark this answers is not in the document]
titilambert: done

miio/viomivacuum.py
titilambert: done

miio/viomivacuum.py
titilambert: done

titilambert:
@rytilahti I just fixed all your comments and the tests are green :)

rytilahti (approved):

rytilahti:
Okay, let's get it merged, thanks for the PR and for your patience! :tada: For further improvements, let's try to do it more gradually (i.e., with smaller, self-contained changes) to make it easier & faster to review!

k-0:
> 
> 
> I've been experimenting with getting specific rooms cleaned, and realised that my Russian hacked Mi Home app is logging the vacuum calls. Here's a successful call to vacuum the Hallway with a single pass:
> 
> ```
> 2021-01-29 15:53:50 -> {"id":163,"method":"set_mode_withroom","params":[0,1,1,15]}
> 2021-01-29 15:53:50 <- {"code":0,"id":163,"otlocalts":1611888830803511,"result":["ok"]}
> ```
> 
> As per previous, my rooms are:
> 
> ```
> jeremy@tvbox:~$ miiocli viomivacuum --ip=192.168.220.115 --token=3a5a32786e57457653576662xxxxxxxx get_rooms
> Running command get_rooms
> {'10': 'Bathroom', '12': 'Kitchen', '15': 'Hallway', '13': 'Living Room'}
> ```

I use this App too. Where can I find this log? Maybe there is a easy way to get the coordinates for zoned-cleaning.

rezmus:
create /vevs/logs/ and kill app. you will find everything (also zone clean coords) in /vevs/logs/miio/did.txt where did is your device id.

nvx0:
Could the Viomi 3 be commited? Thanks.

PR #819, opened 2020-09-18
Add support for dmaker.fan.p9 and dmaker.fan.p10

### LINKED ISSUE #721: support dmaker.fan.p9 and dmaker.fan.p10

swim2sun:
Add support for dmaker.fan.p9 and dmaker.fan.p10. 
Fixes #721

### DIFF
--- a/miio/fan.py
+++ b/miio/fan.py
@@ -1,4 +1,3 @@
-import enum
 import logging
 from typing import Any, Dict, Optional
 
@@ -6,8 +5,8 @@ import click
 
 from .click_common import EnumType, command, format_output
 from .device import Device
-from .exceptions import DeviceException
-from .miot_device import MiotDevice
+from .fan_common import FanException, LedBrightness, MoveDirection, OperationMode
+from .fan_miot import MODEL_FAN_P9, MODEL_FAN_P10, FanMiot
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -18,8 +17,6 @@ MODEL_FAN_ZA1 = "zhimi.fan.za1"
 MODEL_FAN_ZA3 = "zhimi.fan.za3"
 MODEL_FAN_ZA4 = "zhimi.fan.za4"
 MODEL_FAN_P5 = "dmaker.fan.p5"
-MODEL_FAN_P9 = "dmaker.fan.p9"
-MODEL_FAN_P10 = "dmaker.fan.p10"
 
 AVAILABLE_PROPERTIES_COMMON = [
     "angle",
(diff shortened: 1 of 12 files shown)


rytilahti (changes requested):
Thanks for the PR! Could you please move the miot implementation (and the related tests) into its own file, e.g., `fan_miot.py`?

swim2sun:
Hi, @rytilahti 
Thanks for your advice, I have try to refactor it, pls review it. 
Actually I am a java programmer, so I am not sure whether my code meets some best practices of Python

miio/fan.py
```diff
@@ -767,3 +747,27 @@ def delay_off(self, minutes: int):
     def set_rotate(self, direction: MoveDirection):
         """Rotate the fan by -5/+5 degrees left/right."""
         return self.send("m_roll", [direction.value])
+
+
+class FanP9(FanMiot):
```
rytilahti, line 752: Please move these also out to `fan_miot.py`.

miio/tests/test_fan.py
```diff
@@ -914,3 +915,170 @@ def delay_off_countdown():
 
         with pytest.raises(FanException):
             self.device.delay_off(-1)
+
+
+class DummyFanP9(DummyMiotDevice, FanP9):
```
rytilahti, line 920: Move the tests also to `test_fan_miot.py`. The non-shared code between fans and miot fans should be kept separate.

miio/fan_common.py
```diff
@@ -0,0 +1,23 @@
+import enum
```
rytilahti, line 1: Good catch with common functionalities :+1: The long-term plan is to have base classes for common device types, for both enforcing the API as well as making it easier for downstreams to integrate support for new devices.

miio/fan_miot.py
```diff
@@ -0,0 +1,285 @@
...
+}
+
+
+class OperationModeMiot(enum.Enum):
+    Normal = 0
+    Nature = 1
+
+
+class FanStatusMiot:
+    """Container for status reports from the Xiaomi Mi Smart Pedestal Fan DMaker P9/P10."""
```
rytilahti, line 51: Could you add a pretty-formatted response for the status command (use `--debug` to get the output), so that anyone without access to a device can reason about the property values.

miio/fan_miot.py
```diff
@@ -0,0 +1,285 @@
...
+                self.led,
+                self.buzzer,
+                self.child_lock,
+                self.delay_off_countdown,
+            )
+        )
+        return s
+
+    def __json__(self):
+        return self.data
```
rytilahti, line 132:
```suggestion
```

This is unnecessary now thanks to #827.

miio/fan_miot.py
```diff
@@ -0,0 +1,285 @@
...
+    def __init__(
+        self,
+        ip: str = None,
+        token: str = None,
+        start_id: int = 0,
+        debug: int = 0,
+        lazy_discover: bool = True,
+        model: str = MODEL_FAN_P10,
+    ) -> None:
+        super().__init__(MIOT_MAPPING[model], ip, token, start_id, debug, lazy_discover)
```
rytilahti, line 145: This will cause an exception if the model is not in `MIOT_MAPPING`, so it should be moved after the check below.

rytilahti (changes requested):
Hi @swim2sun, no worries, the code is looking pretty good python code to me! :-)

rytilahti:
LGTM! :+1: One small thing, could you update `discovery.py` for these?

swim2sun:
@rytilahti Thanks for your guidance!  `discovery.py` has been updated, pls check

rytilahti (approved):
Looks great, thanks again! :tada:

biscuitee:
Hi, thanks for adding dmaker.fan.p9 support.  How could I add this support into Home Assistant?

PR #825, opened 2020-10-03
Add basic support for xiaomi.aircondition.mc1, mc2, mc4, mc5

zhangjingye03:
This commit introduced basic support and tests for MIoT device
Xiaomi Mi Smart Air Conditioner A including xiaomi.aircondition.mc1,
mc2, mc4 and mc5 since they are nearly the same in miot-spec.

Tested on my xiaomi.aircondition.mc4. All properties and function work
except property `electricity' keeps zero.

Device miot-spec:
http://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:air-conditioner:0000A004:xiaomi-mc4:1

Device Mi Home plugin:
https://cdn.cnbj1.fds.api.mi-img.com/rn-plugins/2020-08-25/signed_10043_1000595_66_ANDROID_bundle_d8d30e612b4b0722e4a74744614f3c53.zip

### DIFF
--- a/miio/airconditioner_miot.py
+++ b/miio/airconditioner_miot.py
@@ -21,7 +21,7 @@ _MAPPING = {
     "dryer": {"siid": 2, "piid": 10},
     "sleep_mode": {"siid": 2, "piid": 11},
     # Fan Control (siid=3)
-    "fan_level": {"siid": 3, "piid": 2},
+    "fan_speed": {"siid": 3, "piid": 2},
     "vertical_swing": {"siid": 3, "piid": 4},
     # Environment (siid=4)
     "temperature": {"siid": 4, "piid": 7},
@@ -39,9 +39,74 @@ _MAPPING = {
     "timer": {"siid": 10, "piid": 3},
 }
 
+CLEANING_STAGES = [
+    "Stopped",
+    "Condensing water",
+    "Frosting the surface",
+    "Defrosting the surface",
+    "Drying",
+]
+
 
 class AirConditionerMiotException(DeviceException):
     pass
+
+
+class CleaningStatus:
+    def __init__(self, status: str):
+        """
+        Auto clean mode indicator.
+
+        Value format: <int>,<int>,<int>,<int>
+        Integer 1: whether auto cleaning mode started.
+        Integer 2: current progress in percent.
+        Integer 3: which stage it is currently under (see CLEANING_STAGE list).
+        Integer 4: if current operation could be cancelled.
+
+        Example auto clean indicator 1: 0,100,0,1
+        indicates the auto clean mode has finished or not started yet.
+        Example auto clean indicator 2: 1,22,1,1
+        indicates auto clean mode finished 22%, it is condensing water and can be cancelled.
+        Example auto clean indicator 3: 1,72,4,0
+        indicates auto clean mode finished 72%, it is drying and cannot be cancelled.
+
+        Only write 1 or 0 to it would start or abort the auto clean mode.
+        """
+        self.status = [int(value) for value in status.split(",")]
+
+    @property
+    def cleaning(self) -> bool:
+        return bool(self.status[0])
+
+    @property
+    def progress(self) -> int:
+        return int(self.status[1])
+
+    @property
+    def stage(self) -> str:
+        try:
+            return CLEANING_STAGES[self.status[2]]
+        except KeyError:
+            return "Unknown stage"
+
+    @property
+    def cancellable(self) -> bool:
+        return bool(self.status[3])
+
+    def __repr__(self) -> str:
+        s = (
+            "<CleaningStage cleaning=%s, "
+            "progress=%s, "
+            "stage=%s, "
+            "cancellable=%s>"
+            % (
+                self.cleaning,
+                self.progress,
+                self.stage,
+                self.cancellable,
+            )
+        )
+        return s
 
 
 class OperationMode(enum.Enum):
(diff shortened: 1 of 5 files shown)


miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+    Level2 = 2
+    Level3 = 3
+    Level4 = 4
+    Level5 = 5
+    Level6 = 6
+    Level7 = 7
+
+
+class AirConditionerMiotStatus:
+    """Container for status reports from the air conditioner which uses MIoT protocol."""
```
rytilahti, line 65: Could you add an example response to the class docstring?

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
+
+    @property
+    def is_on(self) -> bool:
+        """True if the device is turned on."""
+        return self.data["power"]
```
rytilahti, line 73: Is there a need to convert the value to bool, or is it bool already? How about other properties marked to return booleans?

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+
+    @property
+    def clean(self) -> str:
+        """Auto clean mode indicator."""
+        return self.data["clean"]
+
+    @property
+    def running_duration(self) -> float:
+        """Total running duration in hours."""
+        return self.data["running_duration"]
```
rytilahti, line 148: Would it make sense to return a timedelta here?

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+            if sleep_mode
+            else "Turning off sleep mode"
+        ),
+    )
+    def set_sleep_mode(self, sleep_mode: bool):
+        """Turn sleep mode on/off."""
+        return self.set_property("sleep_mode", sleep_mode)
+
+    @command(
+        click.argument("fan_level", type=EnumType(FanLevel, False)),
```
rytilahti, line 329:
```suggestion
        click.argument("fan_level", type=EnumType(FanLevel)),
```
`EnumType` defaults to incasesensitive since some time, so there is no need to pass False here.

rytilahti (changes requested):
Thanks for the PR, it's great to see more miot devices supported :tada: A couple of small changes are needed, but otherwise this looks to be good to go.

miio/airconditioner_miot.py
zhangjingye03: They are already bool according to miio-spec like this:
```JSON
{
	"iid": 1,
	"type": "urn:miot-spec-v2:property:on:00000006:xiaomi-mc4:1",
	"description": "Switch Status",
	"format": "bool",
	"access": [
		"read",
		"write",
		"notify"
	]
}
```

miio/airconditioner_miot.py
zhangjingye03: This property seems to return total running time since its first power on, maybe it's a statistical data for manufacture to refer. Even I unplug and re-plug it back doesn't reset this value.

miio/airconditioner_miot.py
zhangjingye03: Well, it's a MIoT device and I cannot find a example response to refer to.
Should I add something like this?
```
{
	'did': 'power',
	'siid': 2,
	'piid': 1,
	'code': 0,
	'value': False
}, {
	'did': 'mode',
	'siid': 2,
	'piid': 2,
	'code': 0,
	'value': 4
}, {
	'did': 'target_temperature',
	'siid': 2,
	'piid': 4,
	'code': 0,
	'value': 26.0
}, {
	'did': 'eco',
	'siid': 2,
	'piid': 7,
	'code': 0,
	'value': False
}, {
	'did': 'heater',
	'siid': 2,
	'piid': 9,
	'code': 0,
	'value': True
}, {
	'did': 'dryer',
	'siid': 2,
	'piid': 10,
	'code': 0,
	'value': False
}, {
	'did': 'sleep_mode',
	'siid': 2,
	'piid': 11,
	'code': 0,
	'value': False
}, {
	'did': 'fan_level',
	'siid': 3,
	'piid': 2,
	'code': 0,
	'value': 0
}, {
	'did': 'vertical_swing',
	'siid': 3,
	'piid': 4,
	'code': 0,
	'value': True
}, {
	'did': 'temperature',
	'siid': 4,
	'piid': 7,
	'code': 0,
	'value': 27.6
}, {
	'did': 'buzzer',
	'siid': 5,
	'piid': 1,
	'code': 0,
	'value': True
}, {
	'did': 'led',
	'siid': 6,
	'piid': 1,
	'code': 0,
	'value': True
}, {
	'did': 'electricity',
	'siid': 8,
	'piid': 1,
	'code': 0,
	'value': 0.0
}, {
	'did': 'clean',
	'siid': 9,
	'piid': 1,
	'code': 0,
	'value': '0,0,1,1'
}, {
	'did': 'running_duration',
	'siid': 9,
	'piid': 5,
	'code': 0,
	'value': 210.4
}
```

rytilahti: That's quite verbose, but I suppose it's fine. After all, the idea is that the data structure should also be understandable for someone with no access to the device. You can use the `--debug` flag for miiocli to get the data that gets returned by the device, and just format it.

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+
+
+class OperationMode(enum.Enum):
+    Cool = 2
+    Dry = 3
+    Fan = 4
+    Heat = 5
+
+
+class FanLevel(enum.Enum):
```
rytilahti, line 53:
```suggestion
class FanSpeed(enum.Enum):
```
would make the naming consistent with airconditioning*.

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+        """True if aux dryer mode is on."""
+        return self.data["dryer"]
+
+    @property
+    def sleep_mode(self) -> bool:
+        """True if sleep mode is on."""
+        return self.data["sleep_mode"]
+
+    @property
+    def fan_level(self) -> FanLevel:
```
rytilahti, line 111:
```suggestion
    def fan_speed(self) -> FanSpeed:
```

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+        """Auto clean mode indicator."""
+        return self.data["clean"]
+
+    @property
+    def running_duration(self) -> float:
+        """Total running duration in hours."""
+        return self.data["running_duration"]
+
+    @property
+    def fan_percent(self) -> int:
```
rytilahti, line 151:
```suggestion
    def fan_speed_percent(self) -> int:
```

miio/airconditioner_miot.py
rytilahti: Maybe the property should then be renamed to `total_running_time` or alike? In any case, I think this should be returned as `timedelta`.

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+        return self.data["led"]
+
+    @property
+    def electricity(self) -> float:
+        """Power consumption accumulation in kWh."""
+        return self.data["electricity"]
+
+    @property
+    def clean(self) -> str:
+        """Auto clean mode indicator."""
```
rytilahti, line 142: What information does the returned information contain? The test uses `0,100,1,1`, can we parse this information to return something more usable?

miio/airconditioner_miot.py
```diff
@@ -0,0 +1,403 @@
...
+
+    @property
+    def fan_percent(self) -> int:
+        """Current fan level in percent."""
+        return self.data["fan_percent"]
+
+    @property
+    def timer(self) -> str:
+        """Timer indicator."""
+        return self.data["timer"]
```
rytilahti, line 158: The same as with `clean`. Any ideas about the payload values?

miio/airconditioner_miot.py
zhangjingye03: I've added detailed explanation in the source file.

syssi (approved):
LGTM! Let's merge this new device!

rytilahti (approved):
Okay, let's get it merged! :tada:

PR #828, opened 2020-10-05
Correct importlib_metadata python_version bounds

jonringer:
importlib_metadata is a backport of the python 3.8 stdlib library

https://importlib-metadata.readthedocs.io/en/latest/

related: https://github.com/NixOS/nixpkgs/pull/99585

### DIFF
--- a/miio/__init__.py
+++ b/miio/__init__.py
@@ -1,5 +1,11 @@
 # flake8: noqa
-from importlib_metadata import version  # type: ignore
+try:
+    # python 3.7 and earlier
+    from importlib_metadata import version  # type: ignore
+except ImportError:
+    # python 3.8 and later
+    from importlib.metadata import version  # type: ignore
+
 from miio.airconditioningcompanion import (
     AirConditioningCompanion,
     AirConditioningCompanionV3,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ appdirs = "^1"
 tqdm = "^4"
 netifaces = "^0"
 android_backup = { version = "^0", optional = true }
-importlib_metadata = "^1"
+importlib_metadata = { version = "^1", markers = "python_version <= '3.7'" }
 croniter = "^0"
 
 sphinx = { version = "^3", optional = true }


georgewhewell:
Thanks for doing this! But i think we also need to change the import, it's `importlib.metadata` in py38

```
try:
    from importlib_metadata import version
except ImportError:
    from importlib.metadata import version
```

jonringer:
applied changes :)

rytilahti:
I actually wanted to use the lib to avoid try-except "hacking", and this has been working on at least on python3.8 just fine. Do you think there is any cons in this approach (besides requiring you to package importlib-metadata, too)? See also https://gitlab.com/python-devs/importlib_metadata/-/issues/130

Anyways, I'm open to merge this as it is as it is not such a big deal.

jonringer:
It is a backport, and could eventually diverge similar to how the `typing` package is now.

Another "goal" in nixpkgs is to have minimal closures, and the package could be deemed "unnecessary", and does add 280K (isn't much, but also isn't nothing)

```
$ du -hd0 /nix/store/zachx9ik7jw6bwnl00mjf940s5pfz8xl-python3.8-importlib-metadata-1.7.0
280K	/nix/store/zachx9ik7jw6bwnl00mjf940s5pfz8xl-python3.8-importlib-metadata-1.7.0
```

rytilahti:
Yeah, good points, let's merge this!

georgewhewell:
@rytilahti thanks! do you plan to release a new version including this change any time soon?

rytilahti:
I hope I can prepare a new release soonish (this weekend, perhaps). I'm waiting responses for #808, #825, #826 to see if we can get those also merged before a release.

jonringer:
@georgewhewell we could always fetch this commit as a patch. Upstream sources should be able to determine their own release timelines :)

Although a patch bump would be appreciated ;)

rytilahti:
@georgewhewell just a quick ping that I forgot to do earlier: 0.5.4 with this fix was released a week ago: https://github.com/rytilahti/python-miio/releases/tag/0.5.4

georgewhewell:
@rytilahti great, thanks for this version (and ping) :+1:

PR #831, opened 2020-10-09
Throwing GatewayException in get_illumination

javicalle:
I'm not sure if the GatewayException type is only suitable for SubDevice or is also allowed here, but it is the exception HA expects in this case:
https://github.com/home-assistant/core/blob/985e4e1bd942ef4ab56617a77f59baeec10649c5/homeassistant/components/xiaomi_miio/sensor.py#L320:L325

### DIFF
--- a/miio/gateway.py
+++ b/miio/gateway.py
@@ -386,7 +386,7 @@ class Gateway(Device):
             raise GatewayException(
                 "Got an exception while getting gateway illumination"
             ) from ex
-    
+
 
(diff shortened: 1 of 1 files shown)


rytilahti:
LGTM :+1: @starkillerOG do you have an opinion on this?

rytilahti (approved):

starkillerOG:
LGTM, thanks @javicalle.
Can be merged.

PR #842, opened 2020-10-25
Speed settings name update

VivantSenior:
Name change of speed settings to be more aligned with Xiaomi Home app.


rytilahti (commented):
I don't think we should change the naming, simply to be consistent with the v1 fanspeed levels. Changing these would break the API and make it more confusing for anyone using the library to control both types of devices.

PR #845, opened 2020-10-28
Vacuum: Implement TUI for the manual mode

rnovatorov:
I recently discovered this project and would like to thank you - it is awesome :)

I played with it a bit and noticed that the vacuum manual mode is difficult to use interactively via CLI. One has to enter somewhat lengthy (more than a char) commands and arguments, which might feel clumsy if a real-time response from the device is desired.

This PR adds a tiny `curses` based TUI to allow for a real-time movement control. I have not thoroughly explored the repo, so the proposed solution might not be optimal.

### DIFF
--- a/miio/__init__.py
+++ b/miio/__init__.py
@@ -42,6 +42,7 @@ from miio.protocol import Message, Utils
 from miio.pwzn_relay import PwznRelay
 from miio.toiletlid import Toiletlid
 from miio.vacuum import Vacuum, VacuumException
+from miio.vacuum_tui import VacuumTUI
 from miio.vacuumcontainers import (
     CleaningDetails,
     CleaningSummary,
--- a/miio/vacuum_cli.py
+++ b/miio/vacuum_cli.py
@@ -233,6 +233,13 @@ def manual(vac: miio.Vacuum):
 
 @manual.command()
 @pass_dev
+def tui(vac: miio.Vacuum):
+    """TUI for the manual mode."""
+    miio.VacuumTUI(vac).run()
+
+
+@manual.command()
+@pass_dev
 def start(vac: miio.Vacuum):  # noqa: F811  # redef of start
     """Activate the manual mode."""
     click.echo("Activating manual controls")
--- a/miio/vacuum_tui.py
+++ b/miio/vacuum_tui.py
@@ -1,7 +1,7 @@
 try:
     import curses
-except ImportError:  # curses unavailable
-    pass
+except ImportError:
+    curses = None
 
 import enum
 from typing import Tuple
@@ -24,6 +24,9 @@ class Control(enum.Enum):
 
 class VacuumTUI:
     def __init__(self, vac: Vacuum):
+        if curses is None:
+            raise ImportError("curses library is not available")
+
         self.vac = vac
         self.rot = 0
         self.rot_delta = 30
--- a/miio/vacuum_tui.py
+++ b/miio/vacuum_tui.py
@@ -1,0 +1,101 @@
+try:
+    import curses
+except ImportError:  # curses unavailable
+    pass
+
+import enum
+from typing import Tuple
+
+from .vacuum import Vacuum
+
+
+class Control(enum.Enum):
+
+    Quit = "q"
+    Forward = "w"
+    ForwardFast = "W"
+    Backward = "s"
+    BackwardFast = "S"
+    Left = "a"
+    LeftFast = "A"
+    Right = "d"
+    RightFast = "D"
+
+
+class VacuumTUI:
+    def __init__(self, vac: Vacuum):
+        self.vac = vac
+        self.rot = 0
+        self.rot_delta = 30
+        self.rot_min = Vacuum.MANUAL_ROTATION_MIN
+        self.rot_max = Vacuum.MANUAL_ROTATION_MAX
+        self.vel = 0.0
+        self.vel_delta = 0.1
+        self.vel_min = Vacuum.MANUAL_VELOCITY_MIN
+        self.vel_max = Vacuum.MANUAL_VELOCITY_MAX
+        self.dur = 10 * 1000
+
+    def run(self) -> None:
+        self.vac.manual_start()
+        try:
+            curses.wrapper(self.main)
+        finally:
+            self.vac.manual_stop()
+
+    def main(self, screen) -> None:
+        screen.addstr("Use wasd to control the device.\n")
+        screen.addstr("Hold shift to enable fast mode.\n")
+        screen.addstr("Press q to quit.\n")
+        screen.refresh()
+        self.loop(screen)
+
+    def loop(self, win) -> None:
+        done = False
+        while not done:
+            key = win.getkey()
+            text, done = self.handle_key(key)
+            win.clear()
+            win.addstr(text)
+            win.refresh()
+
+    def handle_key(self, key: str) -> Tuple[str, bool]:
+        try:
+            ctl = Control(key)
+        except ValueError as e:
+            return "Ignoring %s: %s.\n" % (key, e), False
+
+        done = self.dispatch_control(ctl)
+        return self.info(), done
+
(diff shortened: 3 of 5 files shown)


miio/vacuum_tui.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Quit = "q"
+    Forward = "l"
+    ForwardFast = "L"
+    Backward = "k"
+    BackwardFast = "K"
+    Left = "j"
+    LeftFast = "J"
+    Right = ";"
+    RightFast = ":"
```
rytilahti, line 22: Rather than using obscure key bindings, how about using the arrow keys & shift as a modifier for the fast modes?

miio/vacuum_tui.py
```diff
@@ -0,0 +1,95 @@
+try:
+    import curses
+except ImportError:
+    pass
```
rytilahti, line 4: When can this happen?

miio/vacuum_tui.py
```diff
@@ -0,0 +1,95 @@
...
+    def main(self, screen) -> None:
+        done = False
+        while not done:
+            key = screen.getkey()
+            text, done = self.handle(key)
+            screen.clear()
+            screen.addstr(text)
+            screen.refresh()
+
+    def handle(self, key: str) -> Tuple[str, bool]:
```
rytilahti, line 56:
```suggestion
    def handle_key(self, key: str) -> Tuple[str, bool]:
```

miio/vacuum_tui.py
```diff
@@ -0,0 +1,95 @@
...
+        except ValueError as e:
+            return "ignoring %s: %s\n" % (key, e), False
+
+        if ctl == Control.Quit:
+            return "exiting normally\n", True
+
+        self.control(ctl)
+        return self.info(), False
+
+    def control(self, ctl: Control) -> None:
```
rytilahti, line 68: Maybe `handle_control` or `adjust_control`?

miio/vacuum_tui.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def run(cls: Type["VacuumTUI"], vac: Vacuum) -> None:
+        vac.manual_start()
+        try:
+            tui = cls(vac=vac)
+            curses.wrapper(tui.main)
+        finally:
+            vac.manual_stop()
+
+    def main(self, screen) -> None:
```
rytilahti, line 47:
```suggestion
    def loop(self, screen) -> None:
```
or `main_loop`?

miio/vacuum_tui.py
```diff
@@ -0,0 +1,95 @@
...
+        self.rot_min = Vacuum.MANUAL_ROTATION_MIN
+        self.rot_max = Vacuum.MANUAL_ROTATION_MAX
+        self.vel = 0.0
+        self.vel_delta = 0.1
+        self.vel_min = Vacuum.MANUAL_VELOCITY_MIN
+        self.vel_max = Vacuum.MANUAL_VELOCITY_MAX
+        self.dur = 10 * 1000
+
+    @classmethod
+    def run(cls: Type["VacuumTUI"], vac: Vacuum) -> None:
```
rytilahti, line 39: How about defining this as a regular method?

rytilahti (changes requested):
Hi and thanks for the PR! The manual mode is indeed quite cumbersome and this looks like a great addition :-)

I added some comments inline, but it would also be great if you could display the key bindings (either on start as a dialog, or print them out to the screen?).

Also, please update the `docs/vacuum.rst` to make it easier to find that such feature exists.

miio/vacuum_tui.py
rnovatorov: Renamed as suggested.

miio/vacuum_tui.py
rnovatorov: Renamed to `main_loop`.

miio/vacuum_tui.py
rnovatorov: Defined as a regular method.

miio/vacuum_tui.py
rnovatorov: What we import here as `curses` is actually an extension module that must be backed by one of few C implementations of a `curses` library. `ImportError` occurs when none of the implementations are available on the system.

I do not think there is much we can do about it except for mentioning in `docs/vacuum.rst`, that `curses` lib is a requirement for TUI.

miio/vacuum_tui.py
rnovatorov: Moved the check for `Control.Quit` inside this method and renamed it to `dispatch_control`.

miio/vacuum_tui.py
rnovatorov: Current bindings are Vim-ish.

Arrow keys work fine but lack support for the fast mode. The reason is that for `curses` there are no such things as Shift+ArrowUp or Shift+ArrowDown (Shift+ArrowLeft and Shift+ArrowRight are OK though).

Shall go with WASD?

miio/vacuum_tui.py
rnovatorov: Ended up splitting the method into two:
- `loop` handles keys
- `main` prints usage and starts the loop

rnovatorov:
Added fixes according to inline comments and usage printing before the main loop. Also updated docs to mention the feature.

miio/vacuum_tui.py
rytilahti: Hmm, I think WASD is a better choice (as `;` and `:` can be problematic on some keyboard layouts), not that everything is QWERTY either, but it's still a commonly used control scheme...

rnovatorov: Alright, changed key bindings to WASD.

miio/vacuum_tui.py
rytilahti: Ah, I see. I thought it is part of the standard library and always available.

How about letting it bubble up and doing the import inside the `tui()` command? It could be caught there to display a user-friendly error message.

miio/vacuum_cli.py
```diff
@@ -231,6 +231,13 @@ def manual(vac: miio.Vacuum):
     # if not vac.manual_mode and command :
 
 
+@manual.command()
+@pass_dev
+def tui(vac: miio.Vacuum):
+    """TUI for the manual mode."""
+    miio.VacuumTUI(vac).run()
```
rytilahti, line 238: Wrt the previous comment on curses import errors, this would be:
```
from miio.vacuum_tui import VacuumTUI
try:
    tui = VacuumTUI(vac)
    tui.run()
except Exception as ex:
    click.echo("Error during TUI: %s" % ex)
```

miio/__init__.py
```diff
@@ -42,6 +42,7 @@
 from miio.pwzn_relay import PwznRelay
 from miio.toiletlid import Toiletlid
 from miio.vacuum import Vacuum, VacuumException
+from miio.vacuum_tui import VacuumTUI
```
rytilahti, line 45:
```suggestion
```

miio/vacuum_tui.py
rnovatorov: Personally, I prefer all the `import`s to be grouped in one place - at the top of a module. They are easier to manage that way.

If all we want is a prettier error message, I would rather delay the exception until `VacuumTUI` is instantiated:

```python
try:
    import curses
except ImportError:
    curses = None

...

class VacuumTUI:
    def __new__(cls, *args, **kwargs):
        if curses is None:
            raise ImportError("curses library is not available")
        return super().__new__(cls)
```

What do you think?

rytilahti: So what I want is it not to simply crash silently in any case, so your suggested solution works fine, too! Maybe check for the Noneness inside the `__init__` though?

rnovatorov: Moved the check to `__init__` in a new commit.

rytilahti (approved):
Looks good to go, thanks again @rnovatorov! :tada:

PR #850, opened 2020-11-04
Add basic dmaker.fan.p11 support

syssi:
I'm trying to get this fan supported in 0.5.4 already.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,7 +105,7 @@ Supported devices
 -  Xiaomi Philips Zhirui Smart LED Bulb E14 Candle Lamp
 -  Xiaomi Philips Zhirui Bedroom Smart Lamp
 -  Xiaomi Universal IR Remote Controller (Chuangmi IR)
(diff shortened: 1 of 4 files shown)


iquix:
I own this fan model (dmaker.fan.p11), and it seems that 'status' siid:2, piid:6 is actually fan speed of level 1~100
I have already tested this, with swim2sun's code.

Therefore, line #50 of fan_miot.py should be
"fan_speed": {"siid": 2, "piid": 6},

Thank you.

syssi:
@iquix Thanks for you support! I will fix the issue right away.

syssi:
@iquix Could you provide the model no. your of device? I want to use it here soon: https://github.com/syssi/xiaomi_airpurifier#supported-devices

iquix:
The brand is mijia and the model no. is BPLDS03DM.
Thank you so much.
https://www.google.com/search?q=mijia+BPLDS03DM

PR #853, opened 2020-11-05
Fix PTC support of the dmaker.airfresh.t2017

syssi:
Breaking change: `PtcLevel.Off` removed. Method `set_ptc(True)` / `set_ptc(False)` introduced.


syssi:
I've removed the breaking change tag because the `PtcLevel.Off` option never worked.

PR #854, opened 2020-11-05
Fix payload of all dmaker.airfresh.t2017 toggles

syssi:


rytilahti:
Did the API change over time, or was this always broken? Feel free to merge as you wish @syssi

syssi:
It was always broken.

PR #862, opened 2020-11-16
Add dmaker.airfresh.a1 support

syssi:
Properties:

- pm25
- co2
- temperature_outside
- favourite_speed
- filter_rate
- filter_day
- control_speed
- power
- mode
- ptc_on
- ptc_status
- child_lock
- sound
- display

Supported methods:

- set_power [power] on/off
- set_mode [mode] sleep/auto/favourite
- set_favourite_speed 0...150
- set_ptc_on [on/off]
- set_display [light] on/off(?)
- set_sound [sound] on/off
- set_child_lock [lock] on/off
- set_filter_rate [100]

Not implemented methods:

- get_sensor [sensor]
- get_timer []
- get_ptc_timer []
- set_ptc_timer [ptcTime]
- set_timer [time]
- get_light_timer []
- set_light_timer [lightTime]
- delete_timer [time]


miio/airfresh_t2017.py
```diff
@@ -134,32 +162,60 @@ def control_speed(self) -> int:
     @property
     def dust_filter_life_remaining(self) -> int:
         """Remaining dust filter life in percent."""
-        return self.data["filter_intermediate"]
+        if (
+            "filter_intermediate" in self.data
+            and self.data["filter_intermediate"] is not None
+        ):
+            return self.data["filter_intermediate"]
```
rytilahti, line 169: This could be simplified by simply checking for both values? Returns none if neither of the property is there, so the return value should be marked as Optional.
```
# filter_intermediate for model X, filter_rate for A1
return self.data.get("filter_intermediate", self.data.get("filter_rate"))
```

miio/airfresh_t2017.py
```diff
@@ -134,32 +162,60 @@ def control_speed(self) -> int:
...
 
     @property
     def dust_filter_life_remaining_days(self) -> int:
         """Remaining dust filter life in days."""
-        return self.data["filter_inter_day"]
+        if (
+            "filter_inter_day" in self.data
+            and self.data["filter_inter_day"] is not None
+        ):
+            return self.data["filter_inter_day"]
```
rytilahti, line 182: See my previous comment on checking for both values at once. This applies also for other properties that I will not comment separately to keep the review shorter.

miio/airfresh_t2017.py
```diff
@@ -182,9 +238,15 @@ def display(self) -> bool:
         return self.data["display"]
 
     @property
-    def display_orientation(self) -> int:
+    def display_orientation(self) -> Optional[int]:
```
rytilahti, line 241: Hmm, looks like the type is wrong here (should be DisplayOrientation, right?).

miio/airfresh_t2017.py
```diff
@@ -134,32 +162,60 @@ def control_speed(self) -> int:
...
     @property
     def ptc(self) -> bool:
         """Return True if PTC is on."""
         return self.data["ptc_on"]
 
     @property
-    def ptc_level(self) -> int:
+    def ptc_level(self) -> Optional[int]:
         """PTC level."""
-        return PtcLevel(self.data["ptc_level"])
```
rytilahti, line 162: Maybe this sort of structure would work for all properties that return an `Optional[SomeEnumOrContainerType]` (btw, the current type hint is wrong):
```
try:
    return PtcLevel(self.data["ptc_level"])
except KeyError:
    return None

File: miio/airfresh_t2017.py [hunk 5]
```diff
@@ -397,3 +434,88 @@ def get_ptc_timer(self):
...
+            "Child lock: {result.child_lock}\n"
+            "Buzzer: {result.buzzer}\n"
+            "Display: {result.display}\n"
+            "Display orientation: {result.display_orientation}\n",
+        )
+    )
+    def status(self) -> AirFreshStatus:
+        """Retrieve properties."""
+
+        return super().status()
```
rytilahti, line 484:
```suggestion
```
This could be removed as no-op, right?

syssi: I added this no-op method because the @command output is slightly changed. I think

syssi: Do you agree? :-)

rytilahti: Oh, I see, I think that's fine then :-)

Generally speaking, I think we should start thinking how to move away from hard-coded responses (maybe simply using the status classes' property markings?) to make it less error-prone and easier to maintain.. Considering that will affect the whole code-base, it is not the task for this PR, just something to keep in mind :-)

PR #865, opened 2020-11-22
Add airdog.airpurifier.{x3,x5,x7sm} support

syssi:


miio/airpurifier_airdog.py
```diff
@@ -0,0 +1,212 @@
...
+        default_output=format_output(
+            lambda lock: "Turning on child lock" if lock else "Turning off child lock"
+        ),
+    )
+    def set_child_lock(self, lock: bool):
+        """Set child lock on/off."""
+        if lock:
+            return self.send("set_lock", [1])
+        else:
+            return self.send("set_lock", [0])
```
rytilahti, line 207:
```suggestion
        return self.send("set_lock", [int(lock)])
```

rytilahti (approved):
Looks good to me, feel free to merge as you wish :+1:

alexeypetrenko:
@syssi, I've got a small feedback regarding this module.

For this device `sleep` does not mean that it is off or `idle`. Probably, more precise term would be `night mode`. In this mode fan still runs (even slower than on the first speed), display and all the lights are off.

If you don't mind, I'd like to rename `Idle` to something more appropriate.

PR #876, opened 2020-11-30
Export MiotDevice for miio module

syssi:
Is there a rule for adding classes to the `__init__.py`?


rytilahti:
There is no clear rule at the moment, except that it should be useful for a library user alone (so e.g., status containers are out of the picture). MiotDevice is useful to have there, as the `miot_info` (#672) and maybe other helper methods can be useful by themselves.

Issue #875, opened 2020-11-30
JSON encode list without spaces to retrieve multiple properties at once


TheDJVG:
When working on issue https://github.com/rytilahti/python-miio/issues/874 I noticed that the Xiaomi Home app sends the `get_prop` params without spaces in between the params to retrieve multiple properties at once. It's possible that this applies to other devices too that now have to request properties one by one.

With the space in between it will fail with an error (the bool at the end controls if there should be spaces):
```
# miiocli device  --ip 10.80.80.233 --token ${token} raw_command get_prop '["program","power"]' False 
Running command raw_command
[2, 'error']
```

Without spaces:
```
miiocli device  --ip 10.80.80.233 --token ${token} raw_command get_prop '["program","power"]' True 
Running command raw_command
[2, 0]
```
Currently achieved by altering https://github.com/rytilahti/python-miio/blob/34a4358b2c908c9ec51d7058e54d4e18bc281a99/miio/protocol.py#L153-L160 to:
```python
    def _encode(self, obj, context, path):
        """Encrypt the given payload with the token stored in the context.

        :param obj: JSON object to encrypt"""
        # pp(context)
        if context["_"].get("no_spaces_in_list") and isinstance(obj.get('params'), (list, set, tuple)):
            json_obj = json.dumps(obj).replace(', ', ',')
        else:
            json_obj = json.dumps(obj)
        return Utils.encrypt(
            json_obj.encode('utf-8') + b"\x00", context["_"]["token"]
        )
```
<sub>I'm sure this is not the best way.</sub>

**Describe the solution you'd like**
Currently I added a `no_spaces_in_list` argument to all the different `.send()` methods to get this specific request in `EncryptionAdapter._encode()` and I tried creating a custom `json.JSONEncoder` encoder for `list` but the standard `json.dumps` is too smart when it encounters a `list` (or where `list` is a parent class) and does not call the default method of the custom encoder class. 

Would be great if we in some way can control the spaces because it greatly speeds up if you can request multiple properties at once.

**Device information:**
See https://github.com/rytilahti/python-miio/issues/874

**Additional context**

Let me know if any specific debug information is required to help with this request. Happy to implement a solution but not sure what the best way would be at this moment in time.

TheDJVG:
How about using the `separators` argument to `json.dumps`?
If we would take this `test_obj` dictionary:
```python
>>> pprint(test_obj)
{'test_dict': {'dict_key': 'dict_value'},
 'test_int': 1337,
[...]

PR #877, opened 2020-12-01
Add support for the Xiaomi/Viomi Dishwasher (viomi.dishwasher.m02)

### LINKED ISSUE #874: Xiaomi Dishwasher (viomi.dishwasher.m02)
**Device information:**

  - Name(s) of the device: Mi Smart Dishwasher 4 sets
  - Link: https://www.mi.com/buy/detail?product_id=11624
  - Model: viomi.dishwasher.m02
  - Hardware version: esp8266
  - Firmware version: 2.1.0

**Additional context**
Hooked up a bus pirate to see what is being send over the wifi chip (conveniently there's place for a header on there) and it display all the possible properties.

### Information I know so far
Property: **programInfo**
- Shows all the available programs, my machine:
```
['0-7102-75|1-7702-75|2-1675-55|3-7522-75|4-6930-60|5-0-0|6-0-0|7-8295-75|8-0-0|9-0-0|248-1800-0|249-4198-75']
```
Based on this information the program's are defined by 3 sections, divided by a `-`.
The first one is the program number, I was able to match these with the programs in the app / on the device:
- 0: Standard
- 1: Eco
- 2: Quick
- 3: Intensive
- 4: Glassware
- 7: Sterilize

The other program(5, 6 ,8 ,9, 248 and 249) are not super clear to me. But I can imagine maybe this is for some factory tests?
For fun I started program 5 and the app sees it as "standard" and the timer always show "00:01" remaining.
When I start program 248 it seems to only run the dry cycle for 30 minutes (which I cannot test as my heating element is broken).

The second part is the total run time of the program in seconds with the last one being the temperature for the program.

Property: **program**
With this you see the current running program (or the last selected program).

You can set (and start) the program with the `set_program` command and a number from `programInfo` as argument. It will start immediately when called (if the door is open it will start when the door is closed).

Property: **wash_status**
Returns the current status of the dishwasher
- 0: Turned off 
- 1: Turn dishwasher on (or stop the current program)
- 2: Start or continue the current program
- 3: Pause the dishwasher

You can set the current status with the `set_wash_status` command and than an option as listed above.

Property: **wash_info** - Lots of unknowns!
Returns information about the current wash. Again divided by `-`

Example output: `['49-1440-2']`
The first number seems to be the seconds it's already running (in this case 49 seconds). The second part seems to be unknown to me, but _maybe_ it's the time when the process will change to the next step? The last part seems to be the current active program.

Property: **wash_process** - Lots of unknowns!
Returns information about the current status of the machine (e.g. what part of the program it's in).

States I've observed so far:
- 0: Standby
- 1: Prewash
- 2: Wash
- 3: Rinse
- 4: Drying

Property: **run_status**

As far as I can see this returns 0 when the door is closed and 128 when the door is open.

Property: **power**

The current status of the device with 1 being on and 0 being off.

You can turn the dishwasher on and off with the `set_power` command

Property: **left_time**
Returns the time in seconds that's left for the current program. 0 when no program is running or when it's finished.

Property: **wash_temp**
Current temperature inside the dishwasher in degrees celsius. Based on a 100k 25c NTC thermistor.

Property: **wash_done_appointment**
Returns information about the next scheduled run for the dishwasher. It's a timestamp in the machine's timezone.

You can set the next schedule by sending the `set_wash_done_appointment` command with `"<timestamp>,<program_number>"` (e.g: `"1606653301,1"`)

If you want to stop the schedule you have to send `"0,<program_number>"` where program number can be any program you want (it will not change to that program after cancelling it).  I could also send anything in place of that program_number when cancelling it.
**Important** to note is that you have to cancel the schedule first before you can schedule a new time. 

Property: **freshdry_interval**
This shows how often the machine should run the fan to refresh the air inside the machine. When it's disabled it shows `0`, when enabled through the app it's set to `8`. It's not clear if this is minutes or something else because I tried to set it to 1 and it didn't start after 1 minute. The maximum value is 255.

You can change the value by using the `set_freshdry_interval_t` command and sending an integer as argument.

Property: **child_lock**
Shows if the child lock is enabled `0` for disabled, `1` for enabled.

You can set this by using the `set_child_lock` command and `0` or `1` as argument.

### Other observations

**Getting properties in a list**
It seems that getting multiple properties with `get_prop` is supported but only if there's no spaces in the list of properties being requested.

This seems to work: `get_prop "wash_status","power"`
This does not work: `get prop "wash_status", "power"`

It looks like the miio cli converts the list to include spaces which causes the result to be:
```
Running command raw_command
[1, 'error']
```

**Internal communication**
It looks like the devices uses two boards to communicate. One with the wifi chip and user interface and the second one that actually executes the commands being sent by the first board. As I've found another RX/TX there I will see if I can get some useful information on the internal bus too (hack it to do even more things?).

**Device token**
The token is refreshed once you pair it again or change the Wireless connection information.

**Unknowns**
Would love to see if it's possible to execute a part of a program (dry for longer for example, but this could maybe be enforced by running program 249).

### Final words

I only spend time so far reverse engineering the properties and commands and did not start with the miio integration yet but would like to keep this issue for that specific work.

Feel free to let me know if you need anything specific tested. My machine is completely apart as the heater is broken so I don't mind probing around and get more information!

TheDJVG:
Supported commands
- Power on/off.
- Start/stop/pause/continue a program.
- Schedule a program for a specific time (and cancel it).
- Retrieve current status + errors of the dishwasher.
- Child lock enable/disable.
- Set air refresh interval.

fix #874

### DIFF
--- a/miio/viomidishwasher.py
+++ b/miio/viomidishwasher.py
@@ -1,0 +1,370 @@
+import enum
+import logging
+from collections import defaultdict
+from datetime import datetime, timedelta
+from typing import Any, Dict, List, Optional
+
+import click
+
+from .click_common import EnumType, command, format_output
+from .device import Device
+from .exceptions import DeviceException
+
+_LOGGER = logging.getLogger(__name__)
+
+MODEL_DISWAHSER_M02 = "viomi.dishwasher.m02"
+
+MODELS_SUPPORTED = [MODEL_DISWAHSER_M02]
+
+
+class MachineStatus(enum.IntEnum):
+    Off = 0
+    On = 1
+    Running = 2
+    Paused = 3
+    Scheduled = 5
+
+
+class ProgramStatus(enum.IntEnum):
+    Standby = 0
+    Prewash = 1
+    Wash = 2
+    Rinse = 3
+    Drying = 4
+    Unknown = -1
+
+
+class Program(enum.IntEnum):
+    Standard = 0
+    Eco = 1
+    Quick = 2
+    Intensive = 3
+    Glassware = 4
+    Sterilize = 7
+    Unknown = -1
+
+    @property
+    def run_time(self):
+        return ProgramRunTime[self.value]
+
+
+ProgramRunTime = {
+    Program.Standard: 7102,
+    Program.Eco: 7702,
+    Program.Quick: 1675,
+    Program.Intensive: 7522,
+    Program.Glassware: 6930,
+    Program.Sterilize: 8295,
+}
+
+
+class ChildLockStatus(enum.IntEnum):
+    Enabled = 1
+    Disabled = 0
+
+
+class DoorStatus(enum.IntEnum):
+    Open = 128
+    Closed = 0
+
+
+class SystemStatus(enum.IntEnum):
+    WaterLeak = 1
+    InsufficientWaterFlow = 4
+    InternalConnectionError = 9
+    ThermistorError = 32
+    InsufficientWaterSoftener = 512
+    HeatingElementError = 2048
+
+
+class PowerStatus(enum.IntEnum):
+    On = 1
+    Off = 0
+
+
+class ViomiDishwasherStatus:
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
+
+    @property
+    def child_lock(self) -> ChildLockStatus:
+        return ChildLockStatus(self.data["child_lock"])
+
+    @property
+    def program(self) -> Program:
+        try:
+            return Program(self.data["program"])
+        except ValueError:
+            return Program.Unknown
+
+    @property
+    def door(self) -> DoorStatus:
+        return DoorStatus(self.data["run_status"] & (1 << 7))
+
+    @property
+    def system_status_raw(self) -> [int, SystemStatus]:
+        return self.data["run_status"]
+
+    @property
+    def status(self) -> MachineStatus:
+        return MachineStatus(self.data["wash_status"])
+
+    @property
+    def temperature(self) -> int:
+        return self.data["wash_temp"]
+
+    @property
+    def power(self) -> PowerStatus:
+        return PowerStatus(self.data["power"])
+
+    @property
+    def time_left(self) -> int:
+        return self.data["left_time"]
+
+    @property
+    def schedule(self) -> [datetime, None]:
+        schedule = self.data["wash_done_appointment"]
+        return datetime.fromtimestamp(schedule) if schedule else None
+
+    @property
+    def air_refresh_interval(self) -> int:
+        return self.data["freshdry_interval"]
+
+    @property
+    def program_progress(self) -> ProgramStatus:
+        try:
+            return ProgramStatus(self.data["wash_process"])
+        except ValueError:
+            return ProgramStatus.Unknown
+
+    @property
+    def errors(self) -> List[Optional[SystemStatus]]:
+        errors = []
+        if self.data["run_status"] & (1 << 0):
+            errors.append(SystemStatus.WaterLeak)
+        if self.data["run_status"] & (1 << 3):
+            errors.append(SystemStatus.InternalConnectionError)
+        if self.data["run_status"] & (1 << 2):
+            errors.append(SystemStatus.InsufficientWaterFlow)
+        if self.data["run_status"] & (1 << 5):
+            errors.append(SystemStatus.ThermistorError)
+        if self.data["run_status"] & (1 << 9):
+            errors.append(SystemStatus.InsufficientWaterSoftener)
+        if self.data["run_status"] & (1 << 11):
+            errors.append(SystemStatus.HeatingElementError)
+
+        return errors
+
+    def __repr__(self) -> str:
+        return (
+            f"<ViomiDishwasherStatus "
+            f"child_lock={self.child_lock}, "
+            f"program={self.program}, "
+            f"door={self.door}, "
+            f"system_status_raw={self.system_status_raw}, "
+            f"status={self.status}, "
+            f"temperature={self.temperature}, "
+            f"power={self.power}, "
+            f"time_left={self.time_left}, "
+            f"schedule={self.schedule}, "
+            f"air_refresh_interval={self.air_refresh_interval}, "
+            f"program_progress={self.program_progress}, "
+            f"errors={self.errors}>"
+        )
+
+
+class ViomiDishwasher(Device):
+    """Main class representing the dishwasher."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        model = self.info().model
+        if model not in MODELS_SUPPORTED:
+            _LOGGER.warning(
+                "Dishwasher model '%s' has not been tested. Please continue with caution and open a GitHub issue to get this model supported.",
+                model,
+            )
+
+    @command(
+        default_output=format_output(
+            "",
+            "Program: {result.program.name}\n"
+            "Program state: {result.program_progress.name}\n"
+            "Program time left: {result.time_left}\n"
+            "Dishwasher status: {result.status.name}\n"
+            "Power status: {result.power.name}\n"
+            "Door: {result.door.name}\n"
+            "Temperature: {result.temperature}\n"
+            "Schedule: {result.schedule}\n"
+            "Air refresh interval: {result.air_refresh_interval}\n"
+            "Child lock: {result.child_lock.name}\n"
+            "System status (raw): {result.system_status_raw}\n"
+            "Errors: {result.errors}",
+        )
+    )
+    def status(self) -> ViomiDishwasherStatus:
+        """Retrieve properties."""
+
+        properties = [
+            "child_lock",
+            "program",
+            "run_status",
+            "wash_status",
+            "wash_temp",
+            "power",
+            "left_time",
+            "wash_done_appointment",
+            "freshdry_interval",
+            "wash_process",
+        ]
+
+        import inspect
+
+        if "flatten_request" in inspect.getfullargspec(self.get_properties).args:
+            values = self.get_properties(properties, flatten_request=True)
+        else:
+            values = self.get_properties(properties, max_properties=1)
+
+        return ViomiDishwasherStatus(defaultdict(lambda: None, zip(properties, values)))
+
+    # FIXME: Change these to use the ViomiDishwasherStatus once we can query multiple properties at once (or cache?).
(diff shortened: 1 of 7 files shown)


TheDJVG:
I'm not sure how I can fix these two Windows test failures as I don't think it has anything to do with my changes. Feedback appreciated.

syssi:
The windows tests are failing sometimes and can be ignored here.

rytilahti (commented):
Great work on the PR! :1st_place_medal: I restarted the windows build and it went fine, it's flaky sometimes as syssi mentioned.

I did a brief initial review, to summarize:
* Prefer booleans instead of enums for boolean values.
* Add docstrings to public functions, examples for data payloads.
* Log unknown responses so users can report/debug it.
* Please also update the README.md.
* Update discovery mapping if the device is discoverable via mdns.

On the flattening of the requests, I haven't given a thought on that for now, but if the same issue affects other devices (maybe a buggy sdk version?) we should deal with it in `Device` class. Otherwise, overriding `send` could work as a stopgap solution. 

@syssi do you have any devices that only support `max_properties=1` to try out if flattening would allow querying multiple properties on those?

syssi:
> @syssi do you have any devices that only support `max_properties=1` to try out if flattening would allow querying multiple properties on those?

I would like to test the `flatten_request` feature on my `deerma.humidifier.jsq`. Could you provide a PR? I couldn't find the implementation.

rytilahti:
Ohh, it wasn't a PR but an issue: https://github.com/rytilahti/python-miio/issues/875 . Testing with a raw-command would work for initial testing :-)

TheDJVG:
Sorry all! Was a bit of a busy week here. Will try to push the changes to address the feedback this weekend.

miio/viomidishwasher.py
```diff
@@ -0,0 +1,370 @@
...
+        if self.data["run_status"] & (1 << 3):
+            errors.append(SystemStatus.InternalConnectionError)
+        if self.data["run_status"] & (1 << 2):
+            errors.append(SystemStatus.InsufficientWaterFlow)
+        if self.data["run_status"] & (1 << 5):
+            errors.append(SystemStatus.ThermistorError)
+        if self.data["run_status"] & (1 << 9):
+            errors.append(SystemStatus.InsufficientWaterSoftener)
+        if self.data["run_status"] & (1 << 11):
+            errors.append(SystemStatus.HeatingElementError)
```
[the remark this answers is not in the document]
TheDJVG: This is currently very unknown. I don't know all the exact status options emitted by the device. These are the ones that I was able to detect by randomly disconnecting cables on the inside and see what happens.

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+                    "program": 2,
+                    "run_status": 512,
+                    "wash_status": 0,
+                    "wash_temp": 86,
+                    "power": 0,
+                    "left_time": 0,
+                    "wash_done_appointment": 0,
+                    "freshdry_interval": 0,
+                    "wash_process": 0
+                }
```
rytilahti, line 98:
```suggestion
            {
                "child_lock": 0,
                "program": 2,
                "run_status": 512,
                "wash_status": 0,
                "wash_temp": 86,
                "power": 0,
                "left_time": 0,
                "wash_done_appointment": 0,
                "freshdry_interval": 0,
                "wash_process": 0
            }
```

rytilahti: `Example:` is already enough to indicate that the following data is an example of what to expect.

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+
+    @property
+    def child_lock(self) -> bool:
+        """Returns the child lock status of the device.
+
+        Returns:
+            bool: The return value. True for enabled, False for disabled.
+
+        Raises:
+            DeviceException: When the returned value by the device is not 1 or 0.
```
rytilahti, line 112:
```suggestion
```

No need for extra docstrings here, if the value is something unexpected an exception will be raised and that's enough to debug the problem. The docstring could be adapted to be something like `Return true if the child lock is active.`, making the `Returns` block unnecessary.

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+        except ValueError:
+            _LOGGER.warning("Program %r is Unknown.", program)
+            return Program.Unknown
+
+    @property
+    def door_open(self) -> bool:
+        """Returns the door status of the device.
+
+        Returns:
+            Bool: True if open False if closed.
```
rytilahti, line 139:
```suggestion
        """Returns True if the door is open.
```

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+            bool: The return value. True for enabled, False for disabled.
+
+        Raises:
+            DeviceException: When the returned value by the device is not 1 or 0.
+        """
+        value = self.data["child_lock"]
+        if value in [0, 1]:
+            return bool(value)
+
+        raise DeviceException(f"{value} is not a valid child lock status.")
```
rytilahti, line 118:
```suggestion
        try:
            return bool(value)
        except Exception as ex:
            raise DeviceException(f"{value} is not a valid child lock status.") from ex
```

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+
+    @property
+    def program(self) -> Program:
+        """Returns the current selected program of the device.
+
+        Returns:
+            Program: The current program or Program.Unknown if an Unknown program was detected.
```
rytilahti, line 125:
```suggestion
```

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+    @property
+    def system_status_raw(self) -> int:
+        """Returns the raw status number of the device.
+
+        This is in general used to detected:
+            - Errors in the system.
+            - If the door is open or not.
+
+        Returns:
+            int: Raw value.
```
rytilahti, line 153:
```suggestion
```

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+        return self.data["run_status"]
+
+    @property
+    def status(self) -> MachineStatus:
+        """Returns the machine status of the device.
+
+        This is in general used to determine what the device is doing.
+
+        Returns:
+            MachineStatus: The machine status Enum.
```
rytilahti, line 165:
```suggestion
```

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+
+    @property
+    def power(self) -> bool:
+        """Returns the power status of the device.
+
+        Returns:
+            bool: The return value. True for power on, False for power off.
+
+        Raises:
+            DeviceException: When the returned value by the device is not 1 or 0.
```
rytilahti, line 188: See my prior commentary on verbose docstrings / boolean coercion :-) If there are similar cases below, I won't comment on them separately.

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+        if value in [0, 1]:
+            return bool(value)
+
+        raise DeviceException(f"{value} is not a valid power status.")
+
+    @property
+    def time_left(self) -> timedelta:
+        """Returns the timedelta in seconds of time left of the current program.
+
+        Will always be 0 if no program is running.
```
rytilahti, line 201: Maybe check for this case in the code and return None, and convert the type hint to `Optional[timedelta]`?

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+        if isinstance(value, int):
+            return timedelta(seconds=value)
+
+        raise DeviceException(f"{value} is not a valid integer for time_left.")
+
+    @property
+    def schedule(self) -> Optional[datetime]:
+        """Returns a datetime when the scheduled program should be finished..
+
+        Will always be 0 if nothing is scheduled.
```
rytilahti, line 219: See my comment on inapplicable values above, I think it's better to return `None` when there's no schedule.

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+        """
+        value = self.data["left_time"]
+        if isinstance(value, int):
+            return timedelta(seconds=value)
+
+        raise DeviceException(f"{value} is not a valid integer for time_left.")
+
+    @property
+    def schedule(self) -> Optional[datetime]:
+        """Returns a datetime when the scheduled program should be finished..
```
rytilahti, line 217:
```suggestion
        """Returns a datetime when the scheduled program should be finished.
```

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+
+        Returns:
+            datetime: A datetime when the selected program should be finished.
+
+        Raises:
+            DeviceException: When the returned value by the device is not a valid integer.
+        """
+
+        value = self.data["wash_done_appointment"]
+        if isinstance(value, int):
```
rytilahti, line 229: Prefer try-catch blocks over `isinstance`, that will make also writing tests much easier (isinstance does not play well with mocks).

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+
+        raise DeviceException(
+            f"{value} is not a valid integer for wash_done_appointment."
+        )
+
+    @property
+    def air_refresh_interval(self) -> int:
+        """Returns an integer on how often the air in the device should be refreshed.
+
+        Returns 0 if disabled.
```
rytilahti, line 240: Prefer `None` here, too.

miio/viomidishwasher.py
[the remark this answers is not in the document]
rytilahti: Fair enough, I think it'd still make sense to go these through bitwise (and report on unknown bits).

miio/viomidishwasher.py
```diff
@@ -0,0 +1,529 @@
...
+
+MODELS_SUPPORTED = [MODEL_DISWAHSER_M02]
+
+
+class MachineStatus(enum.IntEnum):
+    Off = 0
+    On = 1
+    Running = 2
+    Paused = 3
+    Scheduled = 5
```
Mr-Groch, line 25: Completed = 4

rytilahti: Thanks @Mr-Groch :+1:

rytilahti (approved):
Thanks for the PR @TheDJVG and sorry for the delay, hope this can still be useful for someone even when you don't have the device anymore :+1:

PR #868, opened 2020-11-25
Initial support for HUIZUO PISCES For Bedroom

darckly:
I have such a lamp and wrote the basic library to support it. Please review.

Issue #627, opened 2020-02-13
[meta] list of miot-enabled devices


rytilahti:
This issue is to collect devices that use miot protocol for communication, which will help with future development efforts.

If you know about a device that is not listed here, please add a comment and it will be added to this list.

For information about obtaining the specifications for a given device, see https://github.com/rytilahti/python-miio/issues/543#issuecomment-544643792.

devtools directory (https://github.com/rytilahti/python-miio/tree/master/devtools) contains `miottemplate` tool which can be useful for working with the specification files.

## Currently supported miot devices

Mi Air Purifier 3 (zhimi.airpurifier.ma4) (#585)
Mi Air Purifier 3/3H (zhimi.airpurifier.mb3) (#585)

## Currently known miot devices 
One meter intelligent curtain (090615.curtain.1mcu01)
PTX Rolling curtain (090615.curtain.jldj03)
Intelligent 86 socket(WIFI) (090615.plug.plus01)
Intelligent Mobile Plug (WIFI) (090615.plug.plus02)
Piak intelligent one switch (090615.switch.piak01)
Piak intelligent two switch (090615.switch.piak02)
Piak intelligent three switch (090615.switch.piak03)
Ai-Thinker Smart Socket (180.plug.plug0001)
Smart Curtain Motor (666.curtain.gt01)
Airmate mobile floor heating (air.heater.wda14)
Nnuodu Intelligent Living Room Lamp (app.light.wynd1)
Feipujia Intelligent lamps and lanterns (asdds.light.wyfpj1)
AUTORAIL Curtain Motor (babai.curtain.at5810)
MTX Smart Curtain (babai.curtain.mtx850)
Babuy Smart Curtain (babai.curtain.bb82mj)
NanHoo Smart Curtain (babai.curtain.nh5810)
WiFi Wall Switch A1 (babai.switch.bb101s)
WiFi Wall Switch A3 (babai.switch.bb103s)
Smart Curtain Motor (babai.curtain.190812)
WiFi Wall Switch A1 (babai.switch.bb101s)
WiFi Wall Switch A2 (babai.switch.bb102s)
WiFi Wall Switch A3 (babai.switch.bb103s)
LOMEDIQI Intelligent automatic air purifier (bjdm.airfresh.test01)
Qingfly smart portable clothes dryer (coc.dryer.fdpsm)
Heatcold Heat Pump Thermostat (cubee.airrtc.th125t)
[...]

PR #881, opened 2020-12-03
Add support for all Huizuo Lamps (w/ fans, heaters, and scenes)

darckly:
Inspired by previous work (PR #868), I checked specifications for all HUIZUO lamps, mentioned in the #627 [meta] list of miot-enabled devices. In general, almost all of them have a similar schema for light management, and some models have additionally a fan, heater, or predefined scenes.
Code for additional features was written based on official specifications.


miio/huizuo.py
```diff
@@ -34,44 +120,102 @@ class HuizuoException(DeviceException):
...
-
         self.data = data
 
     @property
     def is_on(self) -> bool:
         """Return True if device is on."""
-        return self.data["power"]
+        if "power" in self.data:
+            return self.data["power"]
+        return None
```
syssi, line 131: Please don't return `None` if all devices supports this attribute.

miio/huizuo.py
```diff
@@ -34,44 +120,102 @@ class HuizuoException(DeviceException):
...
 
     @property
     def brightness(self) -> int:
         """Return current brightness."""
-        return self.data["brightness"]
+        if "brightness" in self.data:
+            return self.data["brightness"]
+        return None
```
syssi, line 137: If the attribute is optional please use `Optional[int]` as type of the return value.

miio/huizuo.py
```diff
@@ -87,6 +231,16 @@ def __init__(
         lazy_discover: bool = True,
         model: str = MODEL_HUIZUO_PIS123,
     ) -> None:
+
+        if model in MODELS_WITH_FAN_WY:
```
syssi, line 235: You could implement a new device class per device group. Each device class will pop-up as new parameter/device at the miiocli.

miio/huizuo.py
```diff
@@ -147,7 +312,226 @@ def set_brightness(self, level):
...
+        default_output=format_output("Disable fan reverse"),
+    )
+    def fan_reverse_off(self):
+        """Disable fan reverse (only for models which support this fan option)"""
+        if self.model in MODELS_WITH_FAN_WY:
+            return self.set_property("fan_motor_reverse", False)
+        else:
+            raise HuizuoException("Your device doesn't support a fan management")
+
+    # The next section contains the heater management commands
```
syssi, line 407: The heater commands should be implemented in the new device class for heaters f.e.

miio/huizuo.py
darckly: Removed for all basic lamp methods

miio/huizuo.py
darckly: Added

miio/huizuo.py
darckly: Split with classes

miio/huizuo.py
darckly: Split with classes

darckly:
Guys, RP is ready for review, if you have a time

miio/huizuo.py
```diff
@@ -147,7 +296,288 @@ def set_brightness(self, level):
...
+            }
+        )
+
+
+class HuizuoLampScene(Huizuo):
+    """Support for Huizuo Lamps with additional scene commands"""
+
+    # The next section contains the scene management commands
+    # Right now I have no devices with the scenes for live testing, so the following section
+    # generated based on device specitifations
```
rytilahti, line 493: Please include these to the class docstring, that way they will also be visible in the generated documentation.

miio/huizuo.py
```diff
@@ -147,7 +296,288 @@ def set_brightness(self, level):
...
+        """Switch the on/off (only for models with scenes support)."""
+        if self.model in MODELS_WITH_SCENES:
+            return self.set_property("on_off", 0)
+        else:
+            raise HuizuoException("Your device doesn't support scenes")
+
+    @command(
+        default_output=format_output("Add the brightness"),
+    )
+    def brightness_add(self):
```
rytilahti, line 508:
```suggestion
    def brightness_increase(self):
```
Maybe? And `brightness_decrease` for the reverse?

miio/huizuo.py
```diff
@@ -147,7 +296,288 @@ def set_brightness(self, level):
...
+        """Switch between the brightnesses (only for models with scenes support)."""
+        if self.model in MODELS_WITH_SCENES:
+            return self.set_property("brightness_switch", 0)
+        else:
+            raise HuizuoException("Your device doesn't support scenes")
+
+    @command(
+        default_output=format_output("Add the color temperature"),
+    )
+    def colortemp_add(self):
```
rytilahti, line 538: See my comment wrt add/increase above. The method/property naming does not necessarily need to follow the "raw" property names where better/more comprehensible names are applicable.

miio/huizuo.py
```diff
@@ -34,21 +120,6 @@ class HuizuoException(DeviceException):
 
 class HuizuoStatus:
     def __init__(self, data: Dict[str, Any]) -> None:
-        """
```
rytilahti, line 37: Could you please re-add an example response (with most possible properties, if possible)?

rytilahti: Also, add a note that if the device does not have the property, `None` will be returned (all those optionals) :-)

miio/huizuo.py
```diff
@@ -66,17 +137,84 @@ def color_temp(self) -> int:
...
+        if self.fan_mode is not None:
+            parameters.append("fan_mode=" + self.fan_mode)
+        if self.is_fan_reverse is not None:
+            parameters.append("fan_motor_reverse=" + self.is_fan_reverse)
+        if self.is_heater_on is not None:
+            parameters.append("heater_on=" + self.is_heater_on)
+        if self.heat_level is not None:
+            parameters.append("heat_level=" + self.heat_level)
+        if self.heater_fault_code is not None:
+            parameters.append("heater_fault_code=" + self.heater_fault_code)
```
rytilahti, line 210: Maybe creating a list of available properties, and using `getattr(self, 'property_name')` in a loop to avoid repetition and make the code cleaner?

So something like this, although there is probably a nicer way to get a list of all properties (to avoid the string list):
```
optional_props = ['is_fan_on', ..., 'fan_speed_level']
for prop in optional_props:
    val = getattr(self, prop)
    if val is not None:
        parameters.append(f"{prop}={val}")
```

rytilahti (changes requested):
Nice work, that's plenty of support for new devices in a single PR! :1st_place_medal: See some commentary inline.

miio/huizuo.py
darckly: Updated

miio/huizuo.py
darckly: Changed to *_ increase/*_ decrease everywhere in code

miio/huizuo.py
darckly: Done

miio/huizuo.py
darckly: Returned the example response, but only for one type of lamp (which I have).
Also, additional note was added

miio/huizuo.py
darckly: Changed as you suggested.
I added all available properties to the list of properties to have one approach to describe the properties

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
+        """Switch between the color temperatures (only for models with scenes support)."""
+        if self.model in MODELS_WITH_SCENES:
+            return self.set_property("colortemp_switch", 0)
+        else:
+            raise HuizuoException("Your device doesn't support scenes")
+
+    @command(
+        default_output=format_output("Switch on or increase brightness"),
+    )
+    def on_or_brightness(self):
```
rytilahti, line 583:
```suggestion
    def on_or_increase_brightness(self):
```

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
+        """Switch on or increase brightness (only for models with scenes support)."""
+        if self.model in MODELS_WITH_SCENES:
+            return self.set_property("on_or_brightness", 0)
+        else:
+            raise HuizuoException("Your device doesn't support scenes")
+
+    @command(
+        default_output=format_output("Switch on or increase color temperature"),
+    )
+    def on_or_colortemp(self):
```
rytilahti, line 593:
```suggestion
    def on_or_increase_colortemp(self):
```

rytilahti (approved):
Looks good to me, thanks! :tada:

@syssi, what do you think?

miio/huizuo.py
darckly: Updated

miio/huizuo.py
darckly: Updated

darckly:
I accidentally saw that you requested two small changes.
Updated.

miio/huizuo.py
```diff
@@ -109,8 +109,8 @@
     "colortemp_add": {"siid": 3, "piid": 5},
     "colortemp_dec": {"siid": 3, "piid": 6},
     "colortemp_switch": {"siid": 3, "piid": 7},
-    "on_or_brightness": {"siid": 3, "piid": 8},
-    "on_or_colortemp": {"siid": 3, "piid": 9},
+    "on_or_increasbrightness": {"siid": 3, "piid": 8},
```
rytilahti, line 112:
```suggestion
    "on_or_increase_brightness": {"siid": 3, "piid": 8},
```
Typofix.

darckly:
Good catch!
Also pulled the changes from master to resolve the conflicts

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
 
         return self.set_property("color_temp", color_temp)
+
+
+class HuizuoLampFan(Huizuo):
+    """Support for Huizuo Lamps with fan"""
+
+    # The next section contains the fan management commands
+    # Right now I have no devices with the fan for live testing, so the following section
+    # generated based on device specitifations
```
rytilahti, line 333: Could you move this also to the docstring, then it will be visible in the apidocs (and IDEs).

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
+
+    @command(
+        default_output=format_output("Fan powering on"),
+    )
+    def fan_on(self):
+        """Power fan on (only for models with fan)."""
+        if self.model in MODELS_WITH_FAN_WY or self.model in MODELS_WITH_FAN_WY2:
+            return self.set_property("fan_power", True)
+        else:
+            raise HuizuoException("Your device doesn't support a fan management")
```
rytilahti, line 343: No need for elses for exception handling, as there's a return for supported devices already in place.

This will make the code a bit cleaner to read (especially on some other cases below).

```suggestion

        raise HuizuoException("Your device doesn't support a fan management")
```

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
+
+    @command(
+        click.argument("fan_level", type=int),
+        default_output=format_output("Setting fan speed level to {fan_level}"),
+    )
+    def set_fan_level(self, fan_level):
+        """Set fan speed level (only for models with fan)"""
+        if self.model in MODELS_WITH_FAN_WY or self.model in MODELS_WITH_FAN_WY2:
+            if fan_level < 0 or fan_level > 100:
+                raise HuizuoException("Invalid fan speed level: %s" % fan_level)
```
rytilahti, line 363: I'd move this to be a top-level check, before checking if the model supports the fan. Less indents make the code easier to read!

miio/huizuo.py
rytilahti, line 343:
```suggestion

        raise HuizuoException("Your device doesn't support a fan management")
```

Here too, check and modify also the remaining ones below which I won't comment on separately.

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
+            "------------Fan parameters -------------\n"
+            "Fan power:  {result.is_fan_on}\n"
+            "Fan level: {result.fan_speed_level}\n"
+            "Fan mode: {result.fan_mode}\n"
+            "Fan reverse: {result.is_fan_reverse}\n"
+            "\n",
+        ),
+    )
+    def status(self) -> HuizuoStatus:
+        """Retrieve properties."""
```
rytilahti, line 406:
```suggestion
        """Return status of the device."""
```
Looks nicer in `--help` and apidocs that way.

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
+
+    @command(
+        click.argument("heat_level", type=int),
+        default_output=format_output("Setting heat level to {heat_level}"),
+    )
+    def set_heat_level(self, heat_level):
+        """Set heat level (only for models with heater)"""
+        if self.model in MODELS_WITH_HEATER:
+            if heat_level not in [1, 2, 3]:
+                raise HuizuoException("Invalid heat level: %s" % heat_level)
```
rytilahti, line 472: See my previous comment on indents & failing early (so move this before the model check).

miio/huizuo.py
rytilahti: Also, this probably should be in `Huizuo` class to avoid code duplication? See below.

miio/huizuo.py
```diff
@@ -147,7 +311,288 @@ def set_brightness(self, level):
...
+    def status(self) -> HuizuoStatus:
+        """Retrieve properties."""
+
+        return HuizuoStatus(
+            {
+                prop["did"]: prop["value"] if prop["code"] == 0 else None
+                for prop in self.get_properties_for_mapping()
+            }
+        )
+
```
rytilahti, line 501: This is a duplicate from the base class (`Huizuo`)?

miio/huizuo.py
darckly: Moved

miio/huizuo.py
darckly: Simplified

darckly: Moved

miio/huizuo.py
darckly: We have a different output format (list of available fields) in different classes. It depends on the available functions of the lamp. That's why I'm re-defining the 'HuizuoStatus' method in every class.
Or you have another suggestion (maybe I do not fully understand your idea)?

miio/huizuo.py
darckly: Moved

miio/huizuo.py
darckly: See my comment above

miio/huizuo.py
rytilahti: Hmm, but the unavailable properties would then just be reported as `None` which would be fine (in my opinion). I'm not completely against having separate status outputs, but I'd rather have a single status response for maintainability.

darckly: I tried this already in the version, which I tested internally with my basic lamps, and I made the method overloading for the next reasons:
1. I don't know the full model line of Huizuo lamps, I only added what I found in the Issue description. If they have/will have more devices with different new functions (like RGB support, Ambilight, etc.), this single list will be quite long, mostly filled with "none" for basic devices. Not very comfortable (IMO) for reading
2. Device can have the functions from different sections in the list, which may be separated (e.g. basic in the beginning and RGB at the end of the list).
Method overloading allows simplifying the reading of the properties of the specific device type.

But, if you have already a "single list" approach for other devices with similar classes organizations, indeed, better bring it to the single style.
Just inform me about the decision ;)

rytilahti: Okay, let's have it like it is right now, we can always revisit that in the future if needed :-) Thanks again for the great PR! :tada:

Issue #882, opened 2020-12-04
Add support for Mi Smart Standing Fan 1C (dmaker.fan.p8)


ehn:
Before submitting a new request, use the search to see if there is an existing issue for the device.

**Device information:**

  - Name(s) of the device: Mi Smart Standing Fan 1C
  - Link: https://en.techreviewer.de/xiaomi-standventilator-test/ (can't find an official product page)

Use `miiocli device --ip <ip address> --token <token>`.

  - Model: dmaker.fan.p8
  - Hardware version: esp8266
  - Firmware version: 2.0.8 (out of the box), 2.1.3 (after upgrade)

**Additional context**
Works as an miot-enabled device (see #627). For example:

```console
% miiocli fanmiot --ip <ip address> --token <token> off
Powering off
[{'did': 'power', 'siid': 2, 'piid': 1, 'code': 0}]
% miiocli fanmiot --ip <ip address> --token <token> on
Powering on
[{'did': 'power', 'siid': 2, 'piid': 1, 'code': 0}]
% miiocli fanmiot --ip <ip address> --token <token> status
Power: off
Operation mode: OperationMode.Normal
Speed: 0
Oscillate: None
Angle: None
LED: 0
Buzzer: None
Child lock: False
Power-off time: None
```

However, the ``Speed`` field of the ``status`` output seems to be the minutes remaining until powering off when set.

The [API documentation](http://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:fan:0000A005:dmaker-p8:1), if you can call it that, looks similar but not identical to that of [dmaker.fan.p9](http://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:fan:0000A005:dmaker-p9:1) and [dmaker.fan.p10](http://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:fan:0000A005:dmaker-p10:1):

```json
{
    "type": "urn:miot-spec-v2:device:fan:0000A005:dmaker-p8:1",
    "description": "Fan",
    "services": [
        {
            "iid": 1,
            "type": "urn:miot-spec-v2:service:device-information:00007801:dmaker-p8:1",
            "description": "Device Information",
            "properties": [
                {
                    "iid": 1,
                    "type": "urn:miot-spec-v2:property:manufacturer:00000001:dmaker-p8:1",
                    "description": "Device Manufacturer",
                    "format": "string",
                    "access": [
                        "read"
[...]

PR #883, opened 2020-12-04
Add dmaker.fan.p8 support (Closes: #882)

syssi:
TODO:

- [ ] Handle piid 11 & piid 12
- [ ] Add tests

### DIFF
--- a/miio/fan_miot.py
+++ b/miio/fan_miot.py
@@ -130,12 +130,12 @@ class FanStatusP8(DeviceStatus):
         return self.data["power"]
 
     @property
-    def mode(self) -> OperationMode:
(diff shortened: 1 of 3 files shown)


miio/fan_miot.py
```diff
@@ -63,6 +73,88 @@ class OperationModeMiot(enum.Enum):
...
+                self.speed_level,
+                self.oscillate,
+                self.child_lock,
+                self.delay_off_countdown,
+            )
+        )
+        return s
+
+
 class FanStatusMiot:
```
rytilahti, line 158: This should also be derived from `DeviceStatus`?

syssi:
It looks like P8 and 1C has the same specs.

PR #884, opened 2020-12-05
Retry and timeout can be change by setting a class attribute

titilambert:
No breaking change
With this, each class can have its own retry and timeout setting, ie:
```
class Vacuum(Device):
   retry =  5
   timeout = 10

   ...
```

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -117,6 +117,8 @@ class Device(metaclass=DeviceGroupMeta):
     ) -> None:
         self.ip = ip
         self.token = token
+        if hasattr(self, "timeout"):
+            timeout = self.timeout
         self._protocol = MiIOProtocol(
             ip, token, start_id, debug, lazy_discover, timeout
         )
@@ -143,6 +145,8 @@ class Device(metaclass=DeviceGroupMeta):
         :param int retry_count: How many times to retry on error
         :param dict extra_parameters: Extra top-level parameters
         """
+        if hasattr(self, "retry_count"):
+            retry_count = self.retry_count
         return self._protocol.send(
             command, parameters, retry_count, extra_parameters=extra_parameters
         )
--- a/miio/device.py
+++ b/miio/device.py
@@ -147,7 +147,7 @@ class Device(metaclass=DeviceGroupMeta):
         :param int retry_count: How many times to retry on error
         :param dict extra_parameters: Extra top-level parameters
         """
-        retry_count = retry_count if retry_count is not None else self._retry_count
+        retry_count = retry_count if retry_count is not None else self.retry_count
         return self._protocol.send(
             command, parameters, retry_count, extra_parameters=extra_parameters
         )
--- a/miio/device.py
+++ b/miio/device.py
@@ -106,6 +106,9 @@ class Device(metaclass=DeviceGroupMeta):
     This class should not be initialized directly but a device-specific class inheriting
     it should be used instead of it."""
 
+    retry_count = 3
+    timeout = 5
+
     def __init__(
         self,
         ip: str = None,
@@ -113,12 +116,11 @@ class Device(metaclass=DeviceGroupMeta):
         start_id: int = 0,
         debug: int = 0,
         lazy_discover: bool = True,
-        timeout: int = 5,
+        timeout: int = None,
     ) -> None:
         self.ip = ip
         self.token = token
-        if hasattr(self, "timeout"):
-            timeout = self.timeout
+        timeout = timeout if timeout is not None else self.timeout
         self._protocol = MiIOProtocol(
             ip, token, start_id, debug, lazy_discover, timeout
         )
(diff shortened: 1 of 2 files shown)


titilambert:
@rytilahti 
Here the change made for retry and timeout in viomi vacuum PR (https://github.com/rytilahti/python-miio/pull/808)

miio/device.py
```diff
@@ -143,6 +145,8 @@ def send(
         :param int retry_count: How many times to retry on error
         :param dict extra_parameters: Extra top-level parameters
         """
+        if hasattr(self, "retry_count"):
+            retry_count = self.retry_count
```
rytilahti, line 149: This would override the user-given parameter which is a no-no.

A workable solution for now would be to move these settings to be class variables for `Device`:
```
class Device:
    timeout = 5
    retry_count = 3
```
For individual methods (`__init__()` and this), the corresponding parameters should default to `None` instead of their current values. If those parameters are not filled by the caller, the fallback is to use the class-defined values.

That way this piece of code would became and the current behavior for downstream users stays intact:
```
retry_count = retry_count if retry_count is not None else self.retry_count
```

Could you also please add some tests for this?

rytilahti:
Hi @titilambert, first of all, your efforts to provide tests is really appreciated! :-) I saw some commits on this PR in notifications, and wanted to mention that you can also run tests locally using `pytest miio` in the main source directory, or by calling `tox` to run all checks what the CI system does.

I noticed you are trying to check out if a parameter has been given (rightly so!), so I just wanted to mention that you may want to check if `assert_called_with()` works instead of manually call args list. I'm not that familiar with pytest/mock myself either, so I cannot definitely say if that works or not, but the name implies that it'd be useful in this case.

miio/device.py
titilambert: > Hi @titilambert, first of all, your efforts to provide tests is really appreciated! :-) I saw some commits on this PR in notifications, and wanted to mention that you can also run tests locally using `pytest miio` in the main source directory, or by calling `tox` to run all checks what the CI system does.
> 
> I noticed you are trying to check out if a parameter has been given (rightly so!), so I just wanted to mention that you may want to check if `assert_called_with()` works instead of manually call args list. I'm not that familiar with pytest/mock myself either, so I cannot definitely say if that works or not, but the name implies that it'd be useful in this case.

@rytilahti  Thanks! I just found that function :D I just pushed a working commit, and I just squashed all my tests commits ...

It's now ready to review

titilambert:
> Hi @titilambert, first of all, your efforts to provide tests is really appreciated! :-) I saw some commits on this PR in notifications, and wanted to mention that you can also run tests locally using `pytest miio` in the main source directory, or by calling `tox` to run all checks what the CI system does.> Hi @titilambert, first of all, your efforts to provide tests is really appreciated! :-) I saw some commits on this PR in notifications, and wanted to mention that you can also run tests locally using `pytest miio` in the main source directory, or by calling `tox` to run all checks what the CI system does.
> 
> I noticed you are trying to check out if a parameter has been given (rightly so!), so I just wanted to mention that you may want to check if `assert_called_with()` works instead of manually call args list. I'm not that familiar with pytest/mock myself either, so I cannot definitely say if that works or not, but the name implies that it'd be useful in this case.

@rytilahti  Thanks, I just found it (like 5 minutes before your comment :D ) I just squashed and pushed.
So I guess it's now ready for review :)

Thanks again

miio/tests/test_device.py
```diff
@@ -17,6 +17,17 @@ def test_get_properties_splitting(mocker, max_properties):
     if max_properties is None:
         max_properties = len(properties)
     assert send.call_count == math.ceil(len(properties) / max_properties)
+    assert 5 == d._protocol._timeout
```
rytilahti, line 20: Please create a separate test cases for the default values, maybe simply `test_timeout_and_retry` to check both of these that use the regular timeouts.

miio/tests/test_device.py
```diff
@@ -17,6 +17,17 @@ def test_get_properties_splitting(mocker, max_properties):
...
+
+
+def test_timeout_retry(mocker):
+    send = mocker.patch("miio.miioprotocol.MiIOProtocol.send")
+    d = Device("127.0.0.1", "68ffffffffffffffffffffffffffffff", timeout=4)
+    assert 4 == d._protocol._timeout
+    d.send("fake_command", [], 1)
+    send.assert_called_with("fake_command", [], 1, extra_parameters=None)
```
rytilahti, line 28: I'm not sure if `assert_called_with` also works with kwargs, but if it does, prefer the kwarg format for calls & checks.

It would also be great to have separate tests for setting the timeout and the retry counts in a subclass, you could add a `Device`-derived test class, with some custom timeout & retry count values, to this file and use that for these tests.

miio/tests/test_device.py
titilambert: done

miio/tests/test_device.py
titilambert: it does, but `send` is not called with kwargs only (see: https://github.com/rytilahti/python-miio/blob/master/miio/device.py#L147)

titilambert:
@rytilahti I just added a new test and made some adjustments

rytilahti (approved):
LGTM, thanks! :+1:

PR #887, opened 2020-12-06
Add support for Yeelight Dual Control Module (yeelink.switch.sw1)

IhorSyerkov:
Added support Yeelight Dual Control Module (yeelink.switch.sw1) from #627 list.
This switch doesn’t have LAN-mode and cannot be controlled via [python-yeelight](https://gitlab.com/stavros/python-yeelight)

### DIFF
--- a/miio/tests/dummies.py
+++ b/miio/tests/dummies.py
@@ -66,6 +66,16 @@ class DummyMiotDevice(DummyDevice):
     def get_properties_for_mapping(self):
         return self.state
 
+    def get_properties(self, properties):
+        """Return values only for listed properties"""
+        keys = [p["did"] for p in properties]
+        props = []
+        for prop in self.state:
+            if prop["did"] in keys:
+                props.append(prop)
+
+        return props
+
     def set_property(self, property_key: str, value):
         for prop in self.state:
             if prop["did"] == property_key:
--- a/miio/yeelight_dual_switch.py
+++ b/miio/yeelight_dual_switch.py
@@ -1,0 +1,263 @@
+import enum
+from typing import Any, Dict
+
+import click
+
+from .click_common import EnumType, command, format_output
+from .exceptions import DeviceException
+from .miot_device import MiotDevice
+
+
+class YeelightDualControlModuleException(DeviceException):
+    pass
+
+
+class Switch(enum.Enum):
+    First = 0
+    Second = 1
+
+
+_MAPPING = {
+    # http://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:switch:0000A003:yeelink-sw1:1:0000C809
+    # First Switch (siid=2)
+    "switch_1_state": {"siid": 2, "piid": 1},  # bool
+    "switch_1_default_state": {"siid": 2, "piid": 2},  # 0 - Off, 1 - On
+    "switch_1_off_delay": {"siid": 2, "piid": 3},  # -1 - Off, [1, 43200] - delay in sec
+    # Second Switch (siid=3)
+    "switch_2_state": {"siid": 3, "piid": 1},  # bool
+    "switch_2_default_state": {"siid": 3, "piid": 2},  # 0 - Off, 1 - On
+    "switch_2_off_delay": {"siid": 3, "piid": 3},  # -1 - Off, [1, 43200] - delay in sec
+    # Extensions (siid=4)
+    "interlock": {"siid": 4, "piid": 1},  # bool
+    "flex_mode": {"siid": 4, "piid": 2},  # 0 - Off, 1 - On
+    "rc_list": {"siid": 4, "piid": 3},  # string
+    "rc_list_for_del": {"siid": 4, "piid": 4},  # string
+    "toggle": {"siid": 4, "piid": 5},  # 0 - First switch, 1 - Second switch
+}
+
+
+class DualControlModuleStatus:
+    def __init__(self, data: Dict[str, Any]) -> None:
+        """
+        Response of Yeelight Dual Control Module
+        {
+            'id': 1,
+            'result': [
+                {'did': 'switch_1_state', 'siid': 2, 'piid': 1, 'code': 0, 'value': False},
+                {'did': 'switch_1_default_state', 'siid': 2, 'piid': 2, 'code': 0, 'value': True},
+                {'did': 'switch_1_off_delay', 'siid': 2, 'piid': 3, 'code': 0, 'value': 300},
+                {'did': 'switch_2_state', 'siid': 3, 'piid': 1, 'code': 0, 'value': False},
+                {'did': 'switch_2_default_state', 'siid': 3, 'piid': 2, 'code': 0, 'value': False},
+                {'did': 'switch_2_off_delay', 'siid': 3, 'piid': 3, 'code': 0, 'value': 0},
+                {'did': 'interlock', 'siid': 4, 'piid': 1, 'code': 0, 'value': False},
+                {'did': 'flex_mode', 'siid': 4, 'piid': 2, 'code': 0, 'value': True},
+                {'did': 'rc_list', 'siid': 4, 'piid': 2, 'code': 0, 'value': '[{"mac":"9db0eb4124f8","evtid":4097,"pid":339,"beaconkey":"3691bc0679eef9596bb63abf"}]'},
+            ]
+        }
+        """
+        self.data = data
+
+    @property
+    def switch_1_state(self) -> bool:
+        """First switch state"""
+        return bool(self.data["switch_1_state"])
+
+    @property
+    def switch_1_default_state(self) -> bool:
+        """First switch default state"""
+        return bool(self.data["switch_1_default_state"])
+
+    @property
+    def switch_1_off_delay(self) -> int:
+        """First switch off delay"""
+        return self.data["switch_1_off_delay"]
+
+    @property
+    def switch_2_state(self) -> bool:
+        """Second switch state"""
+        return bool(self.data["switch_2_state"])
+
+    @property
+    def switch_2_default_state(self) -> bool:
+        """Second switch default state"""
+        return bool(self.data["switch_2_default_state"])
+
+    @property
+    def switch_2_off_delay(self) -> int:
+        """Second switch off delay"""
+        return self.data["switch_2_off_delay"]
+
+    @property
+    def interlock(self) -> bool:
+        """Interlock"""
+        return bool(self.data["interlock"])
+
+    @property
+    def flex_mode(self) -> int:
+        """Flex mode"""
+        return self.data["flex_mode"]
+
+    @property
+    def rc_list(self) -> str:
+        """List of paired remote controls"""
+        return self.data["rc_list"]
+
+    def __repr__(self) -> str:
+        s = (
+            "<YeelightDualControlModuleStatus "
+            "switch_1_state=%s, "
+            "switch_1_default_state=%s, "
+            "switch_1_off_delay=%s, "
+            "switch_2_state=%s, "
+            "switch_2_default_state=%s, "
+            "switch_2_off_delay=%s, "
+            "interlock=%s, "
+            "flex_mode=%s, "
+            "rc_list=%s>"
+            % (
+                self.switch_1_state,
+                self.switch_1_default_state,
+                self.switch_1_off_delay,
+                self.switch_2_state,
+                self.switch_2_default_state,
+                self.switch_2_off_delay,
+                self.interlock,
+                self.flex_mode,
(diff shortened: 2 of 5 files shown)


miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,233 @@
...
+    "flex_mode": {"siid": 4, "piid": 2},  # 0 - Off, 1 - On
+    "rc_list": {"siid": 4, "piid": 3},  # string
+    "rc_list_for_del": {"siid": 4, "piid": 4},  # string
+    "toggle": {"siid": 4, "piid": 5},  # 0 - channel0, 1 - channel1
+}
+
+
+class DualControlModuleStatus:
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
```
rytilahti, line 41: Please add a docstring showing an example response from the device, see other miot devices for examples.

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,233 @@
...
+        values = self.get_properties(properties, max_properties=1)
+        return DualControlModuleStatus(
+            dict(map(lambda v: (v["did"], v["value"]), values))
+        )
+
+    @command(
+        click.argument("switch_1_status", type=bool),
+        default_output=format_output("Set First Switch Status to: {switch_1_status}"),
+    )
+    def set_switch_1_status(self, switch_1_status: bool):
```
rytilahti, line 159: I'm wondering if it'd be better to follow the naming scheme of other switch devices (i.e., use `on()` and `off()` for state changes), and add a parameter describing on which switch to actuate?

rytilahti: Same applies to other methods too, not hardcoding these would allow extension in case a new switch comes up with three switches.

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,233 @@
...
+    def set_flex_mode(self, flex_mode: bool):
+        """Set Flex Mode."""
+        return self.set_property("flex_mode", (0, 1)[flex_mode])
+
+    @command(
+        click.argument("rc_list_for_del", type=str),
+        default_output=format_output("Delete remote control: {rc_list_for_del}"),
+    )
+    def delete_rc(self, rc_list_for_del: str):
+        """Delete Remote Control"""
```
rytilahti, line 224: Could you add an example on how to use this, especially concerning on what/how the input value should look like?

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,233 @@
...
+from .miot_device import MiotDevice
+
+
+class YeelightDualControlModuleException(DeviceException):
+    pass
+
+
+class Mode(enum.Enum):
+    Off = 0
+    On = 1
```
rytilahti, line 17: When the mode is clearly a boolean, simply use a boolean.

rytilahti (changes requested):
Thanks for the PR! I added some preliminary comments inline. Could you also add some tests?

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,233 @@
...
+    def set_switch_2_default_state(self, switch_2_default_state: bool):
+        """Set Second Switch Default State."""
+        return self.set_property("switch_1_default_state", switch_2_default_state.value)
+
+    @command(
+        click.argument("switch_2_delay", type=int),
+        default_output=format_output("Set Second switch delay to {switch_2_delay}%"),
+    )
+    def set_switch_2_delay(self, delay: int):
+        """Set Second switch delay, should be between -1 to 43200"""
```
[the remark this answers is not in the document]
IhorSyerkov: seconds

miio/yeelight_dual_switch.py
IhorSyerkov: not sure (. I don't have remote control for testing. Will try to found some info about it

miio/yeelight_dual_switch.py
IhorSyerkov: good idea

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,233 @@
...
+            raise YeelightDualControlModuleException("Invalid switch delay: %s" % delay)
+        return self.set_property("switch_2_delay", delay)
+
+    @command(
+        click.argument("flex_mode", type=bool),
+        default_output=format_output("Set Flex Mode to: {flex_mode}"),
+    )
+    def set_flex_mode(self, flex_mode: bool):
+        """Set Flex Mode."""
+        return self.set_property("flex_mode", (0, 1)[flex_mode])
```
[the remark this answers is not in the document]
IhorSyerkov: thank for suggesting. I'm new in python )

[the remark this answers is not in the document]
rytilahti: Happy to help! :-)

miio/yeelight_dual_switch.py
IhorSyerkov: What about status? Leave it as it is?

rytilahti: Just an idea :-) If done so, the class should expose the number of switches as a property to allow downstream users to figure out how many switches there exist.

miio/yeelight_dual_switch.py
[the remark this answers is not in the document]
rytilahti: What does the delay do? If it's action delay, the docstring could be more clear, maybe something like:
```Set the delay switch action takes to fire (in seconds).```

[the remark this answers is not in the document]
Comment (IhorSyerkov) [reply miio/yeelight_dual_switch.py:206]: If switch is on it sets timeout to off. After turning off this value resets. I'll update docstring

File: miio/yeelight_dual_switch.py (hunk 2 above)
Comment (IhorSyerkov) [reply miio/yeelight_dual_switch.py:159]: What do you suggest to use as parameter for choosing switch, enum or simply int
```python
class Switch(enum.Enum):
    First = 0
    Second = 1

def on(self, switch: Switch):

def on(self, switch: int):
    
```

miio/yeelight_dual_switch.py
IhorSyerkov: need to pass rc MAC as a string.

miio/yeelight_dual_switch.py
rytilahti: Both are fine by me, however you prefer. Maybe enum is nicer, just in case some future devices will have other types of switches?

miio/yeelight_dual_switch.py
rytilahti: Okay, I'd then rename the variable to `mac` and set the docstring to `Delete remote control by mac` or so.

miio/yeelight_dual_switch.py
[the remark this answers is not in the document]
rytilahti: I suppose the negative input stops the timer, too? And zero turns it off immediately?

IhorSyerkov:
@rytilahti I updated PR, including all your comments. It's ready for review

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,254 @@
...
+                self.interlock,
+                self.flex_mode,
+                self.rc_list,
+            )
+        )
+        return s
+
+
+class YeelightDualControlModule(MiotDevice):
+    """Main class representing the air humidifier which uses MIoT protocol."""
```
rytilahti, line 125: Please update the docstring, and include the model (yeelink.switch.sw1) in it, too.

miio/tests/dummies.py
```diff
@@ -66,6 +66,15 @@ def __init__(self, *args, **kwargs):
     def get_properties_for_mapping(self):
         return self.state
 
+    def get_properties(self, properties):
+        keys = [p["did"] for p in properties]
```
rytilahti, line 70: Considering this modifies the common dummies, could you add a simple docstring describing what's its usecase?

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,254 @@
...
+            "switch_2_state",
+            "switch_2_default_state",
+            "switch_2_off_delay",
+            "interlock",
+            "flex_mode",
+            "rc_list",
+        ]
+        properties = [
+            {"did": k, **v}
+            for k, v in filter(lambda item: item[0] in p, _MAPPING.items())
```
rytilahti, line 166: Please add a note for why this filtering is necessary. Also, I think you should use `get_properties_for_mapping()` here like other implementations do, did you try that already and was it not working?

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,254 @@
...
+        return DualControlModuleStatus(
+            dict(map(lambda v: (v["did"], v["value"]), values))
+        )
+
+    @command(
+        click.argument("switch", type=EnumType(Switch)),
+        default_output=format_output("Turn On {switch} Switch"),
+    )
+    def on(self, switch: Switch):
+        """Turn On Switch."""
```
rytilahti, line 178:
```suggestion
        """Turn switch on."""
```
The docstrings should be consistent between different implementations, so better avoid upper-casing the first letter here and in the `format_output()`s.

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,254 @@
...
+        if switch == Switch.First:
+            return self.set_property("switch_1_state", True)
+        elif switch == Switch.Second:
+            return self.set_property("switch_2_state", True)
+
+    @command(
+        click.argument("switch", type=EnumType(Switch)),
+        default_output=format_output("Turn Off {switch} Switch"),
```
rytilahti, line 186:
```suggestion
        default_output=format_output("Turn off switch {switch}"),
```
Is likely better ordering for console output?

rytilahti (changes requested):
Sorry for the delay! 

A couple of minor things:
* Docstrings & `format_output()`s should be consistent with the casing used by other integrations.
* Check out if the `get_properties_for_mapping()` works for this device

miio/yeelight_dual_switch.py
IhorSyerkov: I use it because not all props from mapping are readable.
For ex. toggle accept only `write` opperation and it throw exception during get status.

rytilahti: I thought so. Please add a note about it, so that it doesn't get forgotten. I hope the mappings can be improved in the future to avoid requesting write-only siid/piid combinations but it's not a priority of mine at the moment.

rytilahti (approved):
LGTM! If the API feels "odd", we can always improve on that in the future, thanks for the PR! :1st_place_medal:

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,255 @@
...
+                {'did': 'switch_2_state', 'siid': 3, 'piid': 1, 'code': 0, 'value': False},
+                {'did': 'switch_2_default_state', 'siid': 3, 'piid': 2, 'code': 0, 'value': False},
+                {'did': 'switch_2_off_delay', 'siid': 3, 'piid': 3, 'code': 0, 'value': 0},
+                {'did': 'interlock', 'siid': 4, 'piid': 1, 'code': 0, 'value': False},
+                {'did': 'flex_mode', 'siid': 4, 'piid': 2, 'code': 0, 'value': True},
+                {'did': 'rc_list', 'siid': 4, 'piid': 2, 'code': 0, 'value': '[{"mac":"9db0eb4124f8","evtid":4097,"pid":339,"beaconkey":"3691bc0679eef9596bb63abf"}]'},
+            ]
+        }
+        """
+        print(data)
```
rytilahti, line 58: Oh, was doing a quick re-check and noticed a couple of things.
```suggestion
```

miio/yeelight_dual_switch.py
```diff
@@ -0,0 +1,255 @@
...
+        self.data = data
+
+    @property
+    def switch_1_state(self) -> bool:
+        return bool(self.data["switch_1_state"])
```
rytilahti, line 63: Could you please add docstrings to the status container properties, too? Then this is good to go!

rytilahti (changes requested):

miio/yeelight_dual_switch.py
IhorSyerkov: forgot )

rytilahti (approved):
:+1: can be merged as soon as the tests pass, thanks again!

nick2525:
Can your please link home assistant pull request, or the pr has not yet been created in the home assistant ?

IhorSyerkov:
I didn't create HA PR yet. But planning to do it.

PR #894, opened 2020-12-24
Support resume_or_start for vacuum's segment cleaning

Sian-Lee-SA:
Added check if in segment cleaning and if so then resume segment clean


rytilahti:
Looks straightforward to me, thanks for the PR! :+1:

Issue #880, opened 2020-12-03
zhimi.heater.mc2 not fully supported


ee02217:
**Describe the bug**
zhimi.heater.mc2 is not fully supported.
Info command works correctly providing the below output:
`$ miiocli heater --ip 192.168.86.94 --token XXXXXXXXXXXXXX info
Model: zhimi.heater.mc2
Hardware version: esp32
Firmware version: 2.0.8
`
However, any other command results in error:
`$ miiocli heater --ip 192.168.86.94 --token XXXXXXXXXXXXXX on
Powering on
Error: {'code': -5001, 'message': 'command error'}`

**Version information (please complete the following information):**
 - OS: Ubuntu 20.04
 - python-miio: miiocli, version 0.5.4

**Device information:**
  - Model: zhimi.heater.mc2
  - Hardware version: esp32
  - Firmware version: 2.0.8

**To Reproduce**
1. run command: `miiocli heater --ip 192.168.86.94 --token XXXXXXXXXXXXXX on`

**Expected behavior**
Heater should turn on.

**Console output**
`Powering on
Error: {'code': -5001, 'message': 'command error'}
`

ee02217:
It seems this should not be a bug, but a new feature, but I dont seem to be able to change that.

syssi:
Miot specs: https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:heater:0000A01A:zhimi-mc2:1

```
Service: siid 1: (Device Information): 4 props, 0 actions
  * Property piid: 1 (Device Manufacturer): (string, unit: None) (acc: ['read'], value-list: [], value-range: None)
  * Property piid: 2 (Device Model): (string, unit: None) (acc: ['read'], value-list: [], value-range: None)
  * Property piid: 3 (Device Serial Number): (string, unit: None) (acc: ['read'], value-list: [], value-range: None)
  * Property piid: 4 (Current Firmware Version): (string, unit: None) (acc: ['read'], value-list: [], value-range: None)
Service: siid 2: (Heater): 3 props, 0 actions
  * Property piid: 1 (Switch Status): (bool, unit: None) (acc: ['read', 'write', 'notify'], value-list: [], value-range: None)
  * Property piid: 2 (Device Fault): (uint8, unit: none) (acc: ['read', 'notify'], value-list: [], value-range: [0, 255, 1])
[...]

PR #895, opened 2020-12-25
Add support for zhimi.heater.mc2

bafonins:
This PR adds support for Xiaomi Mi Smart Space Heater S (zhimi.heater.mc2).

References: https://github.com/rytilahti/python-miio/issues/880
miot spec: https://miot-spec.org/miot-spec-v2/instance\?type\=urn:miot-spec-v2:device:heater:0000A01A:zhimi-mc2:1

<details>
<summary>Details</summary>
Product:
Screenshot 2020-12-25 at 15 21 09

Mi Home UI:
ui1
ui2

</details>

Added unit tests as well as tested the new heater module on an actual device.

Please let me know if I missed anything.

### DIFF
--- a/miio/heater_miot.py
+++ b/miio/heater_miot.py
@@ -91,7 +91,7 @@ class HeaterMiotStatus:
         return self.data["buzzer"] is True
 
     @property
-    def brightness(self) -> LedBrightness:
+    def led_brightness(self) -> LedBrightness:
         """LED indicator brightness."""
         return LedBrightness(self.data["led_brightness"])
 
@@ -108,7 +108,7 @@ class HeaterMiotStatus:
                 self.power,
                 self.target_temperature,
                 self.temperature,
-                self.brightness,
+                self.led_brightness,
                 self.buzzer,
                 self.child_lock,
                 self.countdown_time,
--- a/miio/heater_miot.py
+++ b/miio/heater_miot.py
@@ -123,7 +123,7 @@ class HeaterMiotStatus:
 
 
 class HeaterMiot(MiotDevice):
-    """Main class representing the Xiaomi Smart Space Heater S."""
+    """Main class representing the Xiaomi Smart Space Heater S (zhimi.heater.mc2)."""
 
     def __init__(
         self,
--- a/miio/heater_miot.py
+++ b/miio/heater_miot.py
@@ -26,6 +26,11 @@ _MAPPING = {
     "led_brightness": {"siid": 7, "piid": 3},
 }
 
+HEATER_PROPERTIES = {
+    "temperature_range": (18, 28),
+    "delay_off_range": (0, 12 * 3600),
+}
+
 
 class LedBrightness(enum.Enum):
     On = 0
@@ -71,8 +76,8 @@ class HeaterMiotStatus:
         return self.data["target_temperature"]
 
     @property
-    def countdown_time(self) -> int:
-        """Countdown time."""
+    def delay_off_countdown(self) -> int:
+        """Countdown until turning off in seconds."""
         return self.data["countdown_time"]
 
     @property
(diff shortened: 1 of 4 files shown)


miio/heater_miot.py
```diff
@@ -0,0 +1,209 @@
...
+
+class HeaterMiotException(DeviceException):
+    pass
+
+
+class HeaterMiotStatus:
+    """Container for status reports from the heater which uses the MIoT protocol."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
```
rytilahti, line 43: Please add an example data dictionary to the docstring, see other miot devices for example.

rytilahti (changes requested):
Looks fairly straightforward to me, only a minor change needed before this can be merged. Thanks for the PR!

miio/heater_miot.py
```diff
@@ -0,0 +1,209 @@
...
+        return self.data["power"]
+
+    @property
+    def target_temperature(self) -> int:
+        """Target temperature."""
+        return self.data["target_temperature"]
+
+    @property
+    def countdown_time(self) -> int:
+        """Countdown time."""
```
rytilahti, line 62: What does this indicate? Time until the device gets turned off? Maybe the property naming should follow the same format as in heater?

miio/heater_miot.py
```diff
@@ -0,0 +1,209 @@
...
+                self.buzzer,
+                self.child_lock,
+                self.countdown_time,
+            )
+        )
+        return s
+
+
+class HeaterMiot(MiotDevice):
+    """Main class representing the heater which uses MIoT protocol."""
```
rytilahti, line 108: Please add the model also to this docstring to let readers to know which device(s) are supported.

rytilahti (changes requested):

miio/heater_miot.py
rytilahti: Also, please update the README.md to add this to supported devices.

miio/heater_miot.py
bafonins: Exactly. The `countdown_time` field defines the number hours before the device goes off. 
The problem is here that the spec defines this field as `countdown_time` while for miio heater it is `delay_off_countdown`. Shall overwrite it to match the field name in heater?

bafonins:
@rytilahti fixed according to your comments. let me know if I missed anything

miio/heater_miot.py
rytilahti: Okay, looking at the current code base, there are several instances of `delay_off_countdown` which returns the value in seconds.

So I'd say yes, it's better use that same naming scheme (and return the value in seconds even when the precision is much lower) just to be consistent.

miio/heater_miot.py
```diff
@@ -0,0 +1,222 @@
...
+        """True if child lock is on, False otherwise."""
+        return self.data["child_lock"] is True
+
+    @property
+    def buzzer(self) -> bool:
+        """True if buzzer is turned on, False otherwise."""
+        return self.data["buzzer"] is True
+
+    @property
+    def brightness(self) -> LedBrightness:
```
rytilahti, line 94:
```suggestion
    def led_brightness(self) -> LedBrightness:
```

miio/heater_miot.py
```diff
@@ -0,0 +1,228 @@
...
+                self.buzzer,
+                self.child_lock,
+                self.delay_off_countdown,
+            )
+        )
+        return s
+
+
+class HeaterMiot(MiotDevice):
+    """Main class representing the Xiaomi Smart Space Heater S."""
```
rytilahti, line 126:
```suggestion
    """Main class representing the Xiaomi Smart Space Heater S (zhimi.heater.mc2)."""
```
Please add a note in this docstring about the supported model. 

At some point in the future this project will hopefully have a model<->implementation class type of mapping for automatic generation of supported devices in docs, discovery helper and potential other uses. When the time comes, having this information in the source will make the process much easier.

rytilahti (approved):
One last very minor change and this is good to go. Thanks for the PR! :1st_place_medal:

ee02217:
This is great! Thanks!
How about adding support to Home Assistant integration? 👍

Issue #879, opened 2020-12-02
Any plans on supporting Air Detector Lite PM2.5?


luixal:
**Device information:**
I was thinking about buying the new [Xiaomi/Mijia Air Detector Lite PM2.5][1], but I would like is anyone has one or if it would be compatible with this lib.

  - Name(s) of the device: **Xiaomi/Mijia Air Detector Lite PM2.5 (not sure if this is the official name).**
  - Link: **Couldn't find an official link, [this one is from AliExpress][1] (seems legit, I've seen other news about the device).**

Use `miiocli device --ip <ip address> --token <token>`.

**I can't as I don't own a unit yet.**

  - Model: [e.g., lumi.gateway.v3]
  - Hardware version:
  - Firmware version:

**Additional context**
Thank you! :)

[1]: https://es.aliexpress.com/item/1005001738417932.html?src=google&albch=shopping&acnt=439-079-4345&isdl=y&slnk=&plac=&mtctp=&albbt=Gploogle_7_shopping&aff_atform=google&aff_short_key=UneMJZVf&gclsrc=aw.ds&&albagn=888888&&ds_e_adid=475881929441&ds_e_matchtype=&ds_e_device=c&ds_e_network=u&ds_e_product_group_id=743612850954&ds_e_product_id=es1005001738417932&ds_e_product_merchant_id=109241272&ds_e_product_country=ES&ds_e_product_language=es&ds_e_product_channel=online&ds_e_product_store_id=&ds_url_v=2&ds_dest_url=https://es.aliexpress.com/item/1005001738417932.html?&albcp=11492995081&albag=117794251848&gclid=CjwKCAiA8Jf-BRB-EiwAWDtEGsvGzMXvdMnG0Qpnop8UJDS9BlyPBDQGF1mIUjUXLmecCub82k82WRoC6BsQAvD_BwE

syssi:
The device isn't supported right now.

rezmus:
miot model cgllc.airm.cgdn1

syssi:
Miot specs: https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:air-monitor:0000A008:cgllc-cgdn1:1

```
Service: siid 1: (Device Information): 4 props, 0 actions
  * Property piid: 1 (Device Manufacturer): (string, unit: None) (acc: ['read'], value-list: [], value-range: None)
[...]

PR #899, opened 2021-01-01
Add support for Xiaomi Air purifier 3C 

arturdobo:
This PR adds support for Air purifier 3C (`zhimi.airpurifier.mb4`) which according to  [specification](https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:air-purifier:0000A007:zhimi-mb4:2) provides only a subset of i.e. mb3 functionalities.

PR #900, opened 2021-01-01
Add Qingping Air Monitor Lite support (cgllc.airm.cgdn1)

arturdobo:
This PR adds support for Qingping Air Monitor Lite support (`cgllc.airm.cgdn1`).

Specification: https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:air-monitor:0000A008:cgllc-cgdn1:1

Closes: https://github.com/rytilahti/python-miio/issues/879

Support `start_time` and `end_time` has been intentionally omitted in this PR and can be added in the other one. I didn't check whether the time zone is necessary to compute the current time. If yes, I wasn't able to read `time_zone` from my device, but `cgllc-cgdn1:2` spec mentions that property. What's more Xiaomi Home reads somehow this property from the device.

Another thing is I had no idea what `0` regarding `monitoring_frequency` could mean.

### DIFF
--- a/miio/airqualitymonitor_miot.py
+++ b/miio/airqualitymonitor_miot.py
@@ -1,0 +1,254 @@
+import enum
+import logging
+
+import click
+
+from .click_common import command, format_output
+from .exceptions import DeviceException
+from .miot_device import MiotDevice
+
+_LOGGER = logging.getLogger(__name__)
+
+MODEL_AIRQUALITYMONITOR_CGDN1 = "cgllc.airm.cgdn1"
+
+_MAPPING_CGDN1 = {
+    # Source https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:air-monitor:0000A008:cgllc-cgdn1:1
+    # Environment
+    "humidity": {"siid": 3, "piid": 1},  # [0, 100] step 1
+    "pm25": {"siid": 3, "piid": 4},  # [0, 1000] step 1
+    "pm10": {"siid": 3, "piid": 5},  # [0, 1000] step 1
+    "temperature": {"siid": 3, "piid": 7},  # [-30, 100] step 0.00001
+    "co2": {"siid": 3, "piid": 8},  # [0, 9999] step 1
+    # Battery
+    "battery": {"siid": 4, "piid": 1},  # [0, 100] step 1
+    "charging_state": {
+        "siid": 4,
+        "piid": 2,
+    },  # 1 - Charging, 2 - Not charging, 3 - Not chargeable
+    "voltage": {"siid": 4, "piid": 3},  # [0, 65535] step 1
+    # Settings
+    "start_time": {"siid": 9, "piid": 2},  # [0, 2147483647] step 1
+    "end_time": {"siid": 9, "piid": 3},  # [0, 2147483647] step 1
+    "monitoring_frequency": {"siid": 9, "piid": 4},  # 1, 60, 300, 600, 0
+    "screen_off": {"siid": 9, "piid": 5},  # 15, 30, 60, 300, 0
+    "device_off": {"siid": 9, "piid": 6},  # 15, 30, 60, 0
+    "temperature_unit": {"siid": 9, "piid": 7},
+}
+
+
+class AirQualityMonitorMiotException(DeviceException):
+    pass
+
+
+class ChargingStateCGDN1(enum.Enum):
+    Unplugged = 0  # Not mentioned in the spec
+    Charging = 1
+    NotCharging = 2
+    NotChargable = 3
+
+
+class MonitoringFrequencyCGDN1(enum.Enum):
+    Every1Second = 1
+    Every1Minute = 60
+    Every5Minutes = 300
+    Every10Minutes = 600
+    NotSet = 0
+
+
+class ScreenOffCGDN1(enum.Enum):
+    After15Seconds = 15
+    After30Seconds = 30
+    After1Minute = 60
+    After5Minutes = 300
+    Never = 0
+
+
+class DeviceOffCGDN1(enum.Enum):
+    After15Minutes = 15
+    After30Minutes = 30
+    After1Hour = 60
+    Never = 0
+
+
+class DisplayTemeratureUnitCGDN1(enum.Enum):
+    Celcius = "c"
+    Fahrenheit = "f"
+
+
+class AirQualityMonitorCGDN1Status:
+    """Container of air quality monitor CGDN1 status."""
+
+    def __init__(self, data):
+        self.data = data
+
+    @property
+    def humidity(self) -> int:
+        """Return humidity value (0...100%)."""
+        return self.data["humidity"]
+
+    @property
+    def pm25(self) -> int:
+        """Return PM 2.5 value (0...1000ppm)."""
+        return self.data["pm25"]
+
+    @property
+    def pm10(self) -> int:
+        """Return PM 10 value (0...1000ppm)."""
+        return self.data["pm10"]
+
+    @property
+    def temperature(self) -> float:
+        """Return temperature value (-30...100°C)."""
+        return self.data["temperature"]
+
+    @property
+    def co2(self) -> int:
+        """Return co2 value (0...9999ppm)."""
+        return self.data["co2"]
+
+    @property
+    def battery(self) -> int:
+        """Return battery level (0...100%)."""
+        return self.data["battery"]
+
+    @property
+    def charging_state(self) -> ChargingStateCGDN1:
+        """Return charging state."""
+        return ChargingStateCGDN1(self.data["charging_state"])
+
+    @property
+    def monitoring_frequency(self) -> int:
+        """Return monitoring frequency time."""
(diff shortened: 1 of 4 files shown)


miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,254 @@
...
+    "device_off": {"siid": 9, "piid": 6},  # 15, 30, 60, 0
+    "temperature_unit": {"siid": 9, "piid": 7},
+}
+
+
+class AirQualityMonitorMiotException(DeviceException):
+    pass
+
+
+class ChargingStateCGDN1(enum.Enum):
```
rytilahti, line 43:
```suggestion
class ChargingState(enum.Enum):
```
No need to repeat the CGDN1 in class names as these enums may also be useful for some other models in the future.

miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,254 @@
...
+    NotCharging = 2
+    NotChargable = 3
+
+
+class MonitoringFrequencyCGDN1(enum.Enum):
+    Every1Second = 1
+    Every1Minute = 60
+    Every5Minutes = 300
+    Every10Minutes = 600
+    NotSet = 0
```
rytilahti, line 55: Are these the only values accepted by the device, or is it possible to use something like 120, too? If the latter, it'd be better to simply use an integer describing the frequency in seconds.

miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class ScreenOffCGDN1(enum.Enum):
+    After15Seconds = 15
+    After30Seconds = 30
+    After1Minute = 60
+    After5Minutes = 300
+    Never = 0
```
rytilahti, line 63: Same question as above, have you tried if you can input arbitrary values here?

miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,254 @@
...
+class DisplayTemeratureUnitCGDN1(enum.Enum):
+    Celcius = "c"
+    Fahrenheit = "f"
+
+
+class AirQualityMonitorCGDN1Status:
+    """Container of air quality monitor CGDN1 status."""
+
+    def __init__(self, data):
+        self.data = data
```
rytilahti, line 82: Add an example `data` content to the docstring, that can be helpful for anyone looking at the code without having access to such device.

rytilahti (commented):
Thanks for the PR! Here's a very brief initial review.

miio/airqualitymonitor_miot.py
arturdobo: It looks like the device accept:
`0...600 s` for monitoring frequency option
`0..300 s` for screen off option
`0..60 m` for device off option

But worth noting this can be considered as a kind of undocumented feature because the spec lists strictly possible options. So they could fix that one day on the firmware update. Another issue is the Xiaomi Home app shows wrongly what option is selected (see the attachment).

So it's up to you, whether to change it or not.

photo_2021-01-08_14-41-48

miio/airqualitymonitor_miot.py
arturdobo: Done

miio/airqualitymonitor_miot.py
rytilahti: It's similar situation as with vacuum fan speeds, the official app may support only specific, hard-coded speeds but you can adjust them as you like. I'd say use integers instead of enums here, I prefer to give users that choice if they wish to use "unofficial" values (I do this all the time with my vacuum).

If you wish to do so, you could add methods/properties to return the "presets" based on the specified values, to help downstream users of the lib / document the "official" values.

arturdobo: I made `int` to be accepted instead of enums. And similar to I believe the vacuum case I created properties with the official spec options. Let me please know is it now better.

miio/airqualitymonitor_miot.py
arturdobo: Done

miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class DeviceOffCGDN1(enum.Enum):  # Official spec options
+    After15Minutes = 15
+    After30Minutes = 30
+    After1Hour = 60
+    Never = 0
+
+
+class DisplayTemeratureUnitCGDN1(enum.Enum):
```
rytilahti, line 82:
```suggestion
class DisplayTemperatureUnit(enum.Enum):
```

miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,286 @@
...
+    Celcius = "c"
+    Fahrenheit = "f"
+
+
+class AirQualityMonitorCGDN1Status:
```
rytilahti, line 87:
```suggestion
class AirQualityMonitorStatus:
```

miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,286 @@
...
+      'temperature_unit': 'c'
+    }
+
+    """
+
+    def __init__(self, data):
+        self.data = data
+        self._monitoring_frequency_options = MonitoringFrequencyCGDN1
+        self._screen_off_options = ScreenOffCGDN1
+        self._device_off_options = DeviceOffCGDN1
```
rytilahti, line 114: If you want to expose the enums, better do that not in the status class, but in the device class.

miio/airqualitymonitor_miot.py
arturdobo: My intention was not to build any generic air quality monitor now, because I'm not sure whether the set of functionalities that `CGDN1` supports are anyhow representative for a wider array of sensors. I think `AirQualityMonitor` can always be created in the future, but establishing it now can result in breaking changes, because other AirQualityMonitorXYZ may support only i.e. `pm 2.5`.

Last time I experienced a similar issue while adding Xiaomi AirPurifier 3C in this PR https://github.com/rytilahti/python-miio/pull/899

rytilahti: Fair enough!

miio/airqualitymonitor_miot.py
rytilahti: Fine for me. Please take a look at my comment how these are exposed though, that should be changed but this PR can be merged after that's done.

miio/airqualitymonitor_miot.py
rytilahti: Oh, the typo is still there, should still be `TemperatureUnit` instead of `TemeratureUnit`.

miio/airqualitymonitor_miot.py
```diff
@@ -0,0 +1,286 @@
...
+        ip: str = None,
+        token: str = None,
+        start_id: int = 0,
+        debug: int = 0,
+        lazy_discover: bool = True,
+    ) -> None:
+        super().__init__(_MAPPING_CGDN1, ip, token, start_id, debug, lazy_discover)
+        self._monitoring_frequency_options = MonitoringFrequencyCGDN1
+        self._screen_off_options = ScreenOffCGDN1
+        self._device_off_options = DeviceOffCGDN1
```
rytilahti, line 212: Do you have a specific use case for these (underscored variables are generally considered internal in python, i.e., not to be used anyone outside the class itself)?

miio/airqualitymonitor_miot.py
arturdobo: ah, fixed

miio/airqualitymonitor_miot.py
arturdobo: No, I just saw how it's done in `vacuum` case, but here it should be exposed to be used publicly

rytilahti: Okay, I think this can be merged now. Just one more very nitpicky issue, I'm a bit unsure how useful those are, as you could access them anyway simply be importing the corresponding classes? I'm not against having them there, but I don't personally see anyone using `self.screen_off_options.SomeOption` instead of `ScreenOffCGDN1.SomeOption`.

What do you think?

arturdobo: I thought there is a kind of convention to having them here, but yeah, I don't any advantage keeping in here either. Maybe one of the reasons would be to explicitly say, those enums are somehow bound with this particular class, but they can be accessed directly anyway. So I'll remove them

rytilahti:
LGTM, one last thing before merging this, could you add an entry to README's supported devices, @arturdobo?

arturdobo:
@rytilahti updated, but no idea why the build failed

rytilahti:
Looks like a problem at readthedocs' side, could you rebase the README change though? It's conflicting currently with a recently merged PR.

arturdobo:
@rytilahti done

rytilahti:
Great, thanks for the PR! :tada:

The readthedocs problem seems to be related new rust compiler dependency on cryptography lib, which is not available there. I hope that'll get fixed by itself at some point.

Issue #901, opened 2021-01-06
A guide how to communicate with MIOT devices -> Plugin?


SpeedFire0:
I was able to get the data from an unsupported device (HeatCold Heating Floor Controller - cubee.airrtc.th123e) and to set the required parameters from a Terminal on Mac. Can someone help with converting this into a working plugin for Home Assistant and then HomeKit?

Here is my step-by-step guide for all newcomers who want to play with their devices:
1) Install python-miio on Mac. Enter this in Terminal app on Mac:
`sudo pip3 install python-miio`

2) Get model name of your device ("cubee.airrtc.th123e" in my case), IP address (like "192.168.1.152") and Token (like "b110204d86732b019d3d6axxxxb9ad3a"). The easiest way for me was to use this integration for Home Assistant - https://github.com/AlexxIT/XiaomiGateway3

3) Check that your device is accessible by entering the following in a Terminal app on Mac:
`miiocli -d device --ip 192.168.1.152 --token b110204d86732b019d3d6axxxxb9ad3a info`
Note 1: You need to put your device's IP and Token here.
Note 2: "-d" is for Debug and can be excluded for a shorter output.
In my case I got something like this:
"Model: cubee.airrtc.th123e
Hardware version: esp32
Firmware version: 2.1.7"
Device is responding! Nice! Going to the next step.

4) Find your device in the list of MIOT devices: 
http://miot-spec.org/miot-spec-v2/instances?status=all
I searched for "cubee.airrtc.th123e" and found the following line:
`{"status":"released","model":"cubee.airrtc.th123e","version":1,"type":**"urn:miot-spec-v2:device:thermostat:0000A031:cubee-th123e:1"**},`

5) Copy the "urn" part after this url:
https://miot-spec.org/miot-spec-v2/instance?type=
in my case: https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:thermostat:0000A031:cubee-th123e:1
You will see an unformatted JSON text of the device's specs

6) Put this JSON text to any JSON Formatter to make it more human readable. For example here: http://json.parser.online.fr/

7) You will see a hierarchy with a list of Services (later used as "siid"), each of which has a list of Properties ("piid"):
<img width="411" alt="miot_json" src="https://user-images.githubusercontent.com/17302232/103826006-09ca6c00-5087-11eb-875a-e4b7dbd9fffb.png">

8) You need to find some property in the list which you want to get from the device and which is readable 
[...]

PR #905, opened 2021-01-07
Improve MiotDevice API (get_property_by, set_property_by, call_action, call_action_by)

rytilahti:
* `get_property_by(siid, piid)` returns a single property, useful for testing with miiocli
* `set_property_by(siid, piid, value, value_type)` allows setting one, `value_type` is optional and will be used for casting to correct type before passing the value to `send()`.
* `call_action(name, params)` calls an action with the given parameters, anything inside `mapping` that has `aiid` is considered an action.
* `call_action_by(siid, aiid, params)` allows calling arbitrary actions, useful especially for command-line use and testing.

* Also, code cleanup for passing the mapping.
* Now the mapping is a class attribute, avoiding unnecessary __init__ overloading

Testers wanted as I have no test devices!

Related to #901 (no more raw-commands necessary for the simplest test cases).

### DIFF
--- a/miio/miot_device.py
+++ b/miio/miot_device.py
@@ -1,24 +1,26 @@
 import logging
+from enum import Enum
+from typing import Any, Union
 
+import click
+
+from .click_common import EnumType, command
 from .device import Device
 
 _LOGGER = logging.getLogger(__name__)
 
 
+class MiotValueType(Enum):
+    Int = int
+    Float = float
+    Bool = bool
+    Str = str
+
+
 class MiotDevice(Device):
     """Main class representing a MIoT device."""
 
-    def __init__(
-        self,
-        mapping: dict,
-        ip: str = None,
-        token: str = None,
-        start_id: int = 0,
-        debug: int = 0,
-        lazy_discover: bool = True,
-    ) -> None:
-        self.mapping = mapping
-        super().__init__(ip, token, start_id, debug, lazy_discover)
+    mapping = None
 
     def get_properties_for_mapping(self) -> list:
         """Retrieve raw properties based on mapping."""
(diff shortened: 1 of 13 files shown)


gudvinr:
I tested these options with `mmgg.pet_waterer.s1` which I have on hand.  
* `get_property_by` works
* `set_property_by` works, but I only tried `bool` properties since this device doesn't really have many
* `call_action_by` works too, but I tried single action which has no inputs and outputs

rytilahti:
@arturdobo @darckly @bafonins if you don't mind testing this PR on your miot devices, it would be great! The conversion from using passed `_MAPPING` to a mapping defined inside the implementing class may have some glitches.

darckly:
I testes options with my Huizuo lamps - get_properties_by and set_properties_by are working as expected, the fourth parameter (value_type) should be always specified.  
Checked with power (bool), brightness (int), and temperature (int). 

call_action and call_action_by methods I can't test

arturdobo:
Sorry for such a late response. I was able to call `status` on
* `zhimi.airpurifier.mb4`
* `zhimi.airpurifier.m2`
* `zhimi.humidifier.ca4`
* `zhimi.airpurifier.v7`
* `zhimi.airpurifier.mb3`

miio/miot_device.py
```diff
@@ -1,38 +1,117 @@
...
-        mapping: dict,
-        ip: str = None,
-        token: str = None,
-        start_id: int = 0,
-        debug: int = 0,
-        lazy_discover: bool = True,
-    ) -> None:
-        self.mapping = mapping
-        super().__init__(ip, token, start_id, debug, lazy_discover)
+    mapping = None
```
ha0y, line 31: Hello,
This seems to be a breaking change for ``MiotDevice``. We can no longer use 
```python
MiotDevice(ip=host, token=token, mapping=mapping)
```
to instantiate a MiotDevice with a given mapping (ha0y/xiaomi_miot_raw#79).
Is there a method to keep backward compatibility?

rytilahti: Hi, sorry about that! You could assign the mapping directly after constructing the `MiotDevice` object to fix this for now:
```
dev = MiotDevice(...)
dev.mapping = my_mapping
```

Otherwise, feel free to open an issue to re-add the kwarg variant of this parameter back to the constructor!

ha0y: Thanks for your prompt reply!

ha0y: Unfortunatelly, this doesn't solve the problem, as this code doesn't work in version 0.5.4. `mapping` is required in 0.5.4 but unexpected in 0.5.5. I have opened an issue #982. Hope this will be fixed soon because it *is* a breaking change and may affect those who uses this library.

rytilahti: Indeed, I hadn't thought about that, sorry... See my comment on the issue how I plan to rectify it.

PR #914, opened 2021-01-20
Add docformatter to pre-commit hooks

rytilahti:
Docformatter is something like black but for docstring formatting: https://github.com/myint/docformatter

This PR adds it to the pre-commit hooks, and mass-converts the whole code base using it. If there are some cases where the docstring summary should include some text (e.g., for click), these needs to be modified manually.

### DIFF
--- a/miio/airconditioningcompanion.py
+++ b/miio/airconditioningcompanion.py
@@ -140,11 +140,16 @@ class AirConditioningCompanionStatus:
 
     @property
     def air_condition_remote(self) -> int:
-        """Known remote ids:
+        """Remote id.
 
-        0x80111111, 0x80111112 (brand: 0x0182) 0x80222221 (brand: 0x0097) 0x80333331
-        (brand: 0x0037) 0x80444441 (brand: 0x0202) 0x80555551 (brand: 0x2782) 0x80777771
-        (brand: 0x0197) 0x80666661 (brand: 0x0192)
+        Known remote ids:
+        * 0x80111111, 0x80111112 (brand: 0x0182)
+        * 0x80222221 (brand: 0x0097)
+        * 0x80333331 (brand: 0x0037)
+        * 0x80444441 (brand: 0x0202)
+        * 0x80555551 (brand: 0x2782)
+        * 0x80777771 (brand: 0x0197)
+        * 0x80666661 (brand: 0x0192)
         """
         return int(self.air_condition_model[4:8].hex(), 16)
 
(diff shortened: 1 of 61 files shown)


miio/airconditioningcompanion.py
```diff
@@ -133,33 +132,25 @@ def device_type(self) -> int:
...
-        0x80111111, 0x80111112 (brand: 0x0182)
-        0x80222221 (brand: 0x0097)
-        0x80333331 (brand: 0x0037)
-        0x80444441 (brand: 0x0202)
-        0x80555551 (brand: 0x2782)
-        0x80777771 (brand: 0x0197)
-        0x80666661 (brand: 0x0192)
+        """Known remote ids:
 
+        0x80111111, 0x80111112 (brand: 0x0182) 0x80222221 (brand: 0x0097) 0x80333331
```
syssi, line 145: Do you like this new format? I don't have a strong opinion on that. It looks like some docstrings arent' as readible as before now.

syssi (approved):

miio/airconditioningcompanion.py
rytilahti: Hmm, indeed, thanks for spotting! The previous one looks much better in this case, so this should definitely be fixed. 

Maybe this happens as the original got no real summary-line? Or maybe as it is not a list (in which case adding a `-` or `*` will probably fix that)?

I'll take a look into that, and adjust accordingly.

miio/airconditioningcompanionMCN.py
```diff
@@ -40,11 +40,10 @@ class AirConditioningCompanionStatus:
...
 
     def __init__(self, data):
-        """
-        Device model: lumi.acpartner.mcn02
+        """Device model: lumi.acpartner.mcn02.
 
-        Response of "get_prop, params:['power', 'mode', 'tar_temp', 'fan_level', 'ver_swing', 'load_power']":
-        ['on', 'dry', 16, 'small_fan', 'off', 84.0]
+        Response of "get_prop, params:['power', 'mode', 'tar_temp', 'fan_level',
+        'ver_swing', 'load_power']": ['on', 'dry', 16, 'small_fan', 'off', 84.0]
```
rytilahti, line 46: Here's similar case where the original was much prettier, and there's probably some more. Time to do some manual adjustments :-)

miio/airconditioningcompanion.py
rytilahti: Fixed now.

syssi (approved):

PR #924, opened 2021-01-30
Refactor & improve support for gateway devices

starkillerOG:
Due to a merge conflict PR #782 was closed, this is the continueation of that PR.

Restructure the gateway devices to put everything with simular API together.

This PR also creates the base classes for switches and light bulbs, so all known models are now fully supported, although some models have not been tested (they are assummed to have simular API calls as devices that are closely related).

This PR also adds subdevice support for the lumi.gateway.mgl03 and is a first step towards subdevice support for the lumi.gateway.mieu01


starkillerOG:
@rytilahti I finally got around to implementing the changes as you suggested in your rivieuw: https://github.com/rytilahti/python-miio/pull/782#issuecomment-704478051

I think this is now ready for merging.
Could you take another look?

starkillerOG:
@rytilahti can you merge this?
All checks pass.

If this is merged I can continue with implementing the subdevices in HomeAssistant (such as the switches and relays etc.)

miio/gateway/gateway.py
```diff
@@ -0,0 +1,335 @@
...
+
+        self._alarm = Alarm(parent=self)
+        self._radio = Radio(parent=self)
+        self._zigbee = Zigbee(parent=self)
+        self._light = Light(parent=self)
+        self._devices = {}
+        self._info = None
+        self._subdevice_model_map = None
+
+    def _get_subdevice_model_map(self):
```
rytilahti, line 97: Instead of having to call this prior to accessing `self._subdevice_model_map`, how about defining a property `subdevice_model_map` which does the initialization during the first access?

Something like:
```
@property
subdevice_model_map():
    if self._subdevice_model_map is None:
        # initialize 
    return self._subdevice_model_map
```

That would simplify all the places you need this information :-)

miio/gateway/gateway.py
```diff
@@ -0,0 +1,335 @@
...
+        return self._info.model
+
+    @command()
+    def discover_devices(self):
+        """
+        Discovers SubDevices
+        and returns a list of the discovered devices.
+        """
+
+        from .devices import SubDeviceInfo
```
rytilahti, line 151: This should be moved to the beginning of the file.

miio/gateway/gateway.py
```diff
@@ -0,0 +1,335 @@
...
+        return self._devices
+
+    @command(click.argument("zigbee_model", "sid"))
+    def match_zigbee_model(self, zigbee_model, sid):
+        """
+        Match the zigbee_model to obtain the model_info
+        """
+
+        self._get_subdevice_model_map()
+
```
rytilahti, line 201: See my previous comment about converting the map to a property.

miio/gateway/gateway.py
```diff
@@ -0,0 +1,335 @@
...
+        )
+        return self._get_unknown_model()
+
+    @command(click.argument("dev_info", "model_info"))
+    def setup_device(self, dev_info, model_info):
+        """
+        Setup a device using the SubDeviceInfo and model_info
+        """
+
+        from .devices import SubDevice
```
rytilahti, line 242: Move to the top of the file.

rytilahti (commented):
Sorry for the delay @starkillerOG, this looks great! Just a couple of very minor changes and then we can merge this.

miio/gateway/gateway.py
starkillerOG: done

miio/gateway/gateway.py
starkillerOG: Greath suggestion, I already made a commit for this.
Thanks!

miio/gateway/gateway.py
starkillerOG: That is not possible:
Because a gateway can be supplied to the init of the `SubDevice` class, there is a argument type `Gateway`.
Therefore at the top of the `devices.py` file the `Gateway` class is imported from the "gateway.py" file.

Therefore somthing from "devices.py" can only be imported from the "gateway.py" file after the line in which the  `Gateway` class is defined, otherwise you create a cyclic import (a loop that will endlessly jump between the files to try and import something from one another. (same thing holds for the `GatewayException` class).

It can therefore only be moved up (but below the `GatewayException` class) if the type definition is removed from the `__init__` of the `SubDevice` class.

Do you want to keep it like this, or remove the type definition?

miio/gateway/gateway.py
starkillerOG: See previous comment

miio/gateway/gateway.py
```diff
@@ -141,6 +133,15 @@ def model(self):
             self._info = self.info()
         return self._info.model
 
+    @property
+    def subdevice_model_map(self):
+        """Return the subdevice model map."""
+        # Check if catch already has the subdevice_model_map, otherwise read it.
```
rytilahti, line 139: No need for comments that are obvious from the context :-)

miio/gateway/gateway.py
rytilahti: Ah, gotcha! You can remove the import (in `subdevice.py`), and simply use the class name in quotation marks and it will work just as well as imported type hint. Would that work?

starkillerOG: yea that will probably work, will change it and test it now.

miio/gateway/gateway.py
starkillerOG: Remove that

miio/gateway/gateway.py
starkillerOG: I tested it and it works, just pushed the commit.

miio/gateway/devices/subdevice.py
```diff
@@ -0,0 +1,240 @@
+"""Xiaomi Gateway subdevice base class."""
+
+import logging
+from typing import Optional
```
rytilahti, line 4:
```suggestion
from typing import Optional, TYPE_CHECKING
```

miio/gateway/devices/subdevice.py
```diff
@@ -0,0 +1,240 @@
...
+
+import attr
+import click
+
+from ...click_common import command
+from ..gateway import GATEWAY_MODEL_EU, GATEWAY_MODEL_ZIG3, GatewayException
+
+_LOGGER = logging.getLogger(__name__)
+
+
```
rytilahti, line 14:
```suggestion
if TYPE_CHECKING:
    from ..gateway import Gateway
```

rytilahti (commented):
Sorry, I forgot that the checkers will require the import anyway. Here's what you can do to fix that flake8 error about the gateway class.

Btw, if you run `pre-commit install` the repository, the linting checks will be run locally before you can even commit (makes it faster than waiting for CI to check for isort/black/others)  :-)

starkillerOG:
> pre-commit install

Thanks for the tip, but I use Github Desktop, Not that fimiliar with the github command line yet.
Is there also a single command from poetry that runs all checks?
(or a simular command I can run from windows cmd in the github dir to run all checks at once?)

starkillerOG:
@rytilahti all checks now pass

rytilahti:
Great! Thanks for your patience and hard work on refactoring this! :tada:

Issue #925, opened 2021-02-01
Filter out value 1 from the property AQI


bieniu:
**Is your feature request related to a problem? Please describe.**
Currently, the value of the AQI attribute of air purifier in Home Assistant is sometimes set to `1` for a short period of time. The device display never shows these values.

![air_purifier](https://user-images.githubusercontent.com/478555/106430224-c982c180-646b-11eb-9c0f-9c83cf5c4a99.png)

**Describe the solution you'd like**
`python-miio` should filter out values `<=1` from `aqi` property.

**Device information:**
If the enhancement is device-specific, please include also the following information.

  - Name(s) of the device: Xiaomi Mi Air Purifier 3H
  - Link: 

Use `miiocli device --ip <ip address> --token <token>`.

  - Model: `zhimi.airpurifier.mb3`
  - Hardware version: `esp32`
  - Firmware version: `2.1.8`

**Additional context**

syssi:
Could you increase the log level and provide some traffic? Does the device response contain `aqi: 1` sometimes?

```
logger:
  default: warn
  logs:
    homeassistant.components.xiaomi_miio: debug
    miio: debug
```

bieniu:
Sure, I'll provide the log.

syssi:
In the mean time you could pipe the measurements through the statistics component:

https://www.home-assistant.io/integrations/statistics/

bieniu:
That's exactly what I do 😁

bieniu:
I can confirm, the device sends value `1` for `aqi`.
```
[...]

PR #930, opened 2021-02-04
[miot air purifier] Return None if aqi is 1

bieniu:
Probably the air purifier sends a `1` for `aqi` when the measurement failed.
With change from this PR, library returns `None` as `aqi` value in such situation.

Fixes: https://github.com/rytilahti/python-miio/issues/925

### DIFF
--- a/miio/airpurifier_miot.py
+++ b/miio/airpurifier_miot.py
@@ -84,7 +84,7 @@ class AirPurifierMiotStatus:
     @property
     def aqi(self) -> int:
         """Air quality index."""
-        # zhimi-airpurifier-mb3 returns 1 as AQI value if the measurement was 
+        # zhimi-airpurifier-mb3 returns 1 as AQI value if the measurement was
         # unsuccessful
         if self.data["aqi"] == 1:
             return None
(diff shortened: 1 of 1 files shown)


syssi:
Could you add a comment for which model this fix is required? At the moment the miot class provides the support for these models:

```
    "zhimi-airpurifier-mb3": AirPurifierMiot,  # mb3 (3/3H)
    "zhimi-airpurifier-ma4": AirPurifierMiot,  # ma4 (3)
```

rytilahti:
LGTM, thanks :+1:

sgurgul:
It seems that the side effect of this fix is that there is no AQI in case the airpurifier just returns AQI=1, not as an error but as a measured value. 

I just re-confirmed manually - turning on the device, it shows 1 on the screen.
When I imitated some pollution in the air it started returning correct value.

So, it seems AQI=1 => error does not always seems to be true. 
Looking at #925 it seems those incorrect 1 values are peaks. Maybe it would be better to ignore a single "1" value but if they are coming again and again treat them as correct / measured values ?

PR #933, opened 2021-02-07
vacuum: skip pausing on s50 and s6 maxv before return home call

### LINKED ISSUE #918: [Roborock S6 MaxV] Need a delay between pause and charge commands to return to dock
https://github.com/rytilahti/python-miio/blob/5730f535d4691e0c2b4260cd7835b7a43ddc8ab0/miio/vacuum.py#L145:L148

When my Roborock S6 is running, if I run the "home" function, it only makes it pause. I need to run the "home" function a second time to make it go to base.

It needs a delay bewteen the "app_pause" and the "app_charge" commands.

rytilahti:
Adds a blacklist of vacuums that should not get a `app_pause` prior `app_charge` call.

Fixes #918

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -93,6 +93,7 @@ class WaterFlow(enum.Enum):
 
 
 ROCKROBO_V1 = "rockrobo.vacuum.v1"
+ROCKROBO_S5 = "roborock.vacuum.s5"
 ROCKROBO_S6_MAXV = "roborock.vacuum.a10"
 
 
@@ -149,6 +150,7 @@ class Vacuum(Device):
             self._autodetect_model()
 
         SKIP_PAUSE = [
+            ROCKROBO_S5,
             ROCKROBO_S6_MAXV,
         ]
 
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -93,6 +93,7 @@ class WaterFlow(enum.Enum):
 
 
 ROCKROBO_V1 = "rockrobo.vacuum.v1"
+ROCKROBO_S6_MAXV = "roborock.vacuum.a10"
 
 
 class Vacuum(Device):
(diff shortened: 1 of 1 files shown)


stephanerosi:
Just checked out locally and tested.
Works perfectly !

pergolafabio:
This can be implemented too for s50? Mine also pauses first on a return home command

rytilahti:
@pergolafabio added, thanks for reporting!

pergolafabio:
np, i dont know how long the issue was
but for my s50 , when i do a return to home command, when its cleaning, it did indeed always paused
i need todo the return home service twice

stephanerosi:
I cannot tell neither since when it happened, I've just begun to use this integration...

stephalltricks:
@rytilahti Hello, any news on when this will get merged?
Thanks!

rytilahti:
Thanks for the reminder, let's get this merged! I'm not sure why tests are failing (no logs available), nor can I restart them, but that can be fixed in another PR.

Sian-Lee-SA:
Could you possibly add the S6 versioin to this? just did a test and works without "app_pause"

Sian-Lee-SA:
Ive also noticed that if I start a segment clean while the vacuum is already cleaning then it just stops and doesn't do the segment clean, I assume this issue is the same the other way around but I havent tried that. I had tested a check whether the vacuum is idle or not and if not, then stop first and wait 3 seconds then start the segment clean which works for me

pergolafabio:
Hi

i am running home assistant, seems when i do a return command when its cleaning, its still in pause state? according to components, 0.5.5 is loaded in HA? why is it still pausing?
https://github.com/home-assistant/core/blob/dev/homeassistant/components/xiaomi_miio/manifest.json

rytilahti:
@Sian-Lee-SA feel free to open a new issue (or a PR to add the S6 to https://github.com/rytilahti/python-miio/blob/master/miio/vacuum.py#L153)! The segment cleaning sounds like a separate issue, so please open another one for that.

@pergolafabio please open a new issue on that, and include the model information to the report. There are several different vacuums that may behave differently than the one I have used in development (gen1, old firmware).

I'm now closing this issue for discussion, it makes no sense to pollute this already merged PR for issues.

PR #935, opened 2021-02-07
add EU_gateway support through cloud

starkillerOG:
I know a lot of people are waiting on this.
@rytilahti would this approach be acceptable?

### DIFF
--- a/miio/gateway/gateway.py
+++ b/miio/gateway/gateway.py
@@ -80,6 +80,9 @@ class Gateway(Device):
         start_id: int = 0,
         debug: int = 0,
         lazy_discover: bool = True,
+        cloud_username: str = None,
+        cloud_password: str = None,
+        cloud_country: str = "de",
     ) -> None:
         super().__init__(ip, token, start_id, debug, lazy_discover)
 
@@ -97,6 +100,10 @@ class Gateway(Device):
         self._devices = {}
         self._info = None
         self._subdevice_model_map = None
+        self._cloud_username = cloud_username
+        self._cloud_password = cloud_password
+        self._cloud_country = cloud_country
+        self._did = None
 
     def _get_unknown_model(self):
         for model_info in self.subdevice_model_map:
@@ -130,9 +137,15 @@ class Gateway(Device):
         return self._devices
 
     @property
+    def mac(self):
+        """Return the mac address of the gateway."""
+        if self._info is None:
+            self._info = self.info()
+        return self._info.mac_address
+
+    @property
     def model(self):
         """Return the zigbee model of the gateway."""
-        # Check if catch already has the gateway info, otherwise get it from the device
         if self._info is None:
             self._info = self.info()
         return self._info.model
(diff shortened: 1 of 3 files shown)


rytilahti:
I'd rather have a clean separation from the cloud for anything under the device implementations under `miio` package.

Am I understanding the code correctly, that the cloud connectivity is just needed for accessing the token & the list of devices?
If yes, I'd propose improving micloud to dump the required information into a local file, and having `get_device_from_file(filename)` which simply reads the file and initializes the device accesses.
Or we could add a script (similar to `miio-extract-tokens`) that is shipped separately, but is not bound to the implementations. Just like that script depends on `android_backup`, it is just an optional dependency then.

Would that work for you?

starkillerOG:
@rytilahti alright, lets see.
The goal is to get the EU gateway supported in HomeAsssitant.
I don't particalarly like Dumping and reading into a file because it leaves a unnessesary file on your system.

Would the following be okay:
Have the HomeAssistant integration depend on micloud and python_miio.
Get the device list within HomeAssistant from micloud as a dict.
Supply that dict to miio with a function `get_devices_from_cloud_dict(device_dict)`

rytilahti:
> I don't particalarly like Dumping and reading into a file because it leaves a unnessesary file on your system.

I think it would be great to have an option to synchronize tokens for all cloud activated devices for miiocli (at some point), but see my comment below.

> Have the HomeAssistant integration depend on micloud and python_miio.
> Get the device list within HomeAssistant from micloud as a dict.
> Supply that dict to miio with a function `get_devices_from_cloud_dict(device_dict)`

Yes, I think that would be a good way to do it. Thinking about the big picture here, I think the config flow could also allow synchronizing the tokens for other devices besides just the gateway devices. This is also something that is being already done by several homeassistant integrations (for oauth tokens or other cloud activities).

starkillerOG:
@rytilahti yes for sure, the tokens could indeed also be obtained from the cloud.
Probably we could just use the IP to match or something and then users would not even have to know the tokens anymore, which would be greath for many users.

I will make the changes now, goig to close this PR due to the merge conflicst with the poetry.lock file, will open a new PR in a sec withouth the micloud depandancy

rytilahti:
> Probably we could just use the IP to match or something and then users would not even have to know the tokens anymore, which would be greath for many users.

Yes, indeed, that's what I was thinking! We can always do the regular broadcast discovery for finding the IP addresses in local network, and then use try the given tokens against all discovered devices (e.g., by calling `info()`, where available) to map a token with an IP address. Just an idea, I'm not sure if there's a better way / maybe the cloud will even provide mac address information for matching?

starkillerOG:
Does the info() command not require the token?

With the latest PR for the Config Flow I was already making the step of letting the code do the required matching so users don't even need to know what kind of device they have, only the IP + Token (with the IP beeing automatically discovered by zeroconf), if we then use the cloud to get the token, a user only needs its cloud username and pasword to setup all devices.

starkillerOG:
closing this and continuing here: https://github.com/rytilahti/python-miio/pull/936

rytilahti:
Yes, info will require the token. I don't know how the micloud output looks like, so I'm not sure what's the best approach to do the mapping.

PR #941, opened 2021-02-10
Cleanup: add DeviceStatus to simplify status containers

rytilahti:
* Implements `__repr__` to return the usual pretty repr with all defined properties
* Special cases can override this behavior if really needed.

Prints out currently also properties returning `None`, opinions?
Another thing is that the order of elements in the response is not guaranteed, but no one should not be parsing these anyway, so I think it's a non-issue.

### DIFF
--- a/miio/__init__.py
+++ b/miio/__init__.py
@@ -31,7 +31,7 @@ from miio.chuangmi_ir import ChuangmiIr
 from miio.chuangmi_plug import ChuangmiPlug, Plug, PlugV1, PlugV3
 from miio.cooker import Cooker
 from miio.curtain_youpin import CurtainMiot
-from miio.device import Device
+from miio.device import Device, DeviceStatus
 from miio.dreamevacuum_miot import DreameVacuumMiot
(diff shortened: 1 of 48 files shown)


starkillerOG:
@rytilahti nice cleanup, always good to simplify things!

Would it be possible to use the same logic to add a "self.props" property to each class of python miio that is a dict with the keys beeing the name of the property and the value the value of the property?

That would be extremely helpfull when moving state attributes to seperate sensors in HomeAssistant.

rytilahti:
@starkillerOG yes, it'd be great to improve the status containers even further, especially to simplify the integration you are talking about. I'm not sure what's the best approach here, but instead of using only plain properties in the code, we could allow defining also information like unit of measurement and measurement type, to allow easier mapping against what homeassistant wants (https://developers.home-assistant.io/docs/core/entity/sensor/#available-device-classes).

starkillerOG:
Yea you are right, maybe just merge this and let the properties wait untill we have a simular yaml file as with the subdevices to load the info from.

PR #945, opened 2021-02-11
Replaced typing by pyyaml

legacycode:
When using typing with python 3.8 i can not run tox. I am getting the following error:

`AttributeError: type object 'Callable' has no attribute '_abc_registry'`

According to the documenation on https://pypi.org/project/typing/ typing should only by included in projects using python < 3.5. This pull request fixes my build.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,6 +71,10 @@ stages:
           python.version: '3.8'
           vmImage: 'ubuntu-latest'
 
+        Python 3.9 Ubuntu:
+          python.version: '3.9'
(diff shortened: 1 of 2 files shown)


rytilahti:
LGTM, I'm not seeing that error myself, but typing is not indeed necessary as we don't support 3.5 at all. Could you also please add py39 also to the env list, and pyyaml to the docs environment (looks like something requires it, but I'm not sure what)?

legacycode:
I also added py39 to the azure ci file.

rytilahti:
Great, thanks! :+1:

PR #946, opened 2021-02-12
Fix the logic of staring cleaning a room for Viomi

AlexAlexPin:
Found an issue in `miio/viomivacuum.py start_with_room()`

Lines 633 and 635: the `room_ids` is being appended and passed to vacuum as a string, that the vacuum doesn't understand.

Line 642: According to the description in the method `start()` the params `[0, 1, 3, 11, 12, 13]` = start cleaning rooms 11-13, but the line 642 build the line `[0, 1, 0, 3, 11, 12, 13]`, that means that the length of the room ids array is 0. That makes the vacuum to make a full cleaning instead of cleaning a room.

### DIFF
--- a/miio/viomivacuum.py
+++ b/miio/viomivacuum.py
@@ -630,9 +630,9 @@ class ViomiVacuum(Device):
         room_ids = []
         for room in rooms:
             if room in self._cache["rooms"]:
-                room_ids.append(room)
+                room_ids.append(int(room))
             elif room in reverse_rooms:
-                room_ids.append(reverse_rooms[room])
+                room_ids.append(int(reverse_rooms[room]))
             else:
                 room_keys = ", ".join(self._cache["rooms"].keys())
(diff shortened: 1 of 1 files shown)


rytilahti:
@titilambert this one's for you :-)

rytilahti:
Considering there has been no response from @titilambert so far and I'd like to cut a release soonish, @AlexAlexPin could you confirm that you have tested this PR with a real device and this is okay to merge now?

AlexAlexPin:
> Considering there has been no response from @titilambert so far and I'd like to cut a release soonish, @AlexAlexPin could you confirm that you have tested this PR with a real device and this is okay to merge now?

Yes, I tested it with `viomi.vacuum.v8`

rytilahti:
Okay, thanks for the PR! Let's merge this now and in case it breaks something, it can be fixed later on.

PR #947, opened 2021-02-13
Unify subdevice types

starkillerOG:
The subdevice type was used priviously to match the class needed and was therefore unique per device.
(and it is used in self.__repr__)

I now made the type uniform for all devices which schould behave simular.
That way the type property can be used in HomeAssistant to easily identify what kind of entities schould be added.


starkillerOG:
@rytilahti Could you merge this before the new version gets released?

rytilahti (approved):

PR #957, opened 2021-02-26
fix battery typo (battary) in viomivacuum.py

tom2199:


rytilahti:
Are you sure this is correct? My understanding is that the vacuum uses internally that typoed name, see https://github.com/rytilahti/python-miio/issues/550#issuecomment-552052486 . Ping @titilambert

tom2199:
> Are you sure this is correct? My understanding is that the vacuum uses internally that typoed name, see [#550 (comment)](https://github.com/rytilahti/python-miio/issues/550#issuecomment-552052486) . Ping @titilambert

Thanks for your answer, I tried to fix this typo in Home Assistant, if thats the internal name thats a shame but well nothing we can do about it. Sorry about that.

Issue #584, opened 2019-12-01
Discover uses wrong IP for broadcasts on Windows


xobs:
According to the Valetudo documentation at https://github.com/Hypfer/Valetudo/wiki/Installation-Instructions#flashing-the-firmware-image, the following command should be used to discover the token of an unprovisioned device:

```
> mirobo --debug discover --handshake true
```

However, the program attempts to send a broadcast packet to `<broadcast>`.  This maps to `255.255.255.255`, which does not work on Windows.

Instead, one must use the actual broadcast device.  In this case, it's `192.168.8.255`.  Changing to this causes discovery to work.

See https://grokbase.com/p/python/python-list/032g91nq60/broadcast-socket-bind-on-windows-broadcast-or for discussion of similar issue.

PR #965, opened 2021-03-09
Allow subnet to be specified for broadcast

jorticus:
This is to address issue #584 - discover doesn't work on Windows unless you specify the correct interface's subnet.

### DIFF
--- a/miio/miioprotocol.py
+++ b/miio/miioprotocol.py
@@ -62,7 +62,7 @@ class MiIOProtocol:
         :raises DeviceException: if the device could not be discovered after retries.
         """
         try:
-            m = MiIOProtocol.discover(self.ip)
+            m = MiIOProtocol.discover(self.ip, is_broadcast=False)
         except DeviceException as ex:
             if retry_count > 0:
                 return self.send_handshake(retry_count=retry_count - 1)
@@ -90,7 +90,7 @@ class MiIOProtocol:
         return m
 
     @staticmethod
-    def discover(addr: str = None, timeout: int = 5) -> Any:
+    def discover(addr: str = None, is_broadcast: bool = True, timeout: int = 5) -> Any:
         """Scan for devices in the network. This method is used to discover supported
         devices by sending a handshake message to the broadcast address on port 54321.
         If the target IP address is given, the handshake will be send as an unicast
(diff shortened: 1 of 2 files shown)


miio/miioprotocol.py
```diff
@@ -90,18 +90,18 @@ def send_handshake(self, *, retry_count=3) -> Message:
         return m
 
     @staticmethod
-    def discover(addr: str = None, timeout: int = 5) -> Any:
+    def discover(addr: str = None, is_broadcast: bool = True, timeout: int = 5) -> Any:
```
rytilahti, line 93: Could you also add the new param (and the missing timeout) to the docstring?

miio/miioprotocol.py
```diff
@@ -90,18 +90,18 @@ def send_handshake(self, *, retry_count=3) -> Message:
...
         packet.
 
         :param str addr: Target IP address
         """
-        is_broadcast = addr is None
         seen_addrs = []  # type: List[str]
         if is_broadcast:
-            addr = "<broadcast>"
+            if not addr:
+                addr = "<broadcast>"
```
rytilahti, line 104: I don't really like having nested ifs but I think it's fine until this method gets cleaned up... At the moment it is trying to perform both discovery & handshaking.

miio/miioprotocol.py
jorticus: Done. Hopefully this makes sense

miio/miioprotocol.py
jorticus: Agree, might be nice to split it up, but this seemed like the most minimal change

miio/miioprotocol.py
```diff
@@ -90,18 +90,22 @@ def send_handshake(self, *, retry_count=3) -> Message:
         return m
 
     @staticmethod
-    def discover(addr: str = None, timeout: int = 5) -> Any:
+    def discover(addr: str = None, is_broadcast: bool = True, timeout: int = 5) -> Any:
```
rytilahti, line 93:
```suggestion
    def discover(addr: str = None, timeout: int = 5, *, is_broadcast: bool = True) -> Any:
```
Let's move the `is_broadcast` to be the last parameter, and make it kwarg-only for clarity.

rytilahti (commented):
Sorry for the delay, could you please change the parameter order & fix the linting problem (e.g., by running `pre-commit run -a`)? After those this is ready to be merged.

PR #975, opened 2021-03-17
Add support for Walkingpad A1 (ksmb.walkingpad.v3)

### LINKED ISSUE #797: WalkingPad A1 Pro
**Device information:**

  - Name(s) of the device: Xiaomi WalkingPad A1 Pro
  - Link:

Use `miiocli device --ip <ip address> --token <token>`.

  - Model: ksmb.walkingpad.v1
  - Hardware version: MTK7697
  - Firmware version: 2.0.5

**Additional context**

dewgenenny:
Hey there, I had a go at adding support for the Walkingpad A1. 

What's working:

status, turn on/off, start/stop and set_speed.

What's not working:

Full testing for the status command. The Walkingpad A1 allows retrieval of only 1 property at a time. This means a full status update takes quite a long time to answer. Luckily there is an 'all' property that returns the majority of the status elements. However, this makes testing of the status and speed updates really difficult. (I am sure it's possible but it is outside of my current abilities).

Hope this is good enough to merge... if not be gentle... it's my first ever PR ;) 

Fixes #797

### DIFF
--- a/miio/walkingpad.py
+++ b/miio/walkingpad.py
@@ -69,7 +69,7 @@ class WalkingpadStatus(DeviceStatus):
 
     @property
     def start_speed(self) -> float:
-        """Current speed."""
+        """Current start speed."""
         return self.data["start_speed"]
 
     @property
--- a/miio/walkingpad.py
+++ b/miio/walkingpad.py
@@ -28,10 +28,22 @@ class OperationSensitivity(enum.Enum):
 
 
 class WalkingpadStatus(DeviceStatus):
-    """Container for status reports from Xiaomi Walkingpad."""
+    """Container for status reports from Xiaomi Walkingpad.
+
+    Input data dictionary to initialise this class:
+
+    {'cal': '0',
+     'dist': '0',
+     'mode': 2,
+     'power': 'off',
+     'sensitivity': 1,
+     'sp': '0.0',
+     'start_speed': 3.0,
+     'step': '0',
+     'time': '0'}
+    """
 
     def __init__(self, data: Dict[str, Any]) -> None:
-
         # NOTE: Only 1 property can be requested at the same time
         self.data = data
 
--- a/miio/walkingpad.py
+++ b/miio/walkingpad.py
@@ -1,0 +1,189 @@
+import logging
+from collections import defaultdict
+from typing import Any, Dict
+
+import click
+
+from .click_common import command, format_output
+from .device import Device, DeviceStatus
+from .exceptions import DeviceException
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class WalkingpadException(DeviceException):
+    pass
+
+
+class WalkingpadStatus(DeviceStatus):
+    """Container for status reports from Xiaomi Walkingpad."""
+
+    # raw_command set_start_speed '["1.0"]' finally made my WalkingPad autostart on speed 1! Hurray!
+    #
+    # raw_command get_prop '["mode"]' returns
+    #
+    # [0] while mode is auto
+    #     [1] while mode is manual
+    #     [2] when in standby
+    # raw_command get_prop '["step"]' returns [2303]
+    # raw_command get_prop '["time"]' returns [1970] (while time is 32:50)
+    # raw_command get_prop '["dist"]' returns [1869]
+    # raw_command get_prop '["cal"]' returns [67340]
+    # raw_command get_prop '["goal"]' returns [0, 60]
+    # raw_command get_prop '["max"]' returns [6.0]
+    # raw_command get_prop '["initial"]' returns [3]
+    # raw_command get_prop '["offline"]' returns [0]
+    # raw_command get_prop '["sensitivity"]' returns [2]
+    # raw_command get_prop '["sp"]' returns [1.0]
+    # raw_command get_prop '["start_speed"]' returns [1.0]
+    # raw_command get_prop '["auto"]' returns [1]
+    # raw_command get_prop '["disp"]' returns [19]
+    # raw_command get_prop '["lock"]' returns [0]
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+
+        # NOTE: Only 1 property can be requested at the same time
+        self.data = data
+
+    @property
+    def power(self) -> str:
+        """Power state."""
+        return self.data["power"]
+
+    @property
+    def is_on(self) -> bool:
+        """True if the device is turned on."""
+        return self.power == "on"
+
+    @property
+    def time(self) -> int:
+        """Current lock status."""
+        return self.data["time"]
+
+    @property
+    def speed(self) -> float:
+        """Current speed."""
+        return self.data["sp"]
+
+    @property
+    def mode(self) -> int:
+        """Current mode."""
+        return self.data["mode"]
+
+    @property
+    def step(self) -> int:
+        """Current steps."""
+        return self.data["step"]
+
+    @property
+    def distance(self) -> int:
+        """Current distance."""
+        return self.data["dist"]
+
+    @property
+    def calories(self) -> int:
+        """Current distance."""
+        return self.data["cal"]
+
+
+class Walkingpad(Device):
+    """Main class representing Xiaomi Walkingpad."""
+
+    # TODO: - Auto On/Off Not Supported
+    #       - Adjust Scenes with Wall Switch Not Supported
+
+    @command(
+        default_output=format_output(
+            "",
+            "Mode: {result.mode}\n"
+            "Time: {result.time}\n"
+            "Steps: {result.step}\n"
+            "Speed: {result.speed}\n"
+            "Distance: {result.distance}\n"
+            "Calories: {result.calories}  ",
+        )
+    )
+    def status(self) -> WalkingpadStatus:
+        """Retrieve properties."""
+
+        properties_received = []
+        values_received = []
+
+        # Walkingpad A1 allows you to retrieve a subset of values with "all"
+        # eg ['mode:1', 'time:1387', 'sp:3.0', 'dist:1150', 'cal:71710', 'step:2117']
+
+        properties = ["all"]
+
+        # Walkingpad A1 only allows 1 property to be read at a time
+
+        values = self.get_properties(properties, max_properties=1)
+
+        # When running the tests, for some reason the list provided is passed within another list, so I take
+        # care of this here.
+
+        if any(isinstance(el, list) for el in values):
+            values = values[0]
+
+        for x in values:
+            prop_split, value = x.split(":")
+            properties_received.append(prop_split)
+            values_received.append(value)
+
+        properties_additional = ["power", "mode"]
+
+        values_additional = self.get_properties(properties_additional, max_properties=1)
+
+        properties_received = properties_received + properties_additional
+        values_received = values_received + values_additional
+        return WalkingpadStatus(
+            defaultdict(lambda: None, zip(properties_received, values_received))
+        )
+
+    @command(default_output=format_output("Powering on"))
+    def on(self):
+        """Power on."""
+        return self.send("set_power", ["on"])
+
+    @command(default_output=format_output("Powering off"))
+    def off(self):
+        """Power off."""
+        return self.send("set_power", ["off"])
+
+    @command(default_output=format_output("Starting the treadmill"))
+    def start(self):
+        """Starting Up."""
+        return self.send("set_state", ["run"])
+
+    @command(default_output=format_output("Stopping the treadmill"))
+    def stop(self):
+        """Starting Up."""
+        return self.send("set_state", ["stop"])
+
+    @command(
+        click.argument("mode", type=int),
+        default_output=format_output("Setting mode to {mode}"),
+    )
+    def set_mode(self, mode: int):
+        """Set mode (auto/manual)."""
+
+        if not isinstance(mode, int):
+            raise WalkingpadException("Invalid mode: %s" % mode)
+
+        elif mode < 0 or mode >= 3:
+            raise WalkingpadException("Invalid mode: %s" % mode)
+        return self.send("set_mode", [mode])
+
+    @command(
+        click.argument("speed", type=float),
+        default_output=format_output("Setting speed to {speed}"),
+    )
+    def set_speed(self, speed: float):
+        """Set speed."""
+
+        if not isinstance(speed, float):
+            raise WalkingpadException("Invalid speed: %s" % speed)
+
+        elif speed < 0 or speed > 6:
+            raise WalkingpadException("Invalid speed: %s" % speed)
+
+        return self.send("set_speed", [speed])
--- a/miio/walkingpad.py
+++ b/miio/walkingpad.py
@@ -179,7 +179,7 @@ class Walkingpad(Device):
 
     @command(default_output=format_output("Starting the treadmill"))
     def start(self):
-        """Starting Up."""
+        """Start the treadmill."""
         if not self.status().is_on:
             raise WalkingpadException(
                 "Can't start the treadmill, it's not turned on - try issuing an 'on' command first"
@@ -189,7 +189,7 @@ class Walkingpad(Device):
 
     @command(default_output=format_output("Stopping the treadmill"))
     def stop(self):
-        """Starting Up."""
+        """Stop the treadmill."""
         return self.send("set_state", ["stop"])
 
     @command(
--- a/miio/walkingpad.py
+++ b/miio/walkingpad.py
@@ -29,7 +29,7 @@ class OperationSensitivity(enum.Enum):
 
 
 class WalkingpadStatus(DeviceStatus):
-    """Container for status reports from Xiaomi Walkingpad.
+    """Container for status reports from Xiaomi Walkingpad A1 (ksmb.walkingpad.v3).
 
     Input data dictionary to initialise this class:
 
@@ -75,12 +75,12 @@ class WalkingpadStatus(DeviceStatus):
     @property
     def mode(self) -> OperationMode:
         """Current mode."""
-        return self.data["mode"]
+        return OperationMode(self.data["mode"])
 
     @property
     def sensitivity(self) -> OperationSensitivity:
         """Current sensitivity."""
-        return self.data["sensitivity"]
+        return OperationSensitivity(self.data["sensitivity"])
 
     @property
     def step_count(self) -> int:
@@ -105,12 +105,12 @@ class Walkingpad(Device):
         default_output=format_output(
             "",
             "Power: {result.power}\n"
-            "Mode: {result.mode}\n"
-            "Time: {result.time}\n"
+            "Mode: {result.mode.name}\n"
+            "Time: {result.walking_time}\n"
             "Steps: {result.step_count}\n"
             "Speed: {result.speed}\n"
             "Start Speed: {result.start_speed}\n"
-            "Sensitivity: {result.sensitivity}\n"
+            "Sensitivity: {result.sensitivity.name}\n"
             "Distance: {result.distance}\n"
             "Calories: {result.calories}",
         )
@@ -132,8 +132,8 @@ class Walkingpad(Device):
     @command(
         default_output=format_output(
             "",
-            "Mode: {result.mode}\n"
-            "Time: {result.time}\n"
+            "Mode: {result.mode.name}\n"
+            "Time: {result.walking_time}\n"
             "Steps: {result.step_count}\n"
             "Speed: {result.speed}\n"
             "Distance: {result.distance}\n"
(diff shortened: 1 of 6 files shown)


miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+    @command(default_output=format_output("Stopping the treadmill"))
+    def stop(self):
+        """Starting Up."""
+        return self.send("set_state", ["stop"])
+
+    @command(
+        click.argument("mode", type=int),
+        default_output=format_output("Setting mode to {mode}"),
+    )
+    def set_mode(self, mode: int):
```
rytilahti, line 166: Please use an enum for the mode setting.

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Walkingpad A1 only allows 1 property to be read at a time
+
+        values = self.get_properties(properties, max_properties=1)
+
+        # When running the tests, for some reason the list provided is passed within another list, so I take
+        # care of this here.
+
+        if any(isinstance(el, list) for el in values):
+            values = values[0]
```
rytilahti, line 125: You could do a check like this to make sure that a non-empty list is returned:
```
if not values:
    raise ...
```

and  then simply do `values = values.pop()` afterwards.

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+
+        for x in values:
+            prop_split, value = x.split(":")
+            properties_received.append(prop_split)
+            values_received.append(value)
```
rytilahti, line 130: Instead of using two lists, maybe store it all inside a dictionary?

```
data = {}
for x in values:
    prop, val = x.split(":")
    data[prop] = val
```

which can then be passed directly to the status container?

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        return self.data["power"]
+
+    @property
+    def is_on(self) -> bool:
+        """True if the device is turned on."""
+        return self.power == "on"
+
+    @property
+    def time(self) -> int:
+        """Current lock status."""
```
rytilahti, line 60: Misplaced docstring? Is this `run_time` or something else?

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        """Current speed."""
+        return self.data["sp"]
+
+    @property
+    def mode(self) -> int:
+        """Current mode."""
+        return self.data["mode"]
+
+    @property
+    def step(self) -> int:
```
rytilahti, line 74:
```suggestion
    def step_count(self) -> int:
```
Instead of using directly the names provided in the payloads, it makes sense to rename them to be comprehensible. Here one example :-)

rytilahti (changes requested):
Thanks for the PR! I did a cursory code review, but it is looking quite good already :-)

dewgenenny:
Thanks man for the really quick response and review! Will make the changes later today hopefully :)

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        click.argument("mode", type=int),
+        default_output=format_output("Setting mode to {mode}"),
+    )
+    def set_mode(self, mode: int):
+        """Set mode (auto/manual)."""
+
+        if not isinstance(mode, int):
+            raise WalkingpadException("Invalid mode: %s" % mode)
+
+        elif mode < 0 or mode >= 3:
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+            raise WalkingpadException("Invalid mode: %s" % mode)
+        return self.send("set_mode", [mode])
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        click.argument("speed", type=float),
+        default_output=format_output("Setting speed to {speed}"),
+    )
+    def set_speed(self, speed: float):
+        """Set speed."""
+
+        if not isinstance(speed, float):
+            raise WalkingpadException("Invalid speed: %s" % speed)
+
+        elif speed < 0 or speed > 6:
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
dewgenenny: Done

miio/walkingpad.py
dewgenenny: Done

miio/walkingpad.py
dewgenenny: Done

miio/walkingpad.py
dewgenenny: Done

miio/walkingpad.py
dewgenenny: This one I am not sure I fully understood your proposal, but I have worked to simplify / improve:

        if len(values) <= 1:
            values = values.pop()

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+class OperationMode(enum.Enum):
+    Auto = 0
+    Manual = 1
+    Off = 2
+
+
+class WalkingpadStatus(DeviceStatus):
+    """Container for status reports from Xiaomi Walkingpad."""
+
+    # raw_command set_start_speed '["1.0"]' finally made my WalkingPad autostart on speed 1! Hurray!
```
rytilahti, line 28: This is the only mention on `set_start_speed`, are you planning to implement this in the device class?

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+    # raw_command get_prop '["goal"]' returns [0, 60]
+    # raw_command get_prop '["max"]' returns [6.0]
+    # raw_command get_prop '["initial"]' returns [3]
+    # raw_command get_prop '["offline"]' returns [0]
+    # raw_command get_prop '["sensitivity"]' returns [2]
+    # raw_command get_prop '["sp"]' returns [1.0]
+    # raw_command get_prop '["start_speed"]' returns [1.0]
+    # raw_command get_prop '["auto"]' returns [1]
+    # raw_command get_prop '["disp"]' returns [19]
+    # raw_command get_prop '["lock"]' returns [0]
```
rytilahti, line 48: Instead of listing these out here, better handle each specific case in their getter.

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class OperationMode(enum.Enum):
+    Auto = 0
+    Manual = 1
+    Off = 2
+
+
+class WalkingpadStatus(DeviceStatus):
+    """Container for status reports from Xiaomi Walkingpad."""
```
rytilahti, line 26: Please add an example of the input data dictionary which is used to initialize this class. The simplest way is likely to add a temporary `from pprint import pprint; pprint(self.data)` line inside `__init__` and copy&paste it to the docstring.

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        """Power state."""
+        return self.data["power"]
+
+    @property
+    def is_on(self) -> bool:
+        """True if the device is turned on."""
+        return self.power == "on"
+
+    @property
+    def time(self) -> int:
```
rytilahti, line 66:
```suggestion
    def walking_time(self) -> timedelta:
```
Better convert the time into a timedelta object to be more consistent with other integrations.

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+    # raw_command get_prop '["sensitivity"]' returns [2]
+    # raw_command get_prop '["sp"]' returns [1.0]
+    # raw_command get_prop '["start_speed"]' returns [1.0]
+    # raw_command get_prop '["auto"]' returns [1]
+    # raw_command get_prop '["disp"]' returns [19]
+    # raw_command get_prop '["lock"]' returns [0]
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+
+        # NOTE: Only 1 property can be requested at the same time
```
rytilahti, line 52:
```suggestion
```

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @property
+    def speed(self) -> float:
+        """Current speed."""
+        return self.data["sp"]
+
+    @property
+    def mode(self) -> int:
+        """Current mode."""
+        return self.data["mode"]
```
rytilahti, line 78: Should this return an `OperationMode` enum?

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @property
+    def step_count(self) -> int:
+        """Current steps."""
+        return self.data["step"]
+
+    @property
+    def distance(self) -> int:
+        """Current distance."""
```
rytilahti, line 87: Add unit to the docstring, if known.

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @command(
+        default_output=format_output(
+            "",
+            "Mode: {result.mode}\n"
+            "Time: {result.time}\n"
+            "Steps: {result.step}\n"
+            "Speed: {result.speed}\n"
+            "Distance: {result.distance}\n"
+            "Calories: {result.calories}  ",
```
rytilahti, line 110:
```suggestion
            "Calories: {result.calories}",
```

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+    )
+    def status(self) -> WalkingpadStatus:
+        """Retrieve properties."""
+
+        # Walkingpad A1 allows you to retrieve a subset of values with "all"
+        # eg ['mode:1', 'time:1387', 'sp:3.0', 'dist:1150', 'cal:71710', 'step:2117']
+
```
rytilahti, line 118:
```suggestion
```

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        properties = ["all"]
+
+        # Walkingpad A1 only allows 1 property to be read at a time
+
```
rytilahti, line 122:
```suggestion
```

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        data = {}
+        for x in values:
+            prop, value = x.split(":")
+            data[prop] = value
+
+        properties_additional = ["power", "mode"]
+        values_additional = self.get_properties(properties_additional, max_properties=1)
+
+        for i in range(len(properties_additional)):
+            data[properties_additional[i]] = values_additional[i]
```
rytilahti, line 140:
```suggestion
additional_props = dict(zip(properties_additional, values_additional))
data.update(additional_props)
```
would be more pythonic.

miio/walkingpad.py
rytilahti: What I meant with that is you can use an if clause to see if the variable is not None or an empty list, in which case you can simply do the pop to access its first element.

You mention in the comment, that it happens during the test. Do you mean during unit tests or in the data reported by the device?

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        return self.data["step"]
+
+    @property
+    def distance(self) -> int:
+        """Current distance."""
+        return self.data["dist"]
+
+    @property
+    def calories(self) -> int:
+        """Current distance."""
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
```diff
@@ -0,0 +1,189 @@
...
+        return self.data["cal"]
+
+
+class Walkingpad(Device):
+    """Main class representing Xiaomi Walkingpad."""
+
+    # TODO: - Auto On/Off Not Supported
+    #       - Adjust Scenes with Wall Switch Not Supported
```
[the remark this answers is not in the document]
dewgenenny: Eep these comments were left over from the code I copied to start with. Removed.

miio/walkingpad.py
dewgenenny: Implemented :)

miio/walkingpad.py
dewgenenny: Done

miio/walkingpad.py
dewgenenny: Done

miio/walkingpad.py
dewgenenny: Nice! I have changed to this 👍

miio/walkingpad.py
dewgenenny: OK, thanks for the tips. In the end I was solving the wrong problem. I needed to create different behaviour for the dummy device to replicate how the physical device behaves when queried with the 'all' parameter. 

I did some research and figured out that I needed to override the dummy class in the test, which I have done to deal with this case. Now the code doesn't contain a weird workaround for the test, which I guess is much better :)

dewgenenny:
Hey @rytilahti just pushed a new version. Contains all your requested changes and hopefully is closer to something that makes sense to merge :p 

Not sure why the python windows build is failing, seems perhaps to be related to the pipeline instead of something specific with my code?

dewgenenny:
not sure if I should have 'requested new review' - TOTALLY didnt want to be pushy, just not sure what the normal flow is ;) 

appreciate any and all time and help man!

miio/walkingpad.py
```diff
@@ -0,0 +1,249 @@
...
+
+    {'cal': '0',
+     'dist': '0',
+     'mode': 2,
+     'power': 'off',
+     'sensitivity': 1,
+     'sp': '0.0',
+     'start_speed': 3.0,
+     'step': '0',
+     'time': '0'}
```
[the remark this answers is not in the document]
dewgenenny: sorry to ask a potentially stupid question, but when you say cast the data in the getters... you mean like this: 

    @property
    def speed(self) -> float:
        """Current speed."""
        return float(self.data["sp"])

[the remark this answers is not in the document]
dewgenenny: OK cool, implemented. I also needed to take care of the types when receiving them from the 'quick status' call, so now the internal quick status function specifies the data type for each value.

miio/walkingpad.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        # NOTE: Only 1 property can be requested at the same time
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # Walkingpad A1 only allows 1 property to be read at a time
+        values = self.get_properties(properties, max_properties=1)
+
+        data = {}
+        for x in values:
+            prop, value = x.split(":")
+            data[prop] = value
+
+        properties_additional = ["power", "mode", "start_speed", "sensitivity"]
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
```diff
@@ -0,0 +1,249 @@
...
+
+        properties = ["all"]
+
+        # Walkingpad A1 only allows 1 property to be read at a time
+        values = self.get_properties(properties, max_properties=1)
+        # print(values)
+        data = {}
+        for x in values:
+            prop, value = x.split(":")
+            data[prop] = value
```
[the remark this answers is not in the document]
dewgenenny: Done - broke it out into _get_quick_status

dewgenenny:
OK man! Just finished updating as per your comments. I am doing this all while walking on the treadmill, so I am working fast so I can sit down soon 😆

miio/walkingpad.py
```diff
@@ -0,0 +1,280 @@
...
+            elif prop == "cal":
+                value = int(value)
+            elif prop == "time":
+                value = int(value)
+            elif prop == "dist":
+                value = int(value)
+            elif prop == "mode":
+                value = int(value)
+
+            data[prop] = value
```
rytilahti, line 278: Something like this would be more readable:
```
# prepare data dict
data = ...

value_map = {
    "time": int,
    "mode": int,
    "sp": float,
    ...
}

converted_data = {key: type_map[key](value) for key, value in data.items()}
```

miio/walkingpad.py
```diff
@@ -0,0 +1,280 @@
...
+        return int(self.data["dist"])
+
+    @property
+    def calories(self) -> int:
+        """Current calories burnt."""
+        return int(self.data["cal"])
+
+
+class Walkingpad(Device):
+    """Main class representing Xiaomi Walkingpad."""
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
```diff
@@ -0,0 +1,280 @@
...
+
+    @property
+    def start_speed(self) -> float:
+        """Current start speed."""
+        return self.data["start_speed"]
+
+    @property
+    def mode(self) -> OperationMode:
+        """Current mode."""
+        return self.data["mode"]
```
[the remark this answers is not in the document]
dewgenenny: Done & tests updated.. sorry somehow I missed before

miio/walkingpad.py
```diff
@@ -0,0 +1,280 @@
...
+
+    @property
+    def sensitivity(self) -> OperationSensitivity:
+        """Current sensitivity."""
+        return self.data["sensitivity"]
```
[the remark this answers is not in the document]
dewgenenny: Done & tests updated.. sorry somehow I missed before

miio/walkingpad.py
```diff
@@ -0,0 +1,280 @@
...
+        """Set sensitivity."""
+
+        if not isinstance(sensitivity, OperationSensitivity):
+            raise WalkingpadException("Invalid mode: %s" % sensitivity)
+
+        return self.send("set_sensitivity", [sensitivity.value])
+
+    def _get_quick_status(self):
+
+        # internal helper to get the quick status via the "all" property
```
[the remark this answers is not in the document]
dewgenenny: Done

miio/walkingpad.py
dewgenenny: Same here! Really appreciate the feedback and help... learning a lot!

dewgenenny:
Next round done :) thanks for all your patience and guidance man! Some of the checks in the build pipeline were not successful, but I think this is more to do with pipeline stability thank something specific in my code. I would hit re-run but don't have the necessary permissions

miio/walkingpad.py
```diff
@@ -0,0 +1,275 @@
...
+            "Calories: {result.calories}",
+        )
+    )
+    def quick_status(self) -> WalkingpadStatus:
+        """Retrieve quick properties. The walkingpad provides the option.
+
+        to retrieve a subset of properties in one call - steps, mode,
+        speed, distance, calories and time. If you can, use this instead
+        of the full status request, as it is way faster.
+        """
```
rytilahti, line 149:
```suggestion
        """Retrieve quick status. 
        
        The walkingpad provides the option to retrieve a subset of properties in one call:
        steps, mode, speed, distance, calories and time.
        
        `status()` will do four more separate I/O requests for power, mode, start_speed, and sensitivity.
        If you don't need any of that, prefer this method for status updates.
        """
```
Just a suggestion, but the 1st line of docstring should be in any case just "Retrieve quick status" or so without that part from the more elaborate description..

rytilahti (changes requested):
I restarted the CI and looks like everything is fine on that front. Here just a couple minor nitpicks and this is good to go, thanks for your patience!

miio/walkingpad.py
dewgenenny: Done, good call :)

miio/walkingpad.py
```diff
@@ -0,0 +1,275 @@
...
+            "dist": int,
+            "mode": int,
+        }
+
+        data = {}
+        for x in values:
+            prop, value = x.split(":")
+
+            if prop not in value_map:
+                raise WalkingpadException("Invalid data received: %s" % value)
```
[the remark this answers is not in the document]
dewgenenny: agree this is a much nicer way. In the case that we get additional infos provided through a future firmware, this will be a really nice pointer :)

miio/walkingpad.py
```diff
@@ -0,0 +1,275 @@
...
+        """Unlock device."""
+        return self.send("set_lock", [0])
+
+    @command(default_output=format_output("Starting the treadmill"))
+    def start(self):
+        """Start the treadmill."""
+        if not self.status().is_on:
+            raise WalkingpadException(
+                "Can't start the treadmill, it's not turned on - try issuing an 'on' command first"
+            )
```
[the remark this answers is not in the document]
dewgenenny: I thought about this initially, but didn't implement because I wasn't sure if it's nice to implement these types of logics into the library. Regardless, changed to include it now - agree it's a nicer way to behave

dewgenenny:
Added the changes you requested.... really really appreciate the time and patience you have taken with me to get this done!

rytilahti:
Looks good to go, thanks @dewgenenny! :tada:

rytilahti (approved):

PR #981, opened 2021-03-21
Move hardcoded parameter `max_properties`

ha0y:
Not all devices limit max_properties to 15. We could make this parameter customizable.

### DIFF
--- a/miio/miot_device.py
+++ b/miio/miot_device.py
@@ -30,7 +30,7 @@ class MiotDevice(Device):
 
     mapping = None
 
-    def get_properties_for_mapping(self) -> list:
+    def get_properties_for_mapping(self, max_properties=15) -> list:
         """Retrieve raw properties based on mapping."""
 
         # We send property key in "did" because it's sent back via response and we can identify the property.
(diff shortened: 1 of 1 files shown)


syssi (approved):

rytilahti (commented):
I'd personally rather like to see this as a class variable like the mapping itself, considering there should be no reason to change the value on the fly for subsequent requests on the same instance.

ha0y:
Yes. This is a temporary workaround.

I used to use this method to determine whether the device is miot or miio. Both of them has miot-spec, but only miot devices works with it in LAN. I have thought a return code -9999 indicates that they are not miot, but recently found that some devices limit max_properties to a lower value and also returns -9999 when exceeded.

miio/miot_device.py
```diff
@@ -30,7 +30,7 @@ class MiotDevice(Device):
 
     mapping = None
 
-    def get_properties_for_mapping(self) -> list:
+    def get_properties_for_mapping(self, max_properties=15) -> list:
```
rytilahti, line 33:
```suggestion
    def get_properties_for_mapping(self, *, max_properties=15) -> list:
```
let's make this kwarg only to make the code more readable (even when the name is also a bit off..)?

rytilahti (approved):
Great, thanks for the PR! :+1:

PR #990, opened 2021-03-25
Reformat history data if returned as a dict/Roborock S7 Support (#989)

fettlaus:
This should fix #989 by taking care of the new history format, the Roborock S7 is using. I also added some tests just in case. Couldn't find anything else not working with this vacuum so maybe that was all it takes for it to be supported.

I would've used some pull request template but it seems there isn't any so there is that.

PR #1006, opened 2021-04-08
Roborock S7: Parse history details returned as dict

### LINKED ISSUE #1004: Roborock S7 after Firmware Update 4.1.2-0928 - KeyError
**Describe the bug**
Error after Firmware Update.

Status, consumable_status, info and other commands are running.

Rebooting the device does not clear the issue.

**Version information (please complete the following information):**
 - OS: Linux Debian
 - python-miio: [Use `miiocli --version` or `pip show python-miio`]

**Device information:**
If the issue is specific to a device [Use `miiocli device --ip <ip address> --token <token> info`]:
  - Model: roborock.vacuum.a15
  - Hardware version: Linux
  - Firmware version: 4.1.2_0928

**To Reproduce**
1.

**Expected behavior**

**Console output**
miiocli vacuum --ip <my_ip> --token <my_token> clean_history    Running command clean_history
<CleaningSummary count=KeyError ids=KeyError total_area=KeyError total_duration=KeyError>

INFO:miio.cli:Debug mode active
DEBUG:miio.vacuum:Read stored sequence ids: {'seq': 36, 'manual_seq': 0}
Running command consumable_status
DEBUG:miio.protocol:Unable to decrypt, returning raw bytes: b''
DEBUG:miio.miioprotocol:Got a response: Container:
    data = Container:
        data = b'' (total 0)
        value = b'' (total 0)
        offset1 = 32
        offset2 = 32
        length = 0
    header = Container:
        data = b'!1\x00 \x00\x00\x00\x00\x17P\xc6U\x00\x00\x0f%' (total 16)
        value = Container:
            length = 32
            unknown = 0
            device_id = unhexlify('1750c655')
            ts = 1970-01-01 01:04:37
        offset1 = 0
        offset2 = 16
        length = 16
    checksum = b'\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff' (total 16)
DEBUG:miio.miioprotocol:Discovered 1750c655 with ts: 1970-01-01 01:04:37, token: b'ffffffffffffffffffffffffffffffff'
DEBUG:miio.miioprotocol:<my_ip>:54321 >>: {'id': 37, 'method': 'get_consumable', 'params': []}
DEBUG:miio.miioprotocol:<my_ip>:54321 (ts: 1970-01-01 01:04:38, id: 37) << {'id': 37, 'result': [{'main_brush_work_time': 37961, 'side_brush_work_time': 37961, 'filter_work_time': 37961, 'filter_element_work_time': 0, 'sensor_dirty_time': 37504}], 'exe_time': 101}
<ConsumableStatus filter=10:32:41 filter_left=5 days, 19:27:19 main_brush=10:32:41 main_brush_left=12 days, 1:27:19 sensor_dirty=10:25:04 sensor_dirty_left=19:34:56 side_brush=10:32:41 side_brush_left=7 days, 21:27:19>

fettlaus:
I noticed, that history details as of FW 4.1.2_0928 for this vacuum also come formatted as a dict (similar behavior see #990).

This PR also tries to fix #1004 (robot has an empty cleaning record)

### DIFF
--- a/miio/vacuumcontainers.py
+++ b/miio/vacuumcontainers.py
@@ -197,8 +197,9 @@ class CleaningSummary(DeviceStatus):
                 "clean_time": data[0],
                 "clean_area": data[1],
                 "clean_count": data[2],
-                "records": data[3],
             }
+            if len(data) > 3:
+                self.data["records"] = data[3]
         else:
             self.data = data
 
@@ -220,7 +221,10 @@ class CleaningSummary(DeviceStatus):
     @property
     def ids(self) -> List[int]:
         """A list of available cleaning IDs, see also :class:`CleaningDetails`."""
-        return list(self.data["records"])
+        if "records" in self.data:
+            return list(self.data["records"])
+        else:
+            return []
 
     @property
     def dust_collection_count(self) -> Optional[int]:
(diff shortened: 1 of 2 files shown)


bbr111:
I tested the PR and it works for me.

miio/vacuumcontainers.py
```diff
@@ -197,8 +197,9 @@ def __init__(self, data: Union[List[Any], Dict[str, Any]]) -> None:
                 "clean_time": data[0],
                 "clean_area": data[1],
                 "clean_count": data[2],
-                "records": data[3],
             }
+            if len(data) > 3:
+                self.data["records"] = data[3]
```
rytilahti, line 202:
```suggestion
            self.data["records"] = data[3] if len(data) > 3 else []
```
This would also allow removing the records check below and making the getter simpler. Otherwise this PR looks good to me.

fettlaus: Not wanting to discuss styling decisions but the ternary operator is not very pythonic in my opinion. 

And in what way does it make the record check below obsolete? We still need to catch the case in which the new dict-style parameter has no records coming from the device. Maybe it would be more concise to move the check from the getter up here.

rytilahti: Ah, sorry, you're right. I was not thinking about the dict input, just about keeping the getter clean... I'd say it's your call, and I agree with the notion that ternaries are less readable than proper if-elses.

rytilahti:
Looks great, thanks! :+1:

Issue #958, opened 2021-02-27
No air quality value when aqi is 1


exelban:
Hi.
I'm trying to get air quality from my Xiaomi Air purifier 3C. 
And from time to time I see None value in the `aqi`. I figured out that it's happening when `aqi==1`.

The reason is in this if: https://github.com/rytilahti/python-miio/blob/dfda13edb8bcfd844b2e1f2d7d7374a9a0709ac6/miio/airpurifier_miot.py#L105-L112
PR #930

Does `zhimi.airpurifier.mb4` (and any non `zhimi-airpurifier-mb3`) return 1 when the air quality read was unsuccessful?

Mi Home shows `pm2.5==1`. So I suppose for my model `zhimi.airpurifier.mb4` 1 it's a normal read.

**Device information:**
  - Model: Xiaomi Air purifier 3C (`zhimi.airpurifier.mb4`)

rytilahti:
If you are using homeassistant, check out the historical graph to see if it makes sense that it would be 1. Otherwise, save the data over some time period on see if the values make sense. In the screenshot in #925 you can clearly see that the 1 is not a real measurement.

exelban:
Thanks for the response. 
I'm not using homeassistant. I will try to collect the data and share the results. 
I spend a few days playing with `python-miio` and this purifier. And pretty sure that 1 is a normal read for this model.

nefsolive:
@exelban i see the same, 1 ppm! You are using domoticz?
Sometimes i see 0 ppm and  1ppm! 
Did you still have this values? Im using 3C too.

exelban:
Hi. Nope, I do not use any solution (HA or domoticz). I have my own scripts.
I save data from last week in the influx. And there is a chart of `aqi`.
![Zrzut ekranu 2021-03-30 o 20 29 08](https://user-images.githubusercontent.com/13332412/113038289-a6883b80-9196-11eb-9867-e15f290438bb.png)

Truly, It does not look like 1 is an unsuccessful read. Because the chart looks quite normal. There were a few drops to `1`. But also a few to `3` or `4`. PS: read every 30 min.

I will increase sampling to 1 min for next week. And share the result after.

nefsolive:
Thank u for share the results!
On domoticz i just see the value 1ppm on graph, but it as soon for a test, i just install plugin about 5 days ago and few houres day take ON! 
Maybe its normal 1, but i need more time for make sure!

rytilahti:
[...]

PR #1007, opened 2021-04-09
Always return aqi value [Revert PR#930]

bieniu:
This PR reverts change from https://github.com/rytilahti/python-miio/pull/930 because it has caused more problems than good.

Fixes: https://github.com/rytilahti/python-miio/issues/958


syssi (approved):

rytilahti:
Indeed, thanks!

PR #1016, opened 2021-04-12
gateway: fix zigbee lights

starkillerOG:
fix zigbee lights, for more info see https://github.com/home-assistant/core/issues/48988

### DIFF
--- a/miio/gateway/devices/light.py
+++ b/miio/gateway/devices/light.py
@@ -10,16 +10,6 @@ class LightBulb(SubDevice):
     """Base class for subdevice light bulbs."""
 
     @command()
-    def update(self):
-        """Update all device properties."""
-        self._props["brightness"] = self.send("get_bright").pop()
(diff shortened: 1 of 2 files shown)


rytilahti (approved):
I presume these never worked before, right? Or could this be a breaking change for some existing lights?

Other than that, looks good to me!

starkillerOG:
@rytilahti no they probably have newer worked.
I added them, but I don't own them, so I could not test.

Together with a user we have now tested it and now we know which commands work.
He has not tested the exact code of this PR, but the commands schould work.

rytilahti:
Okay, so this can be merged now, or do you want to wait for a confirmation from the user?

starkillerOG:
@rytilahti @xekil already tested it and has confirmed this PR works.
This can now be merged.

xekil:
any idea when the HA update is available that includes this fix?

PR #1017, opened 2021-04-13
Fix start bug and improve error handling in walkingpad integration

dewgenenny:
Hey, a bug fix for the start command, when the device was turned off. Additionally corrected the walking_time return, as it should have been a timedelta and it was returning an int. 

Also added some better error handling for functions to deal with when the device is turned off.

### DIFF
--- a/miio/walkingpad.py
+++ b/miio/walkingpad.py
@@ -41,7 +41,7 @@ class WalkingpadStatus(DeviceStatus):
      'sp': 3.0,
      'start_speed': 3.0,
      'step': 180,
-     'walking_time': 121}
+     'time': 121}
     """
 
     def __init__(self, data: Dict[str, Any]) -> None:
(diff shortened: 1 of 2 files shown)


dewgenenny:
I am guessing the failed build is pipeline instability again :/

miio/walkingpad.py
```diff
@@ -41,7 +41,7 @@ class WalkingpadStatus(DeviceStatus):
      'sp': 3.0,
      'start_speed': 3.0,
      'step': 180,
-     'time': 121}
+     'walking_time': 121}
```
rytilahti, line 44: The code below uses `self.data["time"]` for `walking_time`. Which is the correct input?

rytilahti (commented):
The windows builds on azure pipelines have been flaky for no known reason, so I restarted the build manually.

miio/walkingpad.py
dewgenenny: good catch, you're right it should stay as time. I have corrected.

dewgenenny:
Should be good to go now

rytilahti:
Great, thanks! :+1:

PR #1018, opened 2021-04-14
Initial support for chunmi.cooker.eh1

sschirr:
This provides basic support for the Xiaomi Mi Smart MultiCooker.

This is supported:

- Start/Set a recipe profile
  - With support for duration, schedule and autokeepwarm (akw)
- Get some status information including:
  - operation mode
  - menu
  - cooking stage (incomplete)
  - temperature + history
  - cooking times

This is not supported:

- any other change to  cooking profiles except: duration, schedule, akw
  - like: rice type, taste setting, temperature curve etc...
- cooker settings
- factory reset
- firmware version
- ...


syssi:
Your implementation is pretty advanced especially the custom cooking profiles. Could you provide some tests or test data (of different cooking stages) to be able to implement some tests later on?

FaserF:
I am using the fork from sschirr and can say that everything works fine so far with my chunmi.cooker.eh1

blakadder:
I'm also using the forked version, cooker works without issues

sschirr:
Is there any change to get this merged?
Its been a long time since I implemented it and I am happy to solve the branch conflicts if there is still interest.

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
+import enum
```
rytilahti, line 1: Please move this and all other integration files (including tests) under `miio/integrations/chunmi/cooker_multi` like other integrations currently have it.

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+    "0000000000000000000000000000000000000001": "Fine Rice",
+    "0101000000000000000000000000000000000002": "Quick Rice",
+    "0202000000000000000000000000000000000003": "Congee",
+    "0303000000000000000000000000000000000004": "Keep warm",
+}
+
+
+class CookerException(DeviceException):
+    pass
+
```
rytilahti, line 67:
```suggestion
```
No need for custom exceptions anymore, just use the standard library ones (e.g., `ValueError` for invalid inputs) wherever expected and otherwise raise `DeviceException`. Similarly to what was done in #1558

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+class OperationMode(enum.Enum):
+    Waiting = 1
+    Running = 2
+    AutoKeepWarm = 3
+    PreCook = 4
+
+    Unknown = "unknown"
+
+    @classmethod
+    def _missing_(cls, _):
```
rytilahti, line 78: I don't think this works, should be `__missing__`?

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+
+
+class CookerStatus(DeviceStatus):
+    def __init__(self, data):
+        self.data = data
+
+    @property
+    def mode(self) -> OperationMode:
+        """Current operation mode."""
+        return OperationMode(self.data["status"])
```
rytilahti, line 233: Please add the decorators for entries that should be exposed per default to cli and homeassistant, see https://python-miio.readthedocs.io/en/latest/contributing.html#status-containers or how it's done in yeelight or roborock integrations.

Basically, everything that is properly decorated will be automatically printed out when calling status() and are also exposed to the [next-gen xiaomi_miio integration](https://github.com/rytilahti/homeassistant-xiaomi-ng) whenever it gets ready.

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+        """Current temperature, if idle.
+
+        Example values: 29
+        """
+        return self.data["temp"]
+
+    @property
+    def start_time(self) -> int:
+        """Start time of cooking?"""
+        return int(self.data["t_start"])
```
rytilahti, line 262: If this is a timestamp, it'd be better to expose this as `datetime`?

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+
+    @property
+    def remaining(self) -> int:
```
rytilahti, line 265: And this should also likely be `ready_at(self) -> datetime` or similar to avoid spamming the homeassistant state machine for updates.

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+
+    @property
+    def hardware_version(self) -> None:
+        """Hardware version."""
+        return None
+
+    @property
+    def firmware_version(self) -> None:
+        """Firmware version."""
+        return None
```
rytilahti, line 302: I'd remove these if you are not planning to implement them.

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+            "Temperature: {result.temperature}\n"
+            "Remaining: {result.remaining}\n"
+            "Cooking delayed: {result.cooking_delayed}\n"
+            "Duration: {result.duration}\n"
+            "Keep warm: {result.keep_warm}\n"
+            "Rice: {result.rice}\n"
+            "Taste: {result.taste}\n"
+            "Favorite: {result.favorite}\n",
+        )
+    )
```
rytilahti, line 340:
```suggestion
    @command()
```
This can be cleaned after adding the necessary decorators :-)

rytilahti (changes requested):
Hi @sschirr and thanks for the PR. My sincerest apologies that I have somehow managed to ignore this... I added some comments inline, but for most of the parts this seems to be ready to go in my opinion.

miio/cooker_multi.py
```diff
@@ -0,0 +1,419 @@
...
+    def set_schedule_duration(self, duration):
+        """Set the schedule time (delay before cooking) in minutes."""
+        schedule_flag = self.profile_bytes[14] & 0x80
+        self.profile_bytes[14] = math.floor(duration / 60) & 0xFF
+        self.profile_bytes[14] |= schedule_flag
+        self.profile_bytes[15] = (duration % 60 | self.profile_bytes[15] & 0x80) & 0xFF
+
+        self.update_checksum()
+
+    def is_akw_enabled(self):
```
rytilahti, line 196: Please add docstrings to all public API. Also, is there a better name for `akw` or is this somewhat common knowledge among users of these devices?

JarbasAl: `auto_keep_warm` is longer but more descriptive, i had to look up what `akw` was

miio/cooker_multi.py
sschirr: done

miio/cooker_multi.py
sschirr: done

miio/cooker_multi.py
sschirr: I am not an python expert. But from this doc it looks good to me 
[https://docs.python.org/3/library/enum.html#enum.Enum._missing_](https://docs.python.org/3/library/enum.html#enum.Enum._missing_)

miio/cooker_multi.py
sschirr: done

miio/cooker_multi.py
sschirr: I removed this property.

miio/cooker_multi.py
sschirr: Yes, agree. I change it to calculate a datetime from t_left and t_pre. I also cut millis and seconds from the result to prevent fluctuations.

miio/cooker_multi.py
sschirr: done

miio/cooker_multi.py
sschirr: done

miio/cooker_multi.py
sschirr: changed

sschirr:
Thanks for reviewing the changes!

Right now I am not able to do tests with a real device but I might be able to in the next days.
If someone tries the new changes please leave some feedback :)

miio/integrations/chunmi/cooker_multi/cooker_multi.py
```diff
@@ -0,0 +1,400 @@
...
+    @sensor("Selected favorite recipe")
+    def favorite(self) -> None:
+        """Favored recipe id."""
+        return self.data["favs"]
+
+
+class MultiCooker(Device):
+    """Main class representing the multi cooker."""
+
+    _supported_models = [*MODEL_MULTI]
```
rytilahti, line 319:
```suggestion
    _supported_models = [MODEL_MULTI]
```
I think? Could you try `miiocli device info` on the device to see if it gets detected correctly?

miio/integrations/chunmi/cooker_multi/cooker_multi.py
```diff
@@ -0,0 +1,400 @@
...
+            "code",
+            "en_warm",
+            "t_congee",
+            "t_love",
+            "boil",
+        ]
+
+        values = []
+        for prop in properties:
+            values.append(self.send("get_prop", [prop])[0])
```
rytilahti, line 350: I suppose the device doesn't support querying multiple properties at once? Just wondering as this can be a rather slow operation depending on how responsive the device is..

pyproject.toml
```diff
@@ -41,6 +41,7 @@ croniter = ">=1"
 defusedxml = "^0"
 pydantic = ">=1,<3"
 PyYAML = ">=5,<7"
+crcmod = "^1.7"
```
rytilahti, line 45: It'd be nice if this would be an optional extra, given it's only used for this for now.

rytilahti (commented):
A couple of more minor things I noticed, otherwise this seems good to go from my side, but please give it a proper testing on a real device :-)

miio/cooker_multi.py
rytilahti: Great, thanks :+1:

miio/integrations/chunmi/cooker_multi/cooker_multi.py
sschirr: Correct. Fixed

miio/cooker_multi.py
sschirr: I found a problem about this. Usually the cooking process is a PreCook phase which is a scheduled delay before the actual Running phase starts. During the PreCook phase the cooker only reports the remaining minutes left. If I compute a timestamp here and the pull rate from home assistant is higher than 1 minute the resulting timestamp will fluctuate. For the Running phase the problem is not that bad because in this phase the cooker is providing remaining seconds. Right now I left the "remaining" property which provides remaining minutes and the "ready_at" property for reference.

tl;dr:
"remaining" property will spam HA with the remaining minutes; "ready_at" property will spam HA with a fluctuating timestamp (03:14 -> 03:15 -> 03:14 -> 03:15 ...)

Any idea for a clean solution?

miio/integrations/chunmi/cooker_multi/cooker_multi.py
sschirr: When I try to query multiple properties at once the response is only containing the value for the first property queried. I don't really know if this can be investigated further but I guess it's not possible.

rytilahti: Okay, it happens on some devices so that's fine. If you run `pre-commit run -a` locally, it should fix the linting issues.

miio/cooker_multi.py
sschirr: I removed the ready_at sensor in favor of the remaining sensor to keep the multicooker integration similar to the other cooker integration

pyproject.toml
sschirr: done

sschirr:
I tested the changes with my fork of the xiaomi_cooker HA component ([xiaomi_cooker](https://github.com/sschirr/xiaomi_cooker)) again with a real device. It looks good to me.

rytilahti (approved):
I moved the import of `crcmod` inside the checksum calculation & run poetry update to fix the CI. This looks ready to go for me, I'll merge this as soon as the CI finishes, thanks @sschirr! :+1:

miio/cooker_multi.py
rytilahti: Ah, perhaps some sort of rounding could have worked here. Anyway, as it's now removed I'll just mark this as resolved.

PR #1019, opened 2021-04-20
babai.curtain.190812

xvlady:
add my babai.curtain.190812

### DIFF
--- a/miio/curtain_babai.py
+++ b/miio/curtain_babai.py
@@ -1,0 +1,69 @@
+from enum import Enum
+
+import click
+
+from .click_common import EnumType, command, format_output
+from .miot_device import MiotDevice
+
+_MAPPING = {
+    # Source
+    # https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:curtain:0000A00C:babai-190812:1:0000C805
+    # Curtain
+    # "motor_control": {"siid": 2, "piid": 1},  # 0 - Pause, 1 - Open, 2 - Close
+    "current_position": {"siid": 2, "piid": 2},  # Range: [0, 100, 1]
+    "target_position": {"siid": 2, "piid": 4},  # Range: [0, 100, 1]
+    "mode": {"siid": 2, 'piid': 4},  # 0 - Normal,  1 - Reversal, 2 - Calibrate
+    # Info
+    "Device Manufacturer": {"siid": 1, 'piid': 1},
+    "Device Model": {"siid": 1, 'piid': 2},
+    "Device Serial Number": {"siid": 1, 'piid': 3},
+    "Current Firmware Version": {"siid": 1, 'piid': 4},
+}
+# Model: "OnViz Curtain Controller (Wi-Fi)"
+MODEL_CURTAIN_BABAI = "babai.curtain.190812"
+
+
+class MotorControl(Enum):
+    Pause = 0
+    Open = 1
+    Close = 2
+
+
+class Mode(Enum):
+    Normal = 0
+    Reverse = 1
+    Calibrate = 2
+
+
+class BabaiCurtainMiot(MiotDevice):
+    """Main class representing the babai.curtain.190812 curtain."""
+
+    mapping = _MAPPING
+
+    @command(
+        click.argument("motor_control", type=EnumType(MotorControl)),
+        default_output=format_output("Set motor control to {motor_control}"),
+    )
+    def set_motor_control(self, motor_control: MotorControl):
+        """Set motor control."""
+        return self.set_property("motor_control", motor_control.value)
+
+    @command(
+        click.argument("target_position", type=int),
+        default_output=format_output("Set target position to {target_position}"),
+    )
+    def set_target_position(self, target_position: int):
+        """Set target position."""
+        if target_position < 0 or target_position > 100:
+            raise ValueError(
+                "Value must be between [0, 100] value, was %s" % target_position
+            )
+        return self.set_property("target_position", target_position)
+
+    @command(
+        click.argument("mode", type=EnumType(Mode)),
+        default_output=format_output("Set mode to {mode}"),
+    )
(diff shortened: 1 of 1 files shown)


miio/curtain_babai.py
```diff
@@ -0,0 +1,69 @@
...
+    # Curtain
+    # "motor_control": {"siid": 2, "piid": 1},  # 0 - Pause, 1 - Open, 2 - Close
+    "current_position": {"siid": 2, "piid": 2},  # Range: [0, 100, 1]
+    "target_position": {"siid": 2, "piid": 4},  # Range: [0, 100, 1]
+    "mode": {"siid": 2, 'piid': 4},  # 0 - Normal,  1 - Reversal, 2 - Calibrate
+    # Info
+    "Device Manufacturer": {"siid": 1, 'piid': 1},
+    "Device Model": {"siid": 1, 'piid': 2},
+    "Device Serial Number": {"siid": 1, 'piid': 3},
+    "Current Firmware Version": {"siid": 1, 'piid': 4},
```
rytilahti, line 20:
```suggestion
```
No need for these common properties here.

miio/curtain_babai.py
```diff
@@ -0,0 +1,69 @@
...
+    Normal = 0
+    Reverse = 1
+    Calibrate = 2
+
+
+class BabaiCurtainMiot(MiotDevice):
+    """Main class representing the babai.curtain.190812 curtain."""
+
+    mapping = _MAPPING
+
```
rytilahti, line 42: There is no `status()` nor status container defined in this class for the properties defined in the mapping.  Could you add one (and include an example response to the docstring of the container like shown here https://github.com/rytilahti/python-miio/blob/master/miio/heater_miot.py#L47)?

rytilahti:
Also, regarding to that linting error, you can run `pre-commit run -a` or `tox -e lint` to run all the linters locally.

miio/curtain_babai.py
```diff
@@ -0,0 +1,201 @@
...
+    @property
+    def status(self) -> MotorControl:
+        """Device status."""
+        return MotorControl(self.data["motor_control"])
+
+    @property
+    def motor_control(self) -> MotorControl:
+        """Device status."""
+        return MotorControl(self.data["motor_control"])
+
```
rytilahti, line 55:
```suggestion
```
As this is just a duplicate of status, wouldn't it be better to remove this altogether?

miio/curtain_babai.py
```diff
@@ -0,0 +1,201 @@
...
+    @property
+    def is_open(self) -> bool:
+        """Device status."""
+        return self.motor_control == MotorControl.Open
+
+    @property
+    def is_close(self) -> bool:
```
rytilahti, line 62:
```suggestion
    def is_closed(self) -> bool:
```

miio/curtain_babai.py
```diff
@@ -0,0 +1,201 @@
...
+        """Device status."""
+        return self.motor_control == MotorControl.Close
+
+    @property
+    def mode(self) -> Mode:
+        """Motor rotation polarity."""
+        return Mode(self.data["mode"])
```
rytilahti, line 69: Please update the docstring to contain an example of this information (and also others that are not currently mentioned).

miio/curtain_babai.py
```diff
@@ -0,0 +1,201 @@
...
+            "",
+            "Motor control: {result.motor_control}\n"
+            "Motor mode: {result.mode}\n"
+            "Current position: {result.current_position}\n"
+            "Target position: {result.target_position}\n",
+        )
+    )
+    def status(self) -> CurtainStatus:
+        """Retrieve properties."""
+        # if use get_properties_for_mapping and max_len>1 then code -4004.
```
rytilahti, line 98: So the device allows only requesting one property per request? If you define `max_properties=1`, is that not working in this case?

miio/curtain_babai.py
```diff
@@ -0,0 +1,201 @@
...
+        try:
+            prop = self.get_property_by(siid=v["siid"], piid=v["piid"])
+        except DeviceError as e:
+            if e.code != -9999:
+                raise
+            sleep(5)
+            prop = self.get_property_by(siid=v["siid"], piid=v["piid"])
+        for p in prop:
+            p["did"] = name
+        return prop
```
rytilahti, line 119: There should be no manual handling of this sort, it's better to find a proper solution and use that instead. Sleeping for five seconds sounds also exaggerated.

miio/curtain_babai.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @property
+    def current_position(self) -> int:
+        """Run time of the motor."""
+        return self._get_property_value("current_position")
+
+    @property
+    def target_position(self) -> int:
+        """Target curtain position."""
+        return self._get_property_value("target_position")
```
rytilahti, line 147: Properties should do no I/O, but I'd simply remove these as the information is available inside the status container.

miio/curtain_babai.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @property
+    def is_open(self) -> bool:
+        """Device status."""
+        return self.motor_control == MotorControl.Open
+
+    @property
+    def is_close(self) -> bool:
+        """Device status."""
+        return self.motor_control == MotorControl.Close
```
rytilahti, line 181: No need for these here, the container returned by `status()` should contain all the necessary status information (and accessors for those).

miio/curtain_babai.py
xvlady: {'code': -9999, 'message': 'user ack timeout'}

miio/curtain_babai.py
xvlady: random code = -9999 from status vs 1 request

miio/curtain_babai.py
xvlady: status not lasy ^(

PR #1028, opened 2021-05-02
Fix exception on devices with removed lan_ctrl

Kirmas:
Model: yeelink.light.ceiling22
Firmware version: 2.0.6_0030 
lan_ctrl was removed in firmware 2.0.6_0030 for this lamp. That's why we have exception when trying to convert None to int during yeelight status command. 
This pull request fix this.


rytilahti:
Thanks! :+1:

PR #1029, opened 2021-05-03
Add basic dmaker.fan.p8 support and streamline miOT fan implementations

syssi:
Closes: #882

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,7 +110,7 @@ Supported devices
 -  Xiaomi Philips Zhirui Bedroom Smart Lamp
 -  Huayi Huizuo Lamps
 -  Xiaomi Universal IR Remote Controller (Chuangmi IR)
--  Xiaomi Mi Smart Pedestal Fan V2, V3, SA1, ZA1, ZA3, ZA4, 1C, P5, P9, P10, P11
+-  Xiaomi Mi Smart Pedestal Fan V2, V3, SA1, ZA1, ZA3, ZA4, 1C, P5, P8, P9, P10, P11
 -  Xiaomi Rosou SS4 Ventilator (leshow.fan.ss4)
 -  Xiaomi Mi Air Humidifier V1, CA1, CA4, CB1, MJJSQ, JSQ, JSQ1, JSQ001
 -  Xiaomi Mi Water Purifier (Basic support: Turn on & off)
--- a/miio/__init__.py
+++ b/miio/__init__.py
@@ -36,7 +36,7 @@ from miio.dreamevacuum_miot import DreameVacuumMiot
 from miio.exceptions import DeviceError, DeviceException
 from miio.fan import Fan, FanP5, FanSA1, FanV2, FanZA1, FanZA4
 from miio.fan_leshow import FanLeshow
-from miio.fan_miot import Fan1C, FanMiot, FanP9, FanP10, FanP11
+from miio.fan_miot import Fan1C, FanMiot, FanP8, FanP9, FanP10, FanP11
 from miio.gateway import Gateway
 from miio.heater import Heater
 from miio.heater_miot import HeaterMiot
--- a/miio/discovery.py
+++ b/miio/discovery.py
@@ -31,8 +31,12 @@ from . import (
     Cooker,
     Device,
     Fan,
+    Fan1C,
     FanLeshow,
-    FanMiot,
+    FanP8,
+    FanP9,
+    FanP10,
+    FanP11,
     Heater,
     PhilipsBulb,
     PhilipsEyecare,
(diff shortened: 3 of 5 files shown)


rytilahti (commented):
LGTM, but I'm wondering if we should rethink how we should handle model handling. From the user's viewpoint it doesn't make sense to have separate classes for similar devices, where avoidable.

One potential way to achieve this would be to 1) allow explicit definition for models (as it is currently for the main impl class) and 2) if none is given, resort to calling `info()` on the first I/O call to find out the model. What do you think?

syssi:
I've introduced a class per device to have improved CLI support. I'm unsure it's a good decision or not.

rytilahti:
I think we should not do that as it has more potential for causing confusion in the long run, so here is my proposal for the CLI use:
1. We add `model` keyword argument to `Device` class constructor (much akin to what already exists on some sub-classes). This shall default to `None` / auto-detect and saved in `self._model`.
2. We add `--model` option alongside the `--ip` and `--token`, to allow forcing specific model.
3. If the internal `self._model` is `None`, any call to `@command` decorated function causes `info()` call to find the model.
4. (Optionally,) if the class only supports a single model, it can initialize `self._model` in its constructor to skip the auto-detection.

Would that work for you? Would it be okay for you to rely on manual model definition for the next release (which I wish to get one out ASAP), and then adding this model feature for the next one after that?

syssi:
Fine for me! This PR shouldn't block the release.

rytilahti:
Okay, great! I'll try to prepare a new release later tonight then :-)

rytilahti:
I created #1038 to explore the idea I described above. It is rather hacky at this point and will need some more work, but I think something like that would be good to have in the longer term.

ha0y:
> LGTM, but I'm wondering if we should rethink how we should handle model handling. From the user's viewpoint it doesn't make sense to have separate classes for similar devices, where avoidable.
> 
> One potential way to achieve this would be to 1) allow explicit definition for models (as it is currently for the main impl class) and 2) if none is given, resort to calling `info()` on the first I/O call to find out the model. What do you think?

Agreed. And another thing we should rethink is that how we handle new devices. Now we have a long list of miot-enabled devices(https://github.com/rytilahti/python-miio/issues/627). And for each new device, we write many similar codes.https://github.com/rytilahti/python-miio/blob/62ee7483fd54bbe8a4971c108650cfae7d926e18/miio/fan_miot.py#L16-L82 

So I think we can fetch its MIoT-Spec from the server, and parse it so that it is automatically adapted. That 's what I have done in  https://github.com/ha0y/xiaomi_miot_raw . Thanks to you and @syssi because it would not have been possible without your great works. If you don't mind, please give it a try. And the adapting method is in https://github.com/ha0y/xiaomi_miot_raw/blob/master/custom_components/xiaomi_miot_raw/deps/miot_device_adapter.py , which simply tries to find out all supported services and properties.

rytilahti:
Hi @ha0y and sorry for the late response, I have been following your efforts to provide generic support for miot devices. And I completely agree, that the current way we are doing miot support in this library does not scale. However, I have two concerns here:
1. I think this library should not depend on externally sourced resources (like the definition files)
2. Vendoring those files to be a part of the library is also not great

I wrote some initial ideas on adding generic device types to this library here https://github.com/rytilahti/python-miio/issues/1059#issuecomment-851070398 -- do you have an insight how easy it would to create rather generic, device-type specific interfaces for miot devices? In other words, do you know if, for example the vacuum devices, use same set of siid&piid values for a specific set of basic functionalities, or would that require device-specific mappings for each and every device?

marcelrv:
@rytilahti as dat as I have seen.. the devices do not have same set of siid&piid.

There are some overlaps, but imho you can't assume same piid/siid point to the same sorry of property.
Inside brand e.g. viomi vacuums there is more overlap, but even then it may differ

ha0y:
> I think this library should not depend on externally sourced resources (like the definition files hosted online)
> Vendoring those files to be a part of the library is also not great

Yes, it is not possible to vendor these files in the library, because each device has their own definition files, and dozens of devices join Xiaomi IoT every week. 

> In other words, do you know if, for example the vacuum devices, use same set of siid&piid values for a specific set of basic functionalities, or would that require device-specific mappings for each and every device?

It is also not possible to use the same set of siid & piid because they are not fixed. Accroding to Xiaomi's document, the manufacturer can specify the service and property of their device, which will change the order of them. And the apps doesn't identify those services and properties by iid, but by URN, which (as far as I know) only exists in spec files.

When it comes to fan: many manufacturers create a private service to provide some additional functions, including 1%-100% speed. This is also a tough problem.

PR #1035, opened 2021-05-06
Improve Yeelight support (expose more properties, add support for secondary lights)

Kirmas:
In this pull request I want to add more property to Yeelight (I looked to python-yeelight and operation specification). Base reason of this because Yeelight closing lan controle for theirs lamps that`s why I want to add all functionality from python-yeelight to this repository. This is first PR from many :)

@rytilahti due to this pull request I want to ask some question:
1) First off all Yeelight has very different lamp. Is it ok that we use one Device class to describe them all?
2) Second I use "status" command to get all type of the properties. Maybe better add additional command like "bg_status" for Background light. What is better for Home Assistant integration?

I would be grateful for any criticism

### DIFF
--- a/miio/tests/test_yeelight.py
+++ b/miio/tests/test_yeelight.py
@@ -21,6 +21,21 @@ class DummyLight(DummyDevice, Yeelight):
             "name": "test name",
             "lan_ctrl": "1",
             "save_state": "1",
+            "delayoff": "60",
+            "music_on": "1",
+            "flowing": "0",
+            "flow_params": "",
+            "active_mode": "1",
+            "nl_br": "100",
+            "bg_power": "off",
+            "bg_bright": "100",
+            "bg_lmode": "2",
+            "bg_rgb": "16711680",
+            "bg_hue": "359",
+            "bg_sat": "100",
+            "bg_ct": "3584",
+            "bg_flowing": "0",
+            "bg_flow_params": "",
         }
 
         self.return_values = {
--- a/miio/tests/test_yeelight.py
+++ b/miio/tests/test_yeelight.py
@@ -81,11 +81,11 @@ class TestYeelight(TestCase):
         assert repr(status) == repr(YeelightStatus(self.device.start_state))
 
         assert status.name == self.device.start_state["name"]
-        assert status.lights[0].is_on is False
-        assert status.lights[0].brightness == 100
-        assert status.lights[0].color_temp == 3584
-        assert status.lights[0].color_mode == YeelightMode.ColorTemperature
-        assert status.lights[0].rgb is None
+        assert status.is_on is False
+        assert status.brightness == 100
+        assert status.color_temp == 3584
+        assert status.color_mode == YeelightMode.ColorTemperature
+        assert status.rgb is None
         assert status.developer_mode is True
         assert status.save_state_on_change is True
 
@@ -95,21 +95,21 @@ class TestYeelight(TestCase):
 
     def test_on(self):
         self.device.off()  # make sure we are off
-        assert self.device.status().lights[0].is_on is False
+        assert self.device.status().is_on is False
 
         self.device.on()
-        assert self.device.status().lights[0].is_on is True
+        assert self.device.status().is_on is True
 
     def test_off(self):
         self.device.on()  # make sure we are on
-        assert self.device.status().lights[0].is_on is True
+        assert self.device.status().is_on is True
 
         self.device.off()
-        assert self.device.status().lights[0].is_on is False
+        assert self.device.status().is_on is False
 
     def test_set_brightness(self):
         def brightness():
-            return self.device.status().lights[0].brightness
+            return self.device.status().brightness
 
         self.device.set_brightness(50)
         assert brightness() == 50
@@ -125,7 +125,7 @@ class TestYeelight(TestCase):
 
     def test_set_color_temp(self):
         def color_temp():
-            return self.device.status().lights[0].color_temp
+            return self.device.status().color_temp
 
         self.device.set_color_temp(2000)
         assert color_temp() == 2000
@@ -140,7 +140,7 @@ class TestYeelight(TestCase):
 
     def test_set_rgb(self):
         def rgb():
-            return self.device.status().lights[0].rgb
+            return self.device.status().rgb
 
         self.device._reset_state()
         self.device._set_state("color_mode", [1])
@@ -175,7 +175,7 @@ class TestYeelight(TestCase):
     @pytest.mark.skip("hsv is not properly implemented")
     def test_set_hsv(self):
         self.reset_state()
-        hue, sat, val = self.device.status().lights[0].hsv
+        hue, sat, val = self.device.status().hsv
         assert hue == 359
         assert sat == 100
         assert val == 100
@@ -215,7 +215,7 @@ class TestYeelight(TestCase):
 
     def test_toggle(self):
         def is_on():
-            return self.device.status().lights[0].is_on
+            return self.device.status().is_on
 
         orig_state = is_on()
         self.device.toggle()
--- a/miio/tests/test_yeelight.py
+++ b/miio/tests/test_yeelight.py
@@ -81,11 +81,11 @@ class TestYeelight(TestCase):
         assert repr(status) == repr(YeelightStatus(self.device.start_state))
 
         assert status.name == self.device.start_state["name"]
-        assert status.is_on is False
-        assert status.brightness == 100
-        assert status.color_temp == 3584
-        assert status.color_mode == YeelightMode.ColorTemperature
-        assert status.rgb is None
+        assert status.lights[0].is_on is False
+        assert status.lights[0].brightness == 100
+        assert status.lights[0].color_temp == 3584
+        assert status.lights[0].color_mode == YeelightMode.ColorTemperature
+        assert status.lights[0].rgb is None
         assert status.developer_mode is True
         assert status.save_state_on_change is True
 
@@ -95,21 +95,21 @@ class TestYeelight(TestCase):
 
     def test_on(self):
         self.device.off()  # make sure we are off
-        assert self.device.status().is_on is False
+        assert self.device.status().lights[0].is_on is False
 
         self.device.on()
-        assert self.device.status().is_on is True
+        assert self.device.status().lights[0].is_on is True
 
     def test_off(self):
         self.device.on()  # make sure we are on
-        assert self.device.status().is_on is True
+        assert self.device.status().lights[0].is_on is True
 
         self.device.off()
-        assert self.device.status().is_on is False
+        assert self.device.status().lights[0].is_on is False
 
     def test_set_brightness(self):
         def brightness():
-            return self.device.status().brightness
+            return self.device.status().lights[0].brightness
 
         self.device.set_brightness(50)
         assert brightness() == 50
@@ -125,7 +125,7 @@ class TestYeelight(TestCase):
 
     def test_set_color_temp(self):
         def color_temp():
-            return self.device.status().color_temp
+            return self.device.status().lights[0].color_temp
 
         self.device.set_color_temp(2000)
         assert color_temp() == 2000
@@ -140,7 +140,7 @@ class TestYeelight(TestCase):
 
     def test_set_rgb(self):
         def rgb():
-            return self.device.status().rgb
+            return self.device.status().lights[0].rgb
 
         self.device._reset_state()
         self.device._set_state("color_mode", [1])
@@ -175,7 +175,7 @@ class TestYeelight(TestCase):
     @pytest.mark.skip("hsv is not properly implemented")
     def test_set_hsv(self):
         self.reset_state()
-        hue, sat, val = self.device.status().hsv
+        hue, sat, val = self.device.status().lights[0].hsv
         assert hue == 359
         assert sat == 100
         assert val == 100
@@ -215,7 +215,7 @@ class TestYeelight(TestCase):
 
     def test_toggle(self):
         def is_on():
-            return self.device.status().is_on
+            return self.device.status().lights[0].is_on
 
         orig_state = is_on()
         self.device.toggle()
--- a/miio/tests/test_yeelight.py
+++ b/miio/tests/test_yeelight.py
@@ -38,6 +38,144 @@ class DummyLight(DummyDevice, Yeelight):
             self.state["power"] = "on"
 
 
+class DummyCommonBulb(DummyLight):
+    def __init__(self, *args, **kwargs):
+        self.state = {
+            "name": "test name",
+            "lan_ctrl": "1",
+            "save_state": "1",
+            "delayoff": "0",
+            "music_on": "1",
+            "power": "off",
+            "bright": "100",
+            "color_mode": "2",
+            "rgb": "",
+            "hue": "",
+            "sat": "",
+            "ct": "3584",
+            "flowing": "",
+            "flow_params": "",
+            "active_mode": "",
+            "nl_br": "",
+            "bg_power": "",
+            "bg_bright": "",
+            "bg_lmode": "",
+            "bg_rgb": "",
+            "bg_hue": "",
+            "bg_sat": "",
+            "bg_ct": "",
+            "bg_flowing": "",
+            "bg_flow_params": "",
+        }
+        super().__init__(*args, **kwargs)
+
+
+@pytest.fixture(scope="class")
+def dummycommonbulb(request):
+    request.cls.device = DummyCommonBulb()
+    # TODO add ability to test on a real device
+
+
+@pytest.mark.usefixtures("dummycommonbulb")
+class TestYeelightCommon(TestCase):
+    def test_on(self):
+        self.device.off()  # make sure we are off
+        assert self.device.status().is_on is False
+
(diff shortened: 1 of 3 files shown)


miio/yeelight.py
```diff
@@ -22,8 +22,8 @@ class YeelightMode(IntEnum):
 
 class YeelightStatus(DeviceStatus):
     def __init__(self, data):
-        # ['power', 'bright', 'ct',   'rgb',      'hue', 'sat', 'color_mode', 'name', 'lan_ctrl', 'save_state']
-        # ['on',    '100',    '3584', '16711680', '359', '100', '2',          'name', '1',        '1']
+        # ['name', 'lan_ctrl', 'save_state', 'delayoff', 'music_on', 'power', 'bright', 'color_mode', 'rgb',      'hue', 'sat', 'ct',   'flowing', 'flow_params', 'active_mode', 'nl_br', 'bg_power', 'bg_bright', 'bg_lmode', 'bg_rgb',   'bg_hue', 'bg_sat', 'bg_ct', 'bg_flowing', 'bg_flow_params']
```
rytilahti, line 25: Considering the data is actually in dictionary form here, could you please replace the old docstring with a dictionary showing the information? Adding a simple `print()` for the data to the `__init__` is likely the simplest way to do that!

miio/yeelight.py
```diff
@@ -85,6 +85,109 @@ def name(self) -> str:
...
+    def color_flow_params(self) -> Optional[str]:
+        """Return color flowing params."""
+        if self.color_flowing:
+            return self.data["flow_params"]
+        return None
+
+    @property
+    def delay_off(self) -> int:
+        """Return delay in minute before bulb is off."""
+        return self.data["delayoff"]
```
rytilahti, line 103: I think all data is returned as strings, so this (and maybe others) will require casting. Could you please check that?

miio/yeelight.py
```diff
@@ -109,30 +212,66 @@ def __init__(self, *args, **kwargs):
...
+            # background light properties
+            "bg_power",
+            "bg_bright",
+            "bg_lmode",
+            "bg_rgb",
+            "bg_hue",
+            "bg_sat",
+            "bg_ct",
+            "bg_flowing",
+            "bg_flow_params",
```
rytilahti, line 274: I haven't yet checked this PR out with my lights (which do not support background lights), but it may be needed to do separate queries depending on the light. I'll try to find some time in the near future to check that out and I'll report back.

miio/yeelight.py
```diff
@@ -85,6 +85,109 @@ def name(self) -> str:
...
+            return int(self.data["nl_br"])
+        return None
+
+    @property
+    def is_bg_on(self) -> Optional[bool]:
+        """Return whether the background light is on or off."""
+        bg_power = self.data["bg_power"]
+        if bg_power:
+            return bg_power == "on"
+        return None
```
rytilahti, line 135: Just an idea that may not be relevant depending on how the query works on lights with no background light: it might make sense to have the mainlight as it is, and provide a `lights()` getter that will return a list of all lights (main, background, other?) presented as objects of a `YeelightLight` class. This would remove the need of code duplication (just using a prefixed variable name for background light).

I'm thinking about how one would integrate this downstream, and having a list of exposed lights and using a common API would make it much simpler to handle there (e.g., in homeassistant, creating multiple light instances based on available lights). Just some food for thought for now though, that might need some experiment to get a feel how it would work.

rytilahti (commented):
Just a very quick review, I haven't yet tested the code on my own lights but I'll try to do that soonish to make sure this doesn't break anything.

rytilahti:
Hi @Kirmas and thanks for the PR! :-)

> * First off all Yeelight has very different lamp. Is it ok that we use one Device class to describe them all?

Yes, I think that is fine. Using several different classes would make it really confusing for users at some point, something we have been seeing recently with some other integrations.

At some point in the future, I have some plans to integrate some sort of "supported features" property based on the model to make it simpler to allow the class (and the downstream) to adapt to different types of lights.

> * Second I use "status" command to get all type of the properties. Maybe better add additional command like "bg_status" for Background light. What is better for Home Assistant integration?

I think the best approach wrt homeassistant integration would be to have a `YeelightLight` (or similar) class to represent a single controllable light. A single `status()` could return a regular status of the main lamp, and the status container would have a property like `lights()` to return a list of lights that can be controlled.

This way the homeassistant integration would just require a single update (and single light implementation) even if there are multiple light elements available. Or at least this is how I imagine it would work :-)

miio/yeelight.py
Kirmas: Ok, but I have different lists of available parameters for different types of lamps. 
			
```
yeelink.light.ceiling4, yeelink.light.ceiling20
{'name': '', 'lan_ctrl': '1', 'save_state': '1', 'delayoff': '0', 'music_on': '',  'power': 'off', 'bright': '1',   'color_mode': '2', 'rgb': '',        'hue': '',    'sat': '',   'ct': '4115', 'flowing': '0', 'flow_params': '0,0,2000,3,0,33,2000,3,0,100',                                                                                                          'active_mode': '1', 'nl_br': '1', 'bg_power': 'off', 'bg_bright': '100', 'bg_lmode': '1', 'bg_rgb': '15531811', 'bg_hue': '65',  'bg_sat': '86', 'bg_ct': '4000', 'bg_flowing': '0', 'bg_flow_params': '0,0,3000,4,16711680,100,3000,4,65280,100,3000,4,255,100'}
yeelink.light.ceiling1
{'name': '', 'lan_ctrl': '1', 'save_state': '1', 'delayoff': '0', 'music_on': '',  'power': 'off', 'bright': '100', 'color_mode': '2', 'rgb': '',        'hue': '',    'sat': '',   'ct': '5200', 'flowing': '0', 'flow_params': '',                                                                                                                                      'active_mode': '0', 'nl_br': '0', 'bg_power': '',    'bg_bright': '',    'bg_lmode': '',  'bg_rgb': '', 'bg_hue': '', 'bg_sat': '', 'bg_ct': '', 'bg_flowing': '', 'bg_flow_params': ''}
yeelink.light.ceiling22 - like yeelink.light.ceiling1 but without "lan_ctrl"
{'name': '', 'lan_ctrl': '',  'save_state': '1', 'delayoff': '0', 'music_on': '',  'power': 'off', 'bright': '84',  'color_mode': '2', 'rgb': '',        'hue': '',    'sat': '',   'ct': '4000', 'flowing': '0', 'flow_params': '0,0,800,2,2700,50,800,2,2700,30,1200,2,2700,80,800,2,2700,60,1200,2,2700,90,2400,2,2700,50,1200,2,2700,80,800,2,2700,60,400,2,2700,70', 'active_mode': '0', 'nl_br': '0', 'bg_power': '',    'bg_bright': '',    'bg_lmode': '',  'bg_rgb': '', 'bg_hue': '', 'bg_sat': '', 'bg_ct': '', 'bg_flowing': '', 'bg_flow_params': ''}
yeelink.light.color4, yeelink.light.color5
{'name': '', 'lan_ctrl': '1', 'save_state': '1', 'delayoff': '0', 'music_on': '0', 'power': 'off', 'bright': '100', 'color_mode': '1', 'rgb': '2353663', 'hue': '186', 'sat': '86', 'ct': '6500', 'flowing': '0', 'flow_params': '0,0,1000,1,16711680,100,1000,1,65280,100,1000,1,255,100',                                                                               'active_mode': '',  'nl_br': '',  'bg_power': '',    'bg_bright': '',    'bg_lmode': '',  'bg_rgb': '', 'bg_hue': '', 'bg_sat': '', 'bg_ct': '', 'bg_flowing': '', 'bg_flow_params': ''}
```
Should I combine this or add all types?

miio/yeelight.py
Kirmas: I did not notice, fixed in e06a580. Thanks.

miio/yeelight.py
Kirmas: I already checked it. And its work but looks not cool:
```
Name: 
Developer mode: True
Update default on change: True
Delay in minute before off: 0
Music mode: None
Light
   Power: False
   Brightness: 100
   Color mode: 2
   RGB: None
   HSV: None
   Temperature: 5200
   Color flowing mode: False
   Color flowing parameters: None
Moonlight
   Is in mode: False
   Moonlight mode brightness: 0
Background light
   Power: None
   Brightness: None
   Color mode: None
   RGB: None
   HSV: None
   Temperature: None
   Color flowing mode: None
   Color flowing parameters: None
```
Where I can look on separated subdevices. (My background is C++ and python is additional language for me, i don't really know how @command(default output...) works.)

Kirmas:
@rytilahti additional question: what type off data better use for flow_params and bg_flow_params:
output looks like `'0,0,1000,1,16711680,100,1000,1,65280,100,1000,1,255,100'` I use string but maybe its array of int or something else?

syssi:
I string is easier to manage for now. If it's just passed around and not manipulated. Adding new flows by configuration as strings would be easy.

miio/yeelight.py
rytilahti: I would say that it's best to add this for all types for now. In the future we may use the model information / this information to report downstreams which features are supported by the light (e.g., by having a `supported_features` property for the class) so that they can adapt accordingly.

Having the properties in the request is just fine as long as it doesn't break on devices that do not support the property, which seems to be the case (i.e., they simply return empty string when that happens.

rytilahti:
@Kirmas I think the yeelight's spec defines the elements of the flow, so I think it would be best to use a separate container class (similar to the status container) to parse and store this data for easier use. But having it as a string for the time being is also fine (and better for comprehension than converting it into a list).

python-yeelight does implement some parsing (and building of flows), but I'm not sure if that can be easily used here. We want no external dependencies for this, and there might be a clash in licenses between these two projects which makes direct code sharing harder..

miio/yeelight.py
rytilahti: About the command output, I wouldn't worry about it that much, the cli tool is there just for completeness as the main use case for this lib is to use it as a library :-)

Changing that will require some restructuring / work, so I think the best thing to do (for the time being) is to document which properties are supported by which models.

I am wondering if we could find a way to share the "database" of device features with python-yeelight as they already have collected information that could be useful in the future? See https://gitlab.com/stavros/python-yeelight/-/blob/master/yeelight/main.py#L29 - ping @skorokithakis

miio/yeelight.py
Kirmas: done 5d6132b

miio/yeelight.py
skorokithakis: Hmm, yes, that's a good idea! Maybe we could make a separate common library, to avoid adding either as a dependency of the other.

Kirmas: What can I help with this? As I understand we must wait answer from @skorokithakis? 
As for me. I sink good idea is moving shared data to external lib.

miio/yeelight.py
Kirmas: Yes it looks like good idea. But "yeelight" for some reason use different names for  color mode property. This is "color_mode" for main lamp and "bg_lmode" for background. But I think it can be hardcoded only for this situation. I'll looked to this.

miio/yeelight.py
skorokithakis: Isn't the data all we both need? We could agree on a common format and make a new library that's simply this data.

Kirmas: @skorokithakis  for my opinion your data format is great. But I thinked about simple getter like  
`get_known_models` and maybe something like `get_specs_for_model(str: model_name)` because this 2 function looks like general for any libs. 
But it may be only the data. 
I only start to work with this and not sure what data Is important for me in future. If you know any other general datas that you have I`ll be very heppy if you shared it.

@rytilahti What do you think?

skorokithakis: That makes sense, but it sounds like those getters will be aliases to `MODEL_DICT.keys()` and `MODEL_DICT[model_name]`, which isn't really a hassle to write. Then again, I'm not against writing a simple class either.

miio/yeelight.py
Kirmas: @rytilahti this is done in f43c652. But have problem with cmd output. Could you provide any suggestion how I can show my list of objects?

Kirmas: additional commit 612ec6f with fix command output. Output looks like 
```
Name: 
Developer mode: True
Update default on change: True
Delay in minute before off: 0
Music mode: False
Lights: 
[
   Sub Light - Main 
      Power: False
      Brightness: 100
      Color mode: 2
      RGB: None
      HSV: None
      Temperature: 6500
      Color flowing mode: False
      Color flowing parameters: None
]
Moonlight
   Is in mode: None
   Moonlight mode brightness: None
```
or 
```
Name: 
Developer mode: True
Update default on change: True
Delay in minute before off: 0
Music mode: None
Lights: 
[
   Sub Light - Main 
      Power: False
      Brightness: 100
      Color mode: 2
      RGB: None
      HSV: None
      Temperature: 6500
      Color flowing mode: False
      Color flowing parameters: None
, 
   Sub Light - Background 
      Power: False
      Brightness: 100
      Color mode: 1
      RGB: (255, 77, 106)
      HSV: None
      Temperature: None
      Color flowing mode: False
      Color flowing parameters: None
]
Moonlight
   Is in mode: False
   Moonlight mode brightness: 0
```
But I don't know is it norm to use  `__repr__` for this or not.

rytilahti: I think using the repr is the simplest way to make it print the list of objects, like you already found out :-) I just tested the PR with `yeelink.light.color1` and the status query is working just fine with all the extra properties.

miio/tests/test_yeelight.py
```diff
@@ -66,11 +81,11 @@ def test_status(self):
...
-        assert status.brightness == 100
-        assert status.color_temp == 3584
-        assert status.color_mode == YeelightMode.ColorTemperature
-        assert status.rgb is None
+        assert status.lights[0].is_on is False
+        assert status.lights[0].brightness == 100
+        assert status.lights[0].color_temp == 3584
+        assert status.lights[0].color_mode == YeelightMode.ColorTemperature
+        assert status.lights[0].rgb is None
         assert status.developer_mode is True
```
rytilahti, line 89: It would be great if the original property names would keep working, just internally defaulting to the main light.

Kirmas: Look 5255611

miio/yeelight.py
Kirmas: It must be better after 6d2ce92. @rytilahti Is it ok for now?

Kirmas: Now output more compact (show only not None params) Look 43ca9b3 
```
Name: 
Update default on change: True
Delay in minute before off: 0
Music mode: False
Developer mode: True
Main light
   Power: False
   Brightness: 100
   Color mode: 2
   Temperature: 6500
   Color flowing mode: False
```
```
Name: 
Update default on change: True
Delay in minute before off: 0
Developer mode: True
Main light
   Power: False
   Brightness: 100
   Color mode: 2
   Temperature: 6500
   Color flowing mode: False
Background light
   Power: False
   Brightness: 100
   Color mode: 1
   RGB: (255, 77, 106)
   Color flowing mode: False
Moonlight
   Is in mode: False
   Moonlight mode brightness: 0
```

miio/yeelight.py
skorokithakis: I'm going to unfollow this PR because I got 17 emails, please open an issue if you decide something on the library matter.

rytilahti: Yeah, the issue with shared definitions should be done in a separate issue not to be discussed further in this PR. I think having a repository shipping a yaml file containing model-specific features would be a good start.

Kirmas: Issue was created https://gitlab.com/stavros/python-yeelight/-/issues/68

Kirmas:
@rytilahti what I must to do next, for this PR? Is it ok to be merged or requires additional work?

rytilahti:
@Kirmas a test or two for the secondary light would be great to have to avoid breaking the functionality inadvertently in the future, otherwise this is good to go! :+1:

Kirmas:
@rytilahti I added new devices to the yeelight test but not sure if I did it right. Please review.
Test_status looks good in my opinion. But another test requires additional work beyond the scope of this pull request.
When adding new functions, I will add new tests.
Some tests require https://gitlab.com/stavros/python-yeelight/-/issues/68/ to understand which feature is available.

PS:
It came as a surprise to me that I can move a non-color lamp to RGB mode without any errors. I think this is dangerous. When the separate library will be created, I plan to check it in the code on our side.

Kirmas:
@rytilahti Hello? sorry i missed one week. There was no time. I have added a general test. Now everything is fine? For my opinion test status must be totally unique for different types of lamp.

Kirmas:
And other question. I couldn't understand why tests is failed. I didn't change plase were they failed. Could you help with this?

rytilahti (approved):

rytilahti:
Hi, no worries! Yes, I think it's fine for now as it is.

Hmm, that linting error is odd and I cannot reproduce it locally on this branch. Could you try to modify the `.flake8` file and add `SIM119` to the end of `ignore` list? Those are irrelevant as we still support python 3.6 series.

PR #1039, opened 2021-05-08
Add features for newer vacuums (eg Roborock S7)

### LINKED ISSUE #1040: Add setting for carpet avoidance to vacuums
As mentioned in #1039 newer vacuums like the Roborock S7 feature a setting for carpet avoidance. This settings allows for the robot to lift its mop while cleaning a carpet or alternatively avoiding carpets altogether.

I didn't find the setting by trial and error but it is present in the app.

**Device information:**
 - Model: roborock.vacuum.a15
 - Hardware version: Linux
 - Firmware version: 4.1.2_0978

**Additional context**

fettlaus:
I added the following features to the vacuums. Support for them may vary depending on the model of the vacuum. All of these at least work for the Roborock S7

- mopping mode/route can be changed (does multiple passes on deep clean)
- can detect if the mop is attached (as opposed to detection of attached water box)
- fanspeeds for Roborock S7 added
- child lock can be switched on and off
- water shortage on the water box can be queried if the vacuum has a sensor for this
- carpet cleaning mode/avoidance can be changed (see #1040)

Closes #1040

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -93,6 +93,13 @@ class WaterFlow(enum.Enum):
     Maximum = 203
 
 
+class MopMode(enum.Enum):
+    """Mop routing on S7."""
+
+    Standard = 300
+    Deep = 301
+
+
 ROCKROBO_V1 = "rockrobo.vacuum.v1"
 ROCKROBO_S5 = "roborock.vacuum.s5"
 ROCKROBO_S6 = "roborock.vacuum.s6"
@@ -746,6 +753,19 @@ class Vacuum(Device):
     def set_waterflow(self, waterflow: WaterFlow):
         """Set water flow setting."""
         return self.send("set_water_box_custom_mode", [waterflow.value])
+
+    @command()
+    def mop_mode(self) -> Optional[MopMode]:
+        """Get mop mode setting."""
+        try:
+            return MopMode(self.send("get_mop_mode")[0])
+        except ValueError:
+            return None
+
+    @command(click.argument("mop_mode", type=EnumType(MopMode)))
+    def set_mop_mode(self, mop_mode: MopMode):
+        """Set mop mode setting."""
+        return self.send("set_mop_mode", [mop_mode.value])
 
     @classmethod
     def get_device_group(cls):
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -107,6 +107,14 @@ class MopMode(enum.Enum):
     Deep = 301
 
 
+class CarpetCleaningMode(enum.Enum):
+    """Type of carpet cleaning/avoidance."""
+
+    Avoid = 0
+    Rise = 1
+    Ignore = 2
+
+
 ROCKROBO_V1 = "rockrobo.vacuum.v1"
 ROCKROBO_S5 = "roborock.vacuum.s5"
 ROCKROBO_S6 = "roborock.vacuum.s6"
@@ -697,6 +705,24 @@ class Vacuum(Device):
             "current_integral": integral,
         }
         return self.send("set_carpet_mode", [data])[0] == "ok"
+
+    @command()
+    def carpet_cleaning_mode(self):
+        """Get carpet cleaning mode/avoidance setting."""
+        try:
+            return CarpetCleaningMode(
+                self.send("get_carpet_clean_mode")[0]["carpet_clean_mode"]
+            )
+        except TypeError:
+            return None
+
+    @command(click.argument("mode", type=EnumType(CarpetCleaningMode)))
+    def set_carpet_cleaning_mode(self, mode: CarpetCleaningMode):
+        """Set carpet cleaning mode/avoidance setting."""
+        return (
+            self.send("set_carpet_clean_mode", {"carpet_clean_mode": mode.value})[0]
+            == "ok"
+        )
 
     @command()
     def stop_zoned_clean(self):
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -795,13 +795,14 @@ class Vacuum(Device):
         """Get mop mode setting."""
         try:
             return MopMode(self.send("get_mop_mode")[0])
-        except ValueError:
+        except ValueError as err:
+            _LOGGER.warning("Device returned unknown MopMode: %s", err)
             return None
 
     @command(click.argument("mop_mode", type=EnumType(MopMode)))
     def set_mop_mode(self, mop_mode: MopMode):
         """Set mop mode setting."""
-        return self.send("set_mop_mode", [mop_mode.value])
+        return self.send("set_mop_mode", [mop_mode.value])[0] == "ok"
 
     @command()
     def child_lock(self) -> bool:
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -84,6 +84,13 @@ class FanspeedE2(enum.Enum):
     Turbo = 100
 
 
+class FanspeedS7(enum.Enum):
+    Quiet = 101
+    Balanced = 102
+    Turbo = 103
+    Max = 104
+
+
 class WaterFlow(enum.Enum):
     """Water flow strength on s5 max."""
 
@@ -104,6 +111,7 @@ ROCKROBO_V1 = "rockrobo.vacuum.v1"
 ROCKROBO_S5 = "roborock.vacuum.s5"
 ROCKROBO_S6 = "roborock.vacuum.s6"
 ROCKROBO_S6_MAXV = "roborock.vacuum.a10"
+ROCKROBO_S7 = "roborock.vacuum.a15"
 
 
 class Vacuum(Device):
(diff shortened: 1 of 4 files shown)


miio/vacuum.py
```diff
@@ -84,6 +84,13 @@ class FanspeedE2(enum.Enum):
     Turbo = 100
 
 
+class FanspeedS7(enum.Enum):
+    Quiet = 101
+    Balanced = 102
+    Turbo = 103
+    Max = 104
```
rytilahti, line 91: Please use the similar naming as other vacuums above, the mid-term plan is to consolidate available fan speeds (to make it easier for homeassistant to support also non-rockrobo vacuums supported by this library).

miio/vacuum.py
```diff
@@ -93,10 +100,18 @@ class WaterFlow(enum.Enum):
     Maximum = 203
 
 
+class MopMode(enum.Enum):
+    """Mop routing on S7."""
+
+    Standard = 300
+    Deep = 301
```
rytilahti, line 107: Is this maybe similar to https://github.com/rytilahti/python-miio/blob/master/miio/viomivacuum.py#L251 ? I'm trying to gather what features are supported by different vacuums to make it possible to provide an unified vacuum interface for downstream users.

miio/vacuumcontainers.py
```diff
@@ -175,6 +190,28 @@ def is_water_box_attached(self) -> bool:
         """Return True is water box is installed."""
         return "water_box_status" in self.data and self.data["water_box_status"] == 1
 
+    @property
+    def is_water_box_carriage_attached(self) -> Optional[bool]:
+        """Return True if water box carriage (mop) is installed, None if sensor not
+        present."""
+        if "water_box_carriage_status" in self.data:
+            return self.data["water_box_carriage_status"] == 1
+        else:
```
rytilahti, line 199: No need for else here, simply return None at the end of the method.

miio/vacuumcontainers.py
```diff
@@ -175,6 +190,28 @@ def is_water_box_attached(self) -> bool:
...
+        """Returns True if water is low in the tank, None if sensor not present."""
+        if "water_shortage_status" in self.data:
+            return self.data["water_shortage_status"] == 1
+        else:
+            return None
+
+    @property
+    def is_child_lock(self) -> bool:
+        """Returns True if the child lock is active."""
+        return self.data["lock_status"] == 1
```
rytilahti, line 213: This also needs to check for if the key is given.

miio/vacuum.py
fettlaus: So what is called "turbo" in the app equals Medium and what is called "max" in the app equals Turbo? Maybe we should create some kind of mapping table in the documentation for this or it gets confusing at some point. But I fully agree with the consolidation

miio/vacuum.py
fettlaus: I think so. In deep mode the vacuum follows a tighter pattern for more intense cleaning while mopping. The viomi uses different patterns (Y and S) but in general it changes the routing of the robots so i guess you can call it similar.

Should I adapt my PR to the naming scheme based on the viomi?

rytilahti: Hmm, I think it's okay to leave it as it is here now. Standard and deep sound more generic than S & Y, so when a common API for all vacuum devices is added, the former would be the likely candidates.

miio/vacuumcontainers.py
fettlaus: Thanks to your comment i noticed that `lock_status` doesn't actually contain the status of the child lock, a call to get_child_lock_status is necessary. Will remove this method

miio/vacuum.py
rytilahti: After a second thought, I think this can stay for now as it is. I have some scribbles on a shared API and I'll create an issue to gather feedback on that later.

Anyways, I just took a look at the existing vacuum implementations for fan speeds:
Viomi: Silent, Standard, Medium, Turbo
Dreame: Quiet, Default, Medium, Strong
Rockrobo: Gentle (E2, v2 only?), Silent, Standard, Medium, Turbo

so it's quite a mess with names (and separate enums) but four different speeds is common so I think it'll be fairly simple to find a common base there when the time comes... :-)

fettlaus:
Didn't see your comment before changing it as suggested previously. But you are right, in the future it is just a matter of deciding which generic names to choose for the different speeds. In pretty much every case it is: silent, default, more, maxAm 10.05.2021 00:08 schrieb "Teemu R." ***@***.***>:
@rytilahti commented on this pull request.



In miio/vacuum.py:
> @@ -84,6 +84,13 @@ class FanspeedE2(enum.Enum):
     Turbo = 100
 
 
+class FanspeedS7(enum.Enum):
+    Quiet = 101
+    Balanced = 102
+    Turbo = 103
+    Max = 104

After a second thought, I think this can stay for now as it is. I have some scribbles on a shared API and I'll create an issue to gather feedback on that later.
Anyways, I just took a look at the existing vacuum implementations for fan speeds:
Viomi: Silent, Standard, Medium, Turbo
Dreame: Quiet, Default, Medium, Strong
Rockrobo: Gentle (E2, v2 only?), Silent, Standard, Medium, Turbo
so it's quite a mess with names (and separate enums) but four different speeds is common so I think it'll be fairly simple to find a common base there when the time comes... :-)

miio/vacuum.py
fettlaus: Didn't see your comment before changing it as suggested previously. But you are right, in the future it is just a matter of deciding which generic names to choose for the different speeds. In pretty much every case it is: silent, default, more, max

miio/vacuum.py
```diff
@@ -747,6 +790,29 @@ def set_waterflow(self, waterflow: WaterFlow):
         """Set water flow setting."""
         return self.send("set_water_box_custom_mode", [waterflow.value])
 
+    @command()
+    def mop_mode(self) -> Optional[MopMode]:
+        """Get mop mode setting."""
+        try:
+            return MopMode(self.send("get_mop_mode")[0])
+        except ValueError:
+            return None
```
rytilahti, line 799: Please add a logger warning for unknown values, that way it doesn't go unnoticed.

miio/vacuumcontainers.py
```diff
@@ -175,6 +190,22 @@ def is_water_box_attached(self) -> bool:
...
+        if "water_box_carriage_status" in self.data:
+            return self.data["water_box_carriage_status"] == 1
+        return None
+
+    @property
+    def is_water_shortage(self) -> Optional[bool]:
+        """Returns True if water is low in the tank, None if sensor not present."""
+        if "water_shortage_status" in self.data:
+            return self.data["water_shortage_status"] == 1
+        else:
```
rytilahti, line 206: Unnecessary else.

miio/vacuum_cli.py
```diff
@@ -557,6 +561,25 @@ def carpet_mode(vac: miio.Vacuum, enabled=None):
...
 
+@cli.command()
+@click.argument("mode", required=False, type=str)
+@pass_dev
+def carpet_cleaning_mode(vac: miio.Vacuum, mode=None):
+    """Query or set the carpet cleaning/avoidance mode.
+
+    Allowed values: Avoid, Rise, Ignore
+    """
+    from miio.vacuum import CarpetCleaningMode
```
rytilahti, line 572: Move the import to the top of the file among other imports.

miio/vacuum_cli.py
```diff
@@ -557,6 +561,25 @@ def carpet_mode(vac: miio.Vacuum, enabled=None):
...
+
+    if mode is None:
+        click.echo("Carpet cleaning mode: %s" % vac.carpet_cleaning_mode().name)
```
rytilahti, line 575:
```suggestion
        click.echo("Carpet cleaning mode: %s" % vac.carpet_cleaning_mode())
```
Better simply use the direct response here, I didn't check if this can return none or not, but in case of none this would cause an exception.

miio/vacuum.py
```diff
@@ -681,6 +706,24 @@ def set_carpet_mode(
...
 
+    @command()
+    def carpet_cleaning_mode(self):
+        """Get carpet cleaning mode/avoidance setting."""
+        try:
+            return CarpetCleaningMode(
+                self.send("get_carpet_clean_mode")[0]["carpet_clean_mode"]
+            )
+        except TypeError:
+            return None
```
rytilahti, line 717: Please log the error here.

rytilahti (commented):
A couple of more comments, but otherwise this looks good to go. I'm unsure if the best action for unsupported actions is to raise an exception or simply to log it, but let's go for logging for now to make it simpler for downstreams.

The vacuum class (when the API gets consolidated) should get a `supported_features()` property at some point to allow introspecting which commands and features are available for the given model. Until then, I think it's better to (non-silently) signal for invalid usage instead of raising an exception.

miio/vacuum_cli.py
fettlaus: I explicitly didn't because that's the way it is handled in `reset_consumable()`. I don't mind moving the import to the top though

miio/vacuum.py
```diff
@@ -681,6 +706,25 @@ def set_carpet_mode(
         }
         return self.send("set_carpet_mode", [data])[0] == "ok"
 
+    @command()
+    def carpet_cleaning_mode(self):
```
rytilahti, line 710:
```suggestion
    def carpet_cleaning_mode(self) -> CarpetCleaningMode:
```

rytilahti (approved):
Let's merge it, thanks again for the PR and your patience! :1st_place_medal:

PR #1044, opened 2021-05-11
Add github flow for ci

rytilahti:
This PR converts away from azure pipelines to github actions which simplifies the setup & make coverage working again.

When that works, it will be much simpler to add a workflow for doing releases (incl. changelog generation, tagging, and publishing to pypi), hopefully. The aim is to make it releasing not a chore.


rytilahti:
This got stuck waiting for the pipeline checks that were not anymore part of it, even after removing those specific checks from the branch protection requirements. Removing the branch protection temporarily fixed this :-)

PR #1045, opened 2021-05-11
Added Roborock s7 to troubleshooting guide

Claustn:


Claustn:
This is to make people aware that the Roborock s7 needs to be on the same subnet as miiocli in order to communicate

rytilahti (approved):
Thanks! The CI errors are irrelevant for this documentation change, so let's just merge it :-)

Issue #1059, opened 2021-05-30
Common Vacuum Interface


stek29:
Currently this library provides at least three different implementations for Vacuums (roborock, dreame and viomi) which are incompatible.
As mentioned in home-assistant/core#49188, best way to get different Vacuums into other projects (like Home Assistant) is to provide some kind of common interface for vacuums.

I see three ways to do that:
1. create "abstract" interface for vacuums, then refactor existing vacuums to support that interface – might be backward incompatible?
2. create "abstract" intrerface for vacuums, create implementations of that interface for each vacuum type to keep backward compatibility; if any new types of vacuums are added, they should implement that interface directly
3. create common vacuum interface which acts as an adaptor for exact implementations: might be harder to maintain in future due to centralisation, but keeps backward compatibilty

Also, it'd be great to have some kind of "router" which automatically chooses correct implementation depending on device type – the easiest way to do that is by going with third way

What do you think? I'd be more than happy to make this happen, but it needs to be discussed and thought through first

Also, if I'm not mistaken, there's currently no common interfaces for any kind of device – we should keep that in mind when designing this, so solution is scalable to other kinds of devices.

rytilahti:
Hi, yes, I think it would be great to common interfaces per device type (similarly how homeassistant has `platforms`) but unfortunately I haven't had time to tinker with it and won't be able to do that in the near future either.

What I had in mind is to have a class hierarchy that defines the minimal common interface and to allow device-specific classes extend that if they wish to do so. IMO your second proposal is the way to go for a release or two, while marking the non-standardized APIs as deprecated so in case anyone is using them will get a notice before they get removed after some time.

Here is an example from some old notes I wrote down while looking into the vacuum classes while wondering about the common API:
```
class StatusContainer:
    """Common base-class for all status containers."""
    def is_on():
        """Returns if the device is on."""
    def got_error():
        """Returns if the device is in error state."""

class VacuumStatusContainer(StatusContainer):
    """Common base class for all vacuums."""
    def battery() -> int:
        """Battery charge state."""
    def state_code() -> int:
        """Code of the current vacuum state."""
    def state() -> str:
        """Human readable presentation of the state."""
    def error_code() -> Optional[int]:
        """Error code, if errored."""
    def error() -> Optional[str]:
        """Human readable presentation of the error."""
    def fanspeed() -> int:
        """Fan-speed.."""

class Device:
    def status() -> StatusContainer:
        """All devices should support status container."""

[...]

Issue #929, opened 2021-02-03
HomeAssistant Config Flow for Xiaomi Miio switch etc


starkillerOG:
I am considering writting a HomeAssistant PR to implement config flow for the rest of the Xiaomi Miio devices now supported in HomeAssistant, beginning with the Xiaomi Miio switches.

I don't own any other Xiaomi Miio devices besides the gateway and connected zigbee devices.
Therefore I would not be able to test the code, but I am pretty sure I can write it such that it will probably work.

Would this be appriciated?
@rytilahti what do you think?

syssi:
I can provide some testing!

rytilahti:
I can also provide some testing, and that sounds awesome!

Be aware though, that one of the big problems with current homeassistant implementation is they it'll require restructuring to move all sensors to be under the `sensor` domain instead of misusing the state attributes. There have been some PRs that haven't been accepted to homeassistant due to this.

starkillerOG:
@syssi, @rytilahti I just published my branch with the code for config flow of all xiaomi miio switches:
https://github.com/starkillerOG/home-assistant/tree/Miio_config_flow

If you own a xiaomi miio switch, could you try it for me?
If it works I will continue with the other platforms and make one big PR for HomeAssistant that converts all Miio Platforms to Config flow.

rytilahti:
@starkillerOG let me recommend the following:
* Do not do large PRs to homeassistant - reviewing them makes usually so much more time, and in case changes need to be done, it's much easier to do that at that point.
* Instead, create a PR to convert the switches to a flow and submit it. This will allow easier code reviews and testing. At the moment it's not an easy task to test your repository (where with PR it's just about `git pr <number> upstream` and the PR will be downloaded :-)
* After the switch conversion is done, the conversion can proceed with other devices.

starkillerOG:
Thanks for the comments @rytilahti, sounds good.
I will make the PR for the switches now.

At this point I think the config flow is pretty straight forward, I don't think any changes to the upstream lib are nessesary.
However as I get revieuw comments from the HomeAssistant team, they are most likely going to want to change stuff because it is outdated that will require me to make changes to the upstream lib, but I would not know yet what that would be.

starkillerOG:
@rytilahti one thing I would like to do is make a cache for the device info:
https://github.com/rytilahti/python-miio/blob/91ecc98783d6fc035a49e8dd5a254e9ce0100c9e/miio/device.py#L182
[...]

PR #1060, opened 2021-05-31
WIP: General Vacuum Interface (#1059)

stek29:
WIP: suggestions and corrections are welcome :)

todo:
- [ ] Roborock implementation
- [ ] Consumables (+ resetting)
- [ ] Mopping/Water
- [ ] Feature flags
- [ ] Clean History
- [ ] Execute Custom Action (device specific)

Would be nice to add:
- [ ] Stats
- [ ] RC/Manual control
- [ ] DND/Timers
- [ ] Audio Control (Volume, Voice Packs)
- [ ] Viomi Implementation

### DIFF
--- a/miio/vaxuum/__init__.py
+++ b/miio/vaxuum/__init__.py
@@ -1,0 +1,2 @@
+from . import base
+from . import dreame
--- a/miio/vaxuum/base/__init__.py
+++ b/miio/vaxuum/base/__init__.py
@@ -1,0 +1,11 @@
+""" Common Vacuum Interface """
+
+from .vacuum import BaseVacuum
+from .containers import (
+    VacuumState,
+    CleaningType,
+    ChargeStatus,
+    FanSpeed,
+    VacuumStatus,
+    ConsumableStatus,
+)
--- a/miio/vaxuum/base/containers.py
+++ b/miio/vaxuum/base/containers.py
@@ -1,0 +1,87 @@
+import enum
+
+from miio.device import DeviceStatus
+from typing import Optional
+
+
+class VacuumState(enum.Enum):
+    UNKNOWN = 0
+    # Cleaning in progress
+    CLEANING = 1
+    # Doing nothing
+    IDLE = 2
+    # Cleaning is paused
+    PAUSED = 3
+    ERROR = 4
+    RETURNING_TO_DOCK = 5
+    DOCKED = 6
+    # MANUAL = 7
+
+
+class CleaningType(enum.Enum):
+    UNKNOWN = 0
+    NORMAL = 1
+    ZONE = 2
+    SEGMENT = 3
+
+
+class ChargeStatus(enum.Enum):
+    UNKNOWN = 0
+
+    # is currently charging
+    CHARGING = 1
+    # is currently discharging
+    DISCHARGING = 2
+    # battery is full
+    FULL = 3
+    # on the way to dock to charge
+    GO_CHARGING = 4
+
+
+class FanSpeed(enum.Enum):
+    """Human readable vacuum fan speed presets
+
+    If mode is not supported on current model, closest should be used
+      (for example, SILENT instead of GENTLE)
+    If mode is unknown, set to STANDARD
+    """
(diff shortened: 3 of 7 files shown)


stek29:
@rytilahti there's also another option – using universal MixIns, kinda like traits in Google Assistant/Home

That'd be more universal: there would be "MixIn"s for battery status, start/stop, pause/resume, setting fan speed, which can be reused for other kinds of devices
If designed corretly, mixins might also cover feature detection/flags

but I'm not quite sure if it's worth the effort for now

stek29:
Valetudo also has capabilities based design, see https://valetudo.cloud/pages/general/capabilities-overview.html

stek29:
@rytilahti any thoughts?

rytilahti:
Hey @stek29 and thanks for the PR! I'm currently very busy with other things and cannot currently do a full review, but did you take a look at my comment related to the directory structure? IMO the base classes should be inside the main package, and the implementations moved out to their own directories.

On the mixins, I feel that they are not very pythonic, and will likely just cause more confusion than they are worth. For example: if you are implementing support for new device, you either must search for the wanted mixins (or sadly, simply copy&paste them over from something else). So I think it's better to have explicit interfaces (e.g., using abstract classes) that are directly useful if you are using a proper IDE for development.

When it will get quieter at some point late next week, I'll write some more comments!

starkillerOG:
@stek29 I have also been thinking about how to generalize the interface of python-miio, see https://github.com/rytilahti/python-miio/issues/929#issuecomment-787442355.

In my opinion there schould be one common .update() function that all integrations use to update all available data.
The .update() function can be defined in device.py and overwritten by each device class to implement there own specific update calls.
The resulting data also needs to be stored in a common fasion, preferably there would be a .status() function (again in the device.py) that is overwritten by each implementation that returns all available data.

I think most of the current implementations use info classes to represent the data, so the .status() could return that info class. preferably there would also be a .status_dict() function that would return the status as a dictionarry.

In that way implementations as HomeAssistant can use common functions to extract common device status attributes like "temperature".

I already did this for the subdevices of the xiaomi gateway where I define the available properties in a .yaml file:
https://github.com/rytilahti/python-miio/blob/master/miio/gateway/devices/subdevices.yaml

HomeAssistant uses the common attributes to easily setup sensor entities:
https://github.com/home-assistant/core/blob/12f2482c9bb062f347cd1e80a68e9d0d469f897a/homeassistant/components/xiaomi_miio/sensor.py#L76-L98
https://github.com/home-assistant/core/blob/12f2482c9bb062f347cd1e80a68e9d0d469f897a/homeassistant/components/xiaomi_miio/sensor.py#L137-L147

rytilahti:
Hey @starkillerOG, in most of the cases there exists a `status()` that is used to report the device state, so in my mind that is sort of already a part of the common device API even when it is not defined in the device class. I think this should change though, and the base device API could at least define that (and maybe even `on()` and `off()` for basic controls?).

Now, on how to expose the separate sensors is an open question, but we should find a way to do that. Introducing a separate decorator (instead of using pure `@property`) to allow defining the sensory properties would make sense. The base class can implement a way to get a list of these (as well as implement `as_dict()` for accessing all properties), and downstreams like homeassistant can simply iterate over them when creating and updating the sensor states. This will probably need some experimentation to get a feeling what makes most sense and feels natural...

See also my (rather preliminary) comment about the general (re)structure here https://github.com/rytilahti/python-miio/issues/1059#issuecomment-851070398 . It's a quite a bit of work to do such restructuring without breaking existing use cases, but it will hopefully be worth it at the end.

This should also make it simpler to add support for generic device-type specific miot devices (@ha0y has done great work on creating a custom HA component, but not everyone uses HA, this comment from him/her is relevant on the topic https://github.com/rytilahti/python-miio/pull/1029#issuecomment-834112191). While I'm reluctant to add any sort of cloud requirement to this library, we could still provide a way to use already downloaded miot spec files for generic device support.

stek29:
sadly I'm no longer interested in this since valetudo is now available for Dreame robots

PR #1063, opened 2021-06-02
add fan speed enum 106 as "Auto" for Roborock S6 MaxV

RubenKelevra:
If the room customization function for the fan speeds is used, the Roborock S6 MaxV reports a fan speed of 106 back.

### DIFF
--- a/miio/vacuum.py
+++ b/miio/vacuum.py
@@ -66,6 +66,7 @@ class FanspeedV2(enum.Enum):
     Medium = 103
     Turbo = 104
     Gentle = 105
+    Auto = 106
 
 
 class FanspeedV3(enum.Enum):


RubenKelevra:
Hey @rytilahti can you have a look at this plus the [corresponding Home Assistant test](https://github.com/home-assistant/core/pull/32850) you committed?

Not sure if that's correct what I did. Just noticed the 106 reported back and thought I do something about it :D

rytilahti:
Hi @RubenKelevra and thanks for the PR!

What does the app say about fan speed for this? And what happens, if you try to set the fanspeed to this 106/auto manually?

RubenKelevra:
Well, the app calls it "Customize" which makes no sense.









When I send say "Standard" to the robot, it acts accordingly to get out of 106 (confirming this with a "ping"). But sending 106 to it doesn't work. It will "ping" as confirmation but won't switch to "Auto" anymore.

That said, it will also ping for a 107, so I think there's no confirmation about the valid range going on before "pinging".

To enable the automation we probably need to send a different command, since it will also affect the water usage.

---

Regarding water usage, can I help in any way to get this going? 🤔

rytilahti (approved):

rytilahti:
Thanks for the clarifications, let's merge this now!

> To enable the automation we probably need to send a different command, since it will also affect the water usage.
> 
> Regarding water usage, can I help in any way to get this going? thinking

Unfortunately I don't have any idea how any of that works, I'm still on trusty old gen1 vacuum. But what you could do is to do a traffic capture and find out what commands are being sent to the device :-)

PR #1070, opened 2021-06-09
airpurifier_miot: Move favorite_rpm from MB4 to Basic

SylvainPer:
favorite_rpm is also a 3H parameter so it should be in the basic part

Tested on my local setup, I wasn't able to read this parameter before the modification with the status() function.

### DIFF
--- a/miio/airpurifier_miot.py
+++ b/miio/airpurifier_miot.py
@@ -139,7 +139,7 @@ class BasicAirPurifierMiotStatus(DeviceStatus):
     def motor_speed(self) -> int:
         """Speed of the motor."""
         return self.data["motor_speed"]
-    
+
     @property
     def favorite_rpm(self) -> Optional[int]:
         """Return favorite rpm level."""
@@ -296,7 +296,6 @@ class AirPurifierMB4Status(BasicAirPurifierMiotStatus):
     def led_brightness_level(self) -> int:
         """Return brightness level."""
         return self.data["led_brightness_level"]
-
 
 
 class BasicAirPurifierMiot(MiotDevice):
--- a/miio/airpurifier_miot.py
+++ b/miio/airpurifier_miot.py
@@ -139,6 +139,11 @@ class BasicAirPurifierMiotStatus(DeviceStatus):
     def motor_speed(self) -> int:
         """Speed of the motor."""
         return self.data["motor_speed"]
+    
+    @property
+    def favorite_rpm(self) -> int:
+        """Return favorite rpm level."""
+        return self.data["favorite_rpm"]
 
 
 class AirPurifierMiotStatus(BasicAirPurifierMiotStatus):
@@ -292,10 +297,6 @@ class AirPurifierMB4Status(BasicAirPurifierMiotStatus):
         """Return brightness level."""
         return self.data["led_brightness_level"]
 
-    @property
-    def favorite_rpm(self) -> int:
-        """Return favorite rpm level."""
-        return self.data["favorite_rpm"]
 
 
 class BasicAirPurifierMiot(MiotDevice):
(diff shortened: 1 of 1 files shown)


SylvainPer:
Sorry but I'm not familiar with this check and I don't understand why it failed.

rytilahti:
The error came from black (so code formatting, likely some whitespace changes). The simplest way to fix is to run `pre-commit run -a` and it will correct it for you.

rytilahti (changes requested):
Thanks for the PR! As it is unclear which devices support this, let's play it safe and avoid crashing just in case that property is not available on some devices.

rytilahti:
Please run `pre-commit run -a` in your local copy to fix the linting errors, then this is good to go :-)

SylvainPer:
yes, it's on going:
pi@raspberrypi:~/temp/miio/python-miio $ pre-commit run -a
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-isort.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-isort:toml.
[INFO] Initializing environment for https://github.com/PyCQA/doc8.
[INFO] Initializing environment for https://github.com/myint/docformatter.
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8:flake8-annotations,flake8-bugbear,flake8-builtins,flake8-docstrings,flake8-print,flake8-pytest-style,flake8-return,flake8-simplify.
[INFO] Initializing environment for https://github.com/PyCQA/bandit.
[INFO] Initializing environment for https://github.com/pre-commit/mirrors-mypy.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...

SylvainPer:
Done for the formatting but I had this at the end:

devtools/containers.py:61: error: Incompatible types in assignment (expression has type "List[_T]", variable has type "Optional[List[Dict[str, Any]]]")
Found 1 error in 1 file (checked 127 source files)

rytilahti:
That is probably dependent on your mypy version; I don't get any errors on devtools locally and the CI passed just fine :-)

rytilahti (approved):

SylvainPer:
Hello @rytilahti ,
Any idea when the next version will be released ?
Is there a way to know which library version is installed ?
I'll update my plugin when the next version will be available but I'll get an error on oldest version.
Thanks

rytilahti:
Hi @SylvainPer, I'll make a release in a couple of weeks when I return from the holidays, if @syssi doesn't beat me.

Wrt. checking the library version, you can use the standard `__version__` property to access the version string:
```
>>> import miio
>>> miio.__version__
'0.5.5.2'
```
There is some module (maybe even in stdlib?) that allows parsing the string to something that can be used in sane comparisons, but I cannot recall the name of that.

syssi:
@rytilahti Feel free to release as soon as possible. I'm at holidays too and cannot provide much support at the moment.

SylvainPer:
Thanks for your return, I'm also at holidays now so enjoy them !

SylvainPer:
> There is some module (maybe even in stdlib?) that allows parsing the string to something that can be used in sane comparisons, but I cannot recall the name of that.

found this:
`>>> from packaging import version
>>> version.parse("2.3.1") < version.parse("10.1.2")`

PR #867, opened 2020-11-23
Support for Xiaomi Mijia G1 (mijia.vacuum.v2) 

neturmel:
just a quick and dirty hack based on airhumidifier_miot and dreame.vacuum.mc1808. 
Only some commands are working. But you get some status report and starting and stopping ist possible.
Didn't bother to try it with the newer "get_properties_by_dataclass".

PR #1072, opened 2021-06-12
Added support for Roidmi Eve

martin9000andersen:
I have added support for robot Roidmi Eve (Robot vacuum cleaner).

Many status values are implemented and many features are implemented e.g.:
    start-clean, stop, go-home, set_sound_volume, set_cleaning_mode, set_sweep_type, set_work_station_freq, start_dust, etc.

I do not know the syntax to activate cleaning of specific rooms, any input on how to make the function "start_room_sweep_unknown" work is highly appreciated.

### DIFF
--- a/miio/roidmivacuum_miot.py
+++ b/miio/roidmivacuum_miot.py
@@ -48,13 +48,13 @@ _MAPPING: MiotMapping = {
     "clean_counts": {"siid": 8, "piid": 18},
     "clean_time_sec": {"siid": 8, "piid": 19},
     "double_clean": {"siid": 8, "piid": 20},
-    "edge_sweep": {"siid": 8, "piid": 21},
+    # "edge_sweep": {"siid": 8, "piid": 21}, # 2021-07-11: Roidmi Eve is not changing behavior when this bool is changed
     "led_switch": {"siid": 8, "piid": 22},
     "lidar_collision": {"siid": 8, "piid": 23},
     "station_key": {"siid": 8, "piid": 24},
     "station_led": {"siid": 8, "piid": 25},
     "current_audio": {"siid": 8, "piid": 26},
-    "progress": {"siid": 8, "piid": 28},
+    # "progress": {"siid": 8, "piid": 28}, # 2021-07-11: this is part of the spec, but not implemented in Roidme Eve
     "station_type": {"siid": 8, "piid": 29},  # uint32
     # "voice_conf": {"siid": 8, "piid": 30}, # Always return file not exist !!!
     # "switch_status": {"siid": 2, "piid": 10}, # Enum with only one value: Open
@@ -207,13 +207,11 @@ class RoidmiVacuumStatus(DeviceStatus):
             {'did': 'clean_counts', 'siid': 8, 'piid': 18},
             {'did': 'clean_time_sec', 'siid': 8, 'piid': 19},
             {'did': 'double_clean', 'siid': 8, 'piid': 20},
-            {'did': 'edge_sweep', 'siid': 8, 'piid': 21},
             {'did': 'led_switch', 'siid': 8, 'piid': 22}
             {'did': 'lidar_collision', 'siid': 8, 'piid': 23},
             {'did': 'station_key', 'siid': 8, 'piid': 24},
             {'did': 'station_led', 'siid': 8, 'piid': 25},
             {'did': 'current_audio', 'siid': 8, 'piid': 26},
-            {'did': 'progress', 'siid': 8, 'piid': 28},
             {'did': 'station_type', 'siid': 8, 'piid': 29},
             {'did': 'volume', 'siid': 9, 'piid': 1},
             {'did': 'mute', 'siid': 9, 'piid': 2}
@@ -286,13 +284,16 @@ class RoidmiVacuumStatus(DeviceStatus):
             return PathMode.Unknown
 
     @property
-    def mop_present(self) -> bool:
-        """Return True is mop (water box) is installed."""
+    def is_mop_attached(self) -> bool:
+        """Return True if mop is attached."""
         return self.data["mop_present"]
 
     @property
-    def work_station_frequency(self) -> int:
-        """work_station_freq (2 means base dust colect every second time)."""
+    def dust_collection_frequency(self) -> int:
+        """Frequency for emptying the dust bin.
+
+        Example: 2 means the dust bin is emptied every second cleaning.
+        """
         return self.data["work_station_freq"]
 
     @property
@@ -367,13 +368,8 @@ class RoidmiVacuumStatus(DeviceStatus):
         return self.data["double_clean"]
 
     @property
-    def edge_sweep(self) -> bool:
-        """Is edge clean enabled."""
-        return self.data["edge_sweep"]
-
-    @property
-    def vacuum_led(self) -> bool:
-        """Return if led/display on vaccum is on."""
+    def led(self) -> bool:
+        """Return True if led/display on vaccum is on."""
         return self.data["led_switch"]
 
     @property
@@ -409,11 +405,6 @@ class RoidmiVacuumStatus(DeviceStatus):
     def clean_area(self) -> int:
         """Cleaned area in m2."""
         return self.data["clean_area"]
-
-    @property
-    def progress(self) -> str:
-        """Always return None?"""
-        return self.data["progress"]
 
     @property
     def state_code(self) -> int:
@@ -584,24 +575,24 @@ class RoidmiVacuumMiot(MiotDevice):
         """Start cleaning."""
         return self.call_action("start")
 
-    @command(click.argument("roomstr", type=str))
-    def start_room_sweep_unknown(self, roomstr: str) -> None:
-        """Start room cleaning.
+    # @command(click.argument("roomstr", type=str, required=False))
+    # def start_room_sweep_unknown(self, roomstr: str=None) -> None:
+    #     """Start room cleaning.
 
-        FIXME: the syntax of room_sweep is unknown
-        """
-        return self.call_action("start_room_sweep")
+    #     roomstr: empty means start room clean of all rooms. FIXME: the syntax of an non-empty roomstr is still unknown
+    #     """
+    #     return self.call_action("start_room_sweep", roomstr)
 
-    @command(
-        click.argument("sweep_mode", type=EnumType(SweepMode)),
-        click.argument("clean_info", type=str),
-    )
-    def start_sweep_unknown(self, sweep_mode: SweepMode, clean_info: str) -> None:
-        """Start sweep with mode.
+    # @command(
+    # click.argument("sweep_mode", type=EnumType(SweepMode)),
+    # click.argument("clean_info", type=str),
+    # )
+    # def start_sweep_unknown(self, sweep_mode: SweepMode, clean_info: str=None) -> None:
+    #     """Start sweep with mode.
 
-        FIXME: the syntax of start sweep with mode is unknown
-        """
-        return self.call_action("start_sweep", [sweep_mode.value, clean_info])
+    #     FIXME: the syntax of start_sweep is unknown
+    #     """
+    #     return self.call_action("start_sweep", [sweep_mode.value, clean_info])
 
     @command()
     def stop(self) -> None:
@@ -624,7 +615,7 @@ class RoidmiVacuumMiot(MiotDevice):
         return self.set_property("station_led", on)
 
     @command(click.argument("on", type=bool))
-    def set_vacuum_led(self, on: bool):
+    def set_led(self, on: bool):
         """Enable vacuum led."""
         return self.set_property("led_switch", on)
 
@@ -653,10 +644,13 @@ class RoidmiVacuumMiot(MiotDevice):
         """Set path_mode."""
         return self.set_property("path_mode", path_mode.value)
 
-    @command(click.argument("work_station_freq", type=int))
-    def work_station_frequency(self, work_station_freq: int):
-        """Set work_station_freq (2 means base dust colect every second time)."""
-        return self.set_property("work_station_freq", work_station_freq)
+    @command(click.argument("dust_collection_frequency", type=int))
+    def set_dust_collection_frequency(self, dust_collection_frequency: int):
+        """Set frequency for emptying the dust bin.
+
+        Example: 2 means the dust bin is emptied every second cleaning.
+        """
+        return self.set_property("work_station_freq", dust_collection_frequency)
 
     @command(click.argument("timing", type=str))
     def set_timing(self, timing: str):
@@ -724,11 +718,6 @@ class RoidmiVacuumMiot(MiotDevice):
         """Set double clean (True/False)."""
         return self.set_property("double_clean", double_clean)
 
-    @command(click.argument("edge_sweep", type=bool))
-    def set_edge_sweep(self, edge_sweep: bool):
-        """Set edge_sweep (True/False)."""
-        return self.set_property("edge_sweep", edge_sweep)
-
     @command(click.argument("lidar_collision", type=bool))
     def set_lidar_collision_sensor(self, lidar_collision: bool):
         """When ON, the robot will use lidar as the main detection sensor to help reduce
@@ -740,13 +729,13 @@ class RoidmiVacuumMiot(MiotDevice):
         """Start base dust collection."""
         return self.call_action("start_station_dust_collection")
 
-    @command(click.argument("voice", type=str))
-    def set_voice_unknown(self, voice: str) -> None:
-        """Set voice.
+    # @command(click.argument("voice", type=str))
+    #     def set_voice_unknown(self, voice: str) -> None:
+    #     """Set voice.
 
-        FIXME: the syntax of voice is unknown (assumed to be json format)
-        """
-        return self.call_action("set_voice", voice)
+    #     FIXME: the syntax of voice is unknown (assumed to be json format)
+    #     """
+    #     return self.call_action("set_voice", voice)
 
     @command()
     def reset_filter_life(self) -> None:
--- a/miio/roidmivacuum_miot.py
+++ b/miio/roidmivacuum_miot.py
@@ -1,60 +1,63 @@
 """Vacuum Eve Plus (roidmi.vacuum.v60)"""
 
-# https://github.com/rytilahti/python-miio/issues/543#issuecomment-755767331
 
 import json
 import logging
 import math
+from datetime import timedelta
 from enum import Enum
 
 import click
 
-from .click_common import EnumType, command, format_output
-from .miot_device import DeviceStatus as DeviceStatusContainer
-from .miot_device import MiotDevice, MiotMapping
+from .click_common import EnumType, command
+from .miot_device import DeviceStatus, MiotDevice, MiotMapping
+from .vacuumcontainers import DNDStatus
 
 _LOGGER = logging.getLogger(__name__)
 
 _MAPPING: MiotMapping = {
     "battery_level": {"siid": 3, "piid": 1},
     "charging_state": {"siid": 3, "piid": 2},
-    "device_fault": {"siid": 2, "piid": 2},
-    "device_status": {"siid": 2, "piid": 1},
+    "error_code": {"siid": 2, "piid": 2},
+    "state": {"siid": 2, "piid": 1},
     "filter_life_level": {"siid": 10, "piid": 1},
-    "filter_left_time": {"siid": 10, "piid": 2},
-    "brush_left_time": {"siid": 11, "piid": 1},
-    "brush_life_level": {"siid": 11, "piid": 2},
-    "brush_left_time2": {"siid": 12, "piid": 1},
-    "brush_life_level2": {"siid": 12, "piid": 2},
-    "brush_left_time3": {"siid": 15, "piid": 1},
-    "brush_life_level3": {"siid": 15, "piid": 2},
+    "filter_left_minutes": {"siid": 10, "piid": 2},
+    "main_brush_left_minutes": {"siid": 11, "piid": 1},
+    "main_brush_life_level": {"siid": 11, "piid": 2},
+    "side_brushes_left_minutes": {"siid": 12, "piid": 1},
+    "side_brushes_life_level": {"siid": 12, "piid": 2},
+    "sensor_dirty_time_left_minutes": {
+        "siid": 15,
+        "piid": 1,
+    },  # named brush_left_time in the spec
+    "sensor_dirty_remaning_level": {"siid": 15, "piid": 2},
     "sweep_mode": {"siid": 14, "piid": 1},
-    "cleaning_mode": {"siid": 2, "piid": 4},
+    "fanspeed_mode": {"siid": 2, "piid": 4},
     "sweep_type": {"siid": 2, "piid": 8},
     "path_mode": {"siid": 13, "piid": 8},
     "mop_present": {"siid": 8, "piid": 1},
     "work_station_freq": {"siid": 8, "piid": 2},  # Range: [0, 3, 1]
-    "timing": {"siid": 8, "piid": 6},  # str (example: {"tz":2,"tzs":7200})
+    "timing": {"siid": 8, "piid": 6},
     "clean_area": {"siid": 8, "piid": 7},  # uint32
-    # "uid": {"siid": 8, "piid": 8}, # str
-    "auto_boost": {"siid": 8, "piid": 9},  # bool (auto boost on carpet)
+    "uid": {"siid": 8, "piid": 8},  # str
+    "auto_boost": {"siid": 8, "piid": 9},
     "forbid_mode": {"siid": 8, "piid": 10},  # str
     "water_level": {"siid": 8, "piid": 11},
-    # "siid8_13": {"siid": 8, "piid": 13}, # no-name: (uint32, unit: seconds) (acc: ['read', 'notify'])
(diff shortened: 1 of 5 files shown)


miio/device.py
```diff
@@ -71,30 +71,22 @@ def accesspoint(self):
     @property
     def model(self) -> Optional[str]:
         """Model string if available."""
-        if self.data["model"] is not None:
-            return self.data["model"]
-        return None
+        return self.data.get("model")
```
rytilahti, line 74: Please rebase on top of the current master, #1083 should fix the device info problems.

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
+"""Vacuum Eve Plus (roidmi.vacuum.v60)"""
+
+# https://github.com/rytilahti/python-miio/issues/543#issuecomment-755767331
```
rytilahti, line 3:
```suggestion
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+    GoCharging = 5
+    Charging = 6
+    Error = 7
+    Rfctrl = 8
+    Fullcharge = 9
+    Shutdown = 10
+    FindChargerPause = 11
+
+
+class RoidmiVacuumStatus(DeviceStatusContainer):
```
rytilahti, line 165:
```suggestion
class RoidmiVacuumStatus(DeviceStatus):
```
please extend this directly from the DeviceStatus for now. Also, please use the names and return types consistent to vacuumcontainers.py (e.g., `ConsumableStatus` for consumables), as we want to move slowly towards common interfaces among device types for interoperability.

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+    def __init__(self, data):
+        self.data = data
+
+    @property
+    def battery_level(self) -> int:
+        return self.data["battery_level"]
+
+    @property
+    def filter_left_time(self) -> int:
```
rytilahti, line 174: See my comment above, filter times should use timedeltas.

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+            "Volume: {result.volume}\n" "Mute: {result.mute}\n",
+        )
+    )
+    def status(self) -> RoidmiVacuumStatus:
+        """State of the vacuum."""
+
+        return RoidmiVacuumStatus(
+            {
+                prop["did"]: prop["value"] if prop["code"] == 0 else None
+                # max_properties limmit to 10 to avoid "Checksum error" messages from the device.
```
rytilahti, line 437: This might be fixed by the recent #1075 - could you please confirm if that's the case?

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+        return self.data["progress"]
+
+    @property
+    def voice_conf(self) -> str:
+        return self.data["voice_conf"]
+
+    @property
+    def switch_status(self) -> SwitchStatus:
+        try:
+            return SwitchStatus(self.data["switch_status"])
```
rytilahti, line 359: This doesn't currently seem to be queried at all. What does this indicate?

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+        except TypeError:
+            _LOGGER.error("Unknown SwitchStatus (%s)", self.data["switch_status"])
+            return SwitchStatus.Unknown
+
+    @property
+    def device_status(self) -> DeviceStatus:
```
rytilahti, line 365:
```suggestion
    def status(self) -> DeviceStatus:
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+    Curpoint = 3
+    Point = 4
+    Smart = 7
+    AmartArea = 8
+    DepthTotal = 9
+    AlongWall = 10
+    Idle = 0
+
+
+class FaultStatus(Enum):
```
rytilahti, line 120: It is probably simpler to have this as a dictionary than as an enum (the value being a user-friendly message, see https://github.com/rytilahti/python-miio/blob/master/miio/vacuumcontainers.py#L16).

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+
+    @property
+    def uid(self) -> int:
+        return self.data["uid"]
+
+    @property
+    def auto_boost(self) -> int:
+        return self.data["auto_boost"]
+
+    def parseForbidMode(self, val):
```
rytilahti, line 277: Internal methods should be marked as such. Also, these could probably be moved to be in the scope of `forbid_mode`?

```suggestion
    def _parse_forbid_mode(self, val):
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+        def secToClock(val):
```
rytilahti, line 278:
```suggestion
        def _secToClock(val):
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+            minut = math.floor((val - hour * 3600) / 60)
+            return "{}:{:02}".format(hour, minut)
+
+        asDict = json.loads(val)
+        active = bool(asDict["time"][2])
+        begin = secToClock(asDict["time"][0])
+        end = secToClock(asDict["time"][1])
+        return json.dumps(
+            {"enabled": active, "begin": begin, "end": end, "tz": asDict["tz"]}
+        )
```
rytilahti, line 289: Instead of using json dumps, would it be possible to use the `DNDStatus` from `vacuumcontainers.py` for this? 

If not, please create a separate container class (derived from `DeviceStatus`).

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+            _LOGGER.error("Unknown WaterLevel (%s)", self.data["water_level"])
+            return WaterLevel.Unknown
+
+    # @property
+    # def siid8_13(self) -> int:
+    #     return self.data["siid8_13"]
+
+    # @property
+    # def siid8_14(self) -> int:
+    #     return self.data["siid8_14"]
```
rytilahti, line 310:
```suggestion
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+
+    @property
+    def clean_counts(self) -> int:
+        return self.data["clean_counts"]
+
+    # @property
+    # def siid8_19(self) -> int:
+    #     return self.data["siid8_19"]
```
rytilahti, line 318:
```suggestion
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,590 @@
...
+    def set_auto_boost(self, auto_boost: bool):
+        """Set auto boost on carpet."""
+        return self.set_property("auto_boost", auto_boost)
+
+    @command(
+        click.argument("begin", type=str),
+        click.argument("end", type=str),
+        click.argument("active", type=bool, required=False, default=True),
+    )
+    def set_forbid_mode(self, begin: str, end: str, active: bool = True):
```
rytilahti, line 524:
```suggestion
    def set_dnd(self, start_hr: int, start_min: int, end_hr: int, end_min: int, active: bool = True)
```
This would make it similar to vacuum.py's `set_dnd` with the added `active` flag, for future proofing :-)

rytilahti (changes requested):
Thanks for the PR! I did a first round of reviews and left some specific comments inline, but on generla level:
* Please add descriptive docstrings to all properties and methods belonging to the public API
* Remove commented out code (except for the variable mappings, in case someone wants to look into those later)
* Please add (at least) `is_on`, `is_paused`, `got_error` from https://github.com/rytilahti/python-miio/blob/master/miio/vacuumcontainers.py#L125 - the long term goal is to make all vacuum implementations to have a common interface, so this will be useful for that.

martin9000andersen:
Thank you very much for your comments. I really like your long term goal of a common interface.
I will have a look at it in the coming days.

miio/device.py
martin9000andersen: Yes, that commit fixed the problem

miio/roidmivacuum_miot.py
martin9000andersen: For Roidmi Eve there are two reasons for getting "checksum error": 1) is too many properties at a time (here I use max_properties), 2) Too many properties in total.
Commit #1075 did not affect No. 1, so I need to keep max_properties, but it did solved No. 2 (This means that I can uncomment all properties in MiotMapping). Thank you very much for telling me about this fix.

miio/roidmivacuum_miot.py
martin9000andersen: I am not sure. It is an Enum and can only have the value "Open". I will remove it from the public interface, and only leave an comment about this in MiotMapping.

martin9000andersen:
I have updated the code to be more like vacuumcontainers.py (before it was closer to dreamevacuum_miot.py).

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,702 @@
...
+            _LOGGER.error("Unknown PathMode (%s)", self.data["path_mode"])
+            return PathMode.Unknown
+
+    @property
+    def mop_present(self) -> bool:
+        """Return True is mop (water box) is installed."""
+        return self.data["mop_present"]
+
+    @property
+    def work_station_freq(self) -> int:
```
rytilahti, line 238:
```suggestion
    def work_station_frequency(self) -> int:
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,702 @@
...
+        """Is edge clean enabled."""
+        return self.data["edge_sweep"]
+
+    @property
+    def led_switch(self) -> bool:
+        """The LED on the robot will be always on."""
+        return self.data["led_switch"]
+
+    @property
+    def lidar_collision(self) -> bool:
```
rytilahti, line 329:
```suggestion
    def lidar_enabled(self) -> bool:
```
Maybe?

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,702 @@
...
+        """Set path_mode."""
+        return self.set_property("path_mode", path_mode.value)
+
+    @command(click.argument("work_station_freq", type=int))
+    def set_work_station_freq(self, work_station_freq: int):
+        """Set work_station_freq (2 means base dust colect every second time)."""
+        return self.set_property("work_station_freq", work_station_freq)
+
+    @command(click.argument("timing", type=str))
+    def set_timing_unknown(self, timing: str):
```
rytilahti, line 596:
```suggestion
    def set_timing(self, timing: str):
```
I think there is no need to describe this as unknown?

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,702 @@
...
+        Set timing to 9:00 Monday-Friday, rooms:[12,10]
+        timing = '{"time":[[32400,1,3,0,[1,2,3,4,5],0,[12,10],null]],"tz":2,"tzs":7200}'
+        See also :func:`RoidmiVacuumStatus.timing`
+
+        NOTE: setting timing will override existing settings
+        """
+        return self.set_property("timing", timing)
+
+    @command(click.argument("auto_boost", type=bool))
+    def set_auto_boost(self, auto_boost: bool):
```
rytilahti, line 608:
```suggestion
    def set_carpet_mode(self, auto_boost: bool):
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,702 @@
...
+    def start_dust(self) -> None:
+        """Start base dust collection."""
+        return self.call_action_by(8, 6)
+
+    @command(click.argument("voice", type=str))
+    def set_voice_unknown(self, voice: str) -> None:
+        """Set voice.
+
+        FIXME: the syntax of voice is unknown
+        """
```
rytilahti, line 681: Could this be the name of the setting, e.g., "girl_en"?

miio/tests/test_roidmivacuum_miot.py
```diff
@@ -0,0 +1,188 @@
...
+
+@pytest.mark.usefixtures("dummyroidmivacuum")
+class TestRoidmiVacuum(TestCase):
+    def test_VacuumStatus(self):
+        status = self.device.status()
+        assert status.auto_boost == _INITIAL_STATE["auto_boost"]
+        assert status.battery == _INITIAL_STATE["battery_level"]
+        assertEnum(
+            status.charging_state, ChargingState(_INITIAL_STATE["charging_state"])
+        )
```
rytilahti, line 87:
```suggestion
        assert status.charging_state == ChargingState(_INITIAL_STATE["charging_state"])
```
should be enough for all enum checks. If they are equal, their `repr()`s are also equal.

miio/tests/test_roidmivacuum_miot.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+@pytest.fixture(scope="function")
+def dummyroidmivacuum(request):
+    request.cls.device = DummyRoidmiVacuumMiot()
+
+
+def assertEnum(a, b):
+    assert a == b
+    assert repr(a) == repr(b)
```
rytilahti, line 76:
```suggestion
```
This is unnecessary.

miio/tests/test_roidmivacuum_miot.py
```diff
@@ -0,0 +1,188 @@
...
+    def test_RoidmiCleaningSummary(self):
+        status = self.device.cleaning_summary()
+        assert (
+            status.total_duration.total_seconds()
+            == _INITIAL_STATE["total_clean_time_sec"]
+        )
+        assert status.total_area == _INITIAL_STATE["total_clean_areas"]
+        assert status.count == _INITIAL_STATE["clean_counts"]
+
+    def test_RoidmiConsumableStatus(self):
```
rytilahti, line 122:
```suggestion
    def test_consumable_status(self):
```
Avoid `CamelCasing`, it is also unnecessary to repeat "Roidmi" in tests.

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,702 @@
...
+        return self.call_action_by(8, 12, voice)
+
+    @command()
+    def reset_filter_life(self) -> None:
+        """Reset filter life."""
+        return self.call_action_by(10, 1)
```
rytilahti, line 687: Please convert the actions to use `call_action(str)` and declare the actions in the mapping like: https://github.com/rytilahti/python-miio/pull/867/files#diff-2f2ea69d0a6bf409d623cea17e4744ce7b60739d4c29b79166a8c1979660bfb3R36-R42

The `call_action_by` is rather targeted for `miiocli` usage where there is no known symbol for it available.

rytilahti (commented):
I added some more comments, I think it's good to go after this round but be aware that the API will change for the next major release to allow consolidating the common functionalities among all supported vacuums :-)

While doing the review I did some googling and started to wonder if this PR is for "Roidmi Eve Plus" or is there also a version without the "plus"? If that's the same one, this might be one option when it's time to upgrade away from my gen1 rockrobo.. What's your opinion about the device?

miio/roidmivacuum_miot.py
martin9000andersen: Text from the Mi App: "Priority use of radar - When turned on, the robot will use radar as the main detection sensor to help reduce collisions".
When this option is OFF, lidar is still used for mapping and navigation.
I do not think that "lidar_collision" or "lidar_enabler" are a great names. Do you have a better idea now that you know more about this option.

miio/roidmivacuum_miot.py
martin9000andersen: Yes, I think that "girl_en" can be part of the string. But I also think that it must be json format, and I do not know the json-key. (I have tried {"voice": "girl_en"} without lock)

martin9000andersen:
> I added some more comments, I think it's good to go after this round but be aware that the API will change for the next major release to allow consolidating the common functionalities among all supported vacuums :-)

Thank you for the comments. I have added comments on a few points, the rest of the points look very clear and I hope to get time to solve them within a week.

martin9000andersen:
> While doing the review I did some googling and started to wonder if this PR is for "Roidmi Eve Plus" or is there also a version without the "plus"? If that's the same one, this might be one option when it's time to upgrade away from my gen1 rockrobo.. What's your opinion about the device?

I have the "PLUS" model. The manual covers both models. The robot is the same, the "PLUS" is about the base. The "PLUS" base is with dust-collection. The manual shows a simple base for the non-PLUS model (but I have not seen the non-PLUS version for sale)
The miio protocol has a command to ask about "station-type", that is an int, for me this return 102, and I guess that the simple base will return a different number.

I am very satisfied with Roidmi Eve Plus. The largest draw back I see is missing Home assistant support, but with this pull request and the comming uniform vaccum interface, I think that will be solved.

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,702 @@
...
+        return self.data["double_clean"]
+
+    @property
+    def edge_sweep(self) -> bool:
+        """Is edge clean enabled."""
+        return self.data["edge_sweep"]
+
+    @property
+    def led_switch(self) -> bool:
+        """The LED on the robot will be always on."""
```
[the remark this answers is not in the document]
martin9000andersen: This is a setting that you can change. I have changed the comment, and I have added a function to set this parameter

martin9000andersen:
I have solved (most of) your comments, but also added a few extra functions.
Let me know if there are more changes you like me to make before merging.

miio/roidmivacuum_miot.py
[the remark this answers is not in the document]
rytilahti: Okay, I'd suggest then using `led(self) -> bool` and `set_led(self, bool)` as in elsewhere (like in `airconditioner_miot.py` and `fan_miot.py` among others) to keep consistent naming!

miio/roidmivacuum_miot.py
rytilahti: Okay, let's keep it this way, maybe someone figures it out at some point :-)

miio/roidmivacuum_miot.py
rytilahti: I don't have a better recommendation for this at the moment, so let's keep it as it is.

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+    def path_mode(self) -> PathMode:
+        """Current path-mode:  normal/y-mopping etc."""
+        try:
+            return PathMode(self.data["path_mode"])
+        except ValueError:
+            _LOGGER.error("Unknown PathMode (%s)", self.data["path_mode"])
+            return PathMode.Unknown
+
+    @property
+    def mop_present(self) -> bool:
```
rytilahti, line 289:
```suggestion
    def is_water_box_attached(self) -> bool:
```
is used by rockrobo vacuums for exactly the same thing, right?

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+        """Return True is mop (water box) is installed."""
+        return self.data["mop_present"]
+
+    @property
+    def work_station_frequency(self) -> int:
+        """work_station_freq (2 means base dust colect every second time)."""
```
rytilahti, line 295: Maybe?
```suggestion
    def dust_collection_frequency(self) -> int:
        """Frequency for emptying the dust bin.
        
            Example: 2 means base the dust bin is emptied every second cleaning.
        """
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+        """Is double clean enabled."""
+        return self.data["double_clean"]
+
+    @property
+    def edge_sweep(self) -> bool:
+        """Is edge clean enabled."""
+        return self.data["edge_sweep"]
+
+    @property
+    def vacuum_led(self) -> bool:
```
rytilahti, line 375:
```suggestion
    def led(self) -> bool:
```
this will make it consistent with other integrations. It is enough to have the `station_` prefix below for the station led to differentiate it from this.

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+        """Return if led/display on vaccum is on."""
```
rytilahti, line 376:
```suggestion
        """Return True if led/display on vaccum is on."""
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+        return timedelta(seconds=self.data["clean_time_sec"])
+
+    @property
+    def clean_area(self) -> int:
+        """Cleaned area in m2."""
+        return self.data["clean_area"]
+
+    @property
+    def progress(self) -> str:
+        """Always return None?"""
```
rytilahti, line 415: Is this `None` or `"None"` (as a string)? I'd suggest either removing this if it is not working as expected, as that would otherwise confuse anyone trying to use it (causing issue reports on github for a broken feature :-).

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+    @command()
+    def start(self) -> None:
+        """Start cleaning."""
+        return self.call_action("start")
+
+    @command(click.argument("roomstr", type=str))
+    def start_room_sweep_unknown(self, roomstr: str) -> None:
+        """Start room cleaning.
+
+        FIXME: the syntax of room_sweep is unknown
```
rytilahti, line 591: Considering `roomstr` is not used at all, does something happen if this command gets executed?

If something happens, I'd move the FIXME to be `_LOGGER.warning()` to let users know about it (maybe someone finds out what the string should be) and otherwise remove the command decorator (so that it's not shown in the mioocli but left here just for a reference).

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+
+    @command(
+        click.argument("sweep_mode", type=EnumType(SweepMode)),
+        click.argument("clean_info", type=str),
+    )
+    def start_sweep_unknown(self, sweep_mode: SweepMode, clean_info: str) -> None:
+        """Start sweep with mode.
+
+        FIXME: the syntax of start sweep with mode is unknown
+        """
```
rytilahti, line 603: Is this fixme still relevant considering the parameters are passed to the action?

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+        """Locate the device (i am here)."""
+        return self.call_action("identify")
+
+    @command(click.argument("on", type=bool))
+    def set_station_led(self, on: bool):
+        """Enable station led display."""
+        return self.set_property("station_led", on)
+
+    @command(click.argument("on", type=bool))
+    def set_vacuum_led(self, on: bool):
```
rytilahti, line 627:
```suggestion
    def set_led(self, on: bool):
```

miio/roidmivacuum_miot.py
```diff
@@ -0,0 +1,769 @@
...
+        return self.set_property("sweep_type", sweep_type.value)
+
+    @command(click.argument("path_mode", type=EnumType(PathMode)))
+    def set_path_mode(self, path_mode: PathMode):
+        """Set path_mode."""
+        return self.set_property("path_mode", path_mode.value)
+
+    @command(click.argument("work_station_freq", type=int))
+    def work_station_frequency(self, work_station_freq: int):
+        """Set work_station_freq (2 means base dust colect every second time)."""
```
rytilahti, line 658: See my comment above about the getter for this.

miio/roidmivacuum_miot.py
martin9000andersen: On Roidme Eve the water-box is integrated in the dust-bin. The mop-module is approximately 1 cm high and can be attached below the combined dust+water-box.
Strictly speaking the mop is only a piece of fabric, and Roidmi Eve does not have a sensor to detect if that is pressent (and I guess that this is the same for Roborock).
If you would like a common name for Roborock and Roidme Eve, then we might be able to argue that the water-box on Roborock is a mop-module?
For now I think that best name for Roidme Eve it "is_mop_module_attached"?

miio/roidmivacuum_miot.py
martin9000andersen: I have now tried this again. The returned value is the same as if the feature is in not implemented. I will remove this function, and just leave a comment in the _MAPPING dict

miio/roidmivacuum_miot.py
martin9000andersen: The problem is that I have not figured out how to start an "action" with an argument.
I will comment out all the functions that are calling an action with an argument.

miio/roidmivacuum_miot.py
martin9000andersen: The problem is that I have not figured out how to start an "action" with an argument.
I will comment out all the functions that are calling an action with an argument.

miio/roidmivacuum_miot.py
rytilahti: Ah, so this is really about the mop, so I'd just use `is_mop_attached`.

rytilahti (approved):
All set, thanks @martin9000andersen!

kali876:
I see the subject is a bit frozen :D 
I do not hide that it needs the functionality of vacuuming a specific zone.
If you know how to get mapping for zone vacuuming - I can help :)
I can follow all instructions as long as I get them :)
thx

PR #1075, opened 2021-06-20
increase socket buffer size 1024->4096

starkillerOG:
When a gateway has 30+ devices connected the message of the device list becomes bigger than 1024 bytes.
That means only part of the message is read, and therefore the checksum will not match and you get an error.
simply increasing the buffer size fixed this issue.
See https://github.com/home-assistant/core/issues/51229

### DIFF
--- a/miio/miioprotocol.py
+++ b/miio/miioprotocol.py
@@ -190,7 +190,7 @@ class MiIOProtocol:
             raise DeviceException from ex
 
         try:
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti could you merge this extremely small PR?

rytilahti:
Good catch! LGTM :+1:

PR #1077, opened 2021-06-20
Calculate `depth` for zhimi.humidifier.ca1

bieniu:
zhimi.humidifier.ca1 same aszhimi.humidifier.cb2 reports `127` whithout water tank and `125` when water tank is full.


rytilahti (approved):
LGTM, thanks! :+1:

Issue #1076, opened 2021-06-20
set_rotate FanP10 sends the wrong command


pooyashahidi:
**Describe the bug**

I wanted to ask if anyone has actually tested the `set_rotate` method in `FanMiot`. I'm using the `FanP10` class with a P18 device as they have the same services definition. As far as I was able to test, everything works as expected except the set_rotate method.

This is the message that is being sent:
```
[{'did': 'set_move', 'siid': 2, 'piid': 9, 'value': ['right']}]
```

This is what I think should've been sent (it works):
```
[{'did': 'set_move', 'siid': 2, 'piid': 9, 'value': 2}]
```
the urn file for both p10 and p18 says value should be uint8 (if I understant that correctly)

```json
               {
                    "iid": 9,
                    "description": "Motor Control",
                    "format": "uint8",
                    "access": [
                        "write"
                    ],
                    "unit": "none",
                    "value-list": [
                        {
                            "value": 0,
                            "description": "NO"
                        },
                        {
                            "value": 1,
                            "description": "LEFT"
                        },
                        {
                            "value": 2,
                            "description": "RIGHT"
                        }
                    ]
                },
```

**Version information (please complete the following information):**
 - OS: Linux,Windows
 - python-miio: 0.5.6

**Device information:**
If the issue is specific to a device:
  - Model: P10 (I'm using it with P18 but hopefully that should not matter)
  - Hardware version: esp8266
  - Firmware version: 2.0.7

**To Reproduce**
1. try to send set_rotate command for an instance of FanP10 by:
```fan.set_rotate(MoveDirection("left"))```
2. the device beeps but nothing happens 

**Expected behavior**
device should turn 5°

**Console output**
If applicable, add console output to help explain your problem.
If the issue is about communication with a specific device, consider including the output using the `--debug` flag.

syssi:
Do you like to provide a PR?

pooyashahidi:
> Do you like to provide a PR?

I summited a possible fix for it.
P9,P10,P11 (P15,P18) all should expect the same values.

rytilahti:
Closed with #1078.

PR #1078, opened 2021-06-20
Fix set_rotate for dmaker.fan.p10 (#1076)

pooyashahidi:
A suggestion to fix rotation command for P9,P10,P11,P15,P18,...
fixes issue #1076

### DIFF
--- a/miio/fan_miot.py
+++ b/miio/fan_miot.py
@@ -381,6 +381,7 @@ class FanMiot(MiotDevice):
         default_output=format_output("Rotating the fan to the {direction}"),
     )
     def set_rotate(self, direction: MoveDirection):
+        """Rotate fan to given direction."""
         # Values for: P9,P10,P11,P15,P18,...
         # { "value": 0, "description": "NONE" },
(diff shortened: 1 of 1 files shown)


miio/fan_miot.py
```diff
@@ -381,7 +381,16 @@ def delay_off(self, minutes: int):
         default_output=format_output("Rotating the fan to the {direction}"),
     )
     def set_rotate(self, direction: MoveDirection):
-        return self.set_property("set_move", [direction.value])
+        # Values for: P9,P10,P11,P15,P18,...
```
rytilahti, line 384:
```suggestion
        """Rotate fan to given direction."""
        # Values for: P9,P10,P11,P15,P18,...
```

rytilahti (commented):
Thanks for the PR! LGTM, but could you please add a docstring to the method while you are at it?

miio/fan_miot.py
pooyashahidi: Thank you for pointing that out.

Issue #788, opened 2020-08-04
Smart Mi Standing fan 3 (Xiaomi Pedestal Fan 3, zhimi.fan.za5)


minseokhappylee:
Before submitting a new request, use the search to see if there is an existing issue for the device.

**Device information:**

  - Name(s) of the device: Smart mi standing fan 3
  - Link: https://en.techreviewer.de/xiaomi-smartmi-standing-fan-3/

Use `miiocli device --ip <ip address> --token <token>`.

  - Model: zhimi.fan.za5
  - Hardware version: esp32
  - Firmware version: 2.0.9

**Additional context**
If you know already about potential commands or any other useful information to add support for the device, please add that information here.

rezmus:
miotspec.

RockyLim92:
Is there any plan for supporting zhimi.fan.za5 or any workaround for supporting it?

Thank you

al-one:
Try https://github.com/al-one/hass-xiaomi-miot

gordon-zhao:
Quick updates:
I reviewed the miot-spec of this fan, but the specification is a huge mess (look at the custom-service). We definitely need a new mapping to custom services for some basic fan services (e.g. left/right turn, speed level, ac state). 

Besides, the left/right turn service (urn:zhimi-spec:property:swing-step-move:00000003:zhimi-za5:1) is different from the "motor-control" service seen in dmaker fans. It receives string instead of int, and there is no document about what strings it will receive. With a bit of guessing, here are some strings it receives: ["left", "right"]. On my device, it will turn left/right accordingly if I send these strings.

ha0y:
This device is tested fully supported by https://github.com/ha0y/xiaomi_miot_raw .

PR #1087, opened 2021-06-26
Add support for Smartmi Standing Fan 3 (zhimi.fan.za5)

rnovatorov:
Resolves https://github.com/rytilahti/python-miio/issues/788.

### DIFF
--- a/miio/fan_miot.py
+++ b/miio/fan_miot.py
@@ -11,6 +11,7 @@ MODEL_FAN_P9 = "dmaker.fan.p9"
 MODEL_FAN_P10 = "dmaker.fan.p10"
 MODEL_FAN_P11 = "dmaker.fan.p11"
 MODEL_FAN_1C = "dmaker.fan.1c"
+MODEL_FAN_ZA5 = "zhimi.fan.za5"
 
 MIOT_MAPPING = {
     MODEL_FAN_1C: {
@@ -67,12 +68,34 @@ MIOT_MAPPING = {
         "power_off_time": {"siid": 3, "piid": 1},
         "set_move": {"siid": 6, "piid": 1},
     },
+    MODEL_FAN_ZA5: {
+        # https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:fan:0000A005:zhimi-za5:1
+        "power": {"siid": 2, "piid": 1},
+        "fan_level": {"siid": 2, "piid": 2},
+        "swing_mode": {"siid": 2, "piid": 3},
+        "swing_mode_angle": {"siid": 2, "piid": 5},
+        "mode": {"siid": 2, "piid": 7},
+        "power_off_time": {"siid": 2, "piid": 10},
+        "anion": {"siid": 2, "piid": 11},
+        "child_lock": {"siid": 3, "piid": 1},
+        "light": {"siid": 4, "piid": 3},
+        "buzzer": {"siid": 5, "piid": 1},
+        "buttons_pressed": {"siid": 6, "piid": 1},
+        "battery_supported": {"siid": 6, "piid": 2},
+        "set_move": {"siid": 6, "piid": 3},
+        "speed_rpm": {"siid": 6, "piid": 4},
+        "powersupply_attached": {"siid": 6, "piid": 5},
+        "fan_speed": {"siid": 6, "piid": 8},
+        "humidity": {"siid": 7, "piid": 1},
+        "temperature": {"siid": 7, "piid": 7},
+    },
 }
 
 SUPPORTED_ANGLES = {
     MODEL_FAN_P9: [30, 60, 90, 120, 150],
     MODEL_FAN_P10: [30, 60, 90, 120, 140],
     MODEL_FAN_P11: [30, 60, 90, 120, 140],
+    MODEL_FAN_ZA5: [30, 60, 90, 120],
 }
 
 
@@ -164,8 +187,7 @@ class FanStatus1C(DeviceStatus):
 
     def __init__(self, data: Dict[str, Any]) -> None:
         self.data = data
-        """
-        Response of a Fan1C (dmaker.fan.1c):
+        """Response of a Fan1C (dmaker.fan.1c):
 
         {
           'id': 1,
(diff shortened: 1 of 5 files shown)


idantene:
Tested locally as well - works great! 👏🏻

miio/fan_miot.py
```diff
@@ -525,3 +548,260 @@ def delay_off(self, minutes: int):
...
+class OperationModeFanZA5(enum.Enum):
+    Nature = 0
+    Normal = 1
+
+
+class FanStatusZA5(DeviceStatus):
+    """Container for status reports for FanZA5."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
```
rytilahti, line 562: Please add an example input for the data.

miio/fan_miot.py
```diff
@@ -525,3 +548,260 @@ def delay_off(self, minutes: int):
...
+
+    # TODO: Docstrings.
```
rytilahti, line 564: Please add the docstrings :-)

miio/fan_miot.py
```diff
@@ -525,3 +548,260 @@ def delay_off(self, minutes: int):
...
+    @property
+    def anion(self) -> bool:
```
rytilahti, line 566: What does this mean?

miio/fan_miot.py
```diff
@@ -525,3 +548,260 @@ def delay_off(self, minutes: int):
...
+    def mode(self) -> OperationMode:
+        return OperationMode[OperationModeFanZA5(self.data["mode"]).name]
+
+    @property
+    def power(self) -> bool:
+        return self.data["power"]
+
+    @property
+    def power_off_time(self) -> int:
+        return self.data["power_off_time"]
```
rytilahti, line 618: Could you do a brief check which properties may be common among the other fans? The plan is to move towards a common, device-type specific APIs, and having matching naming schemes & return types will help when that (breaking) change happens.

rytilahti (commented):
Thanks for the PR @rnovatorov! I added a couple of comments inline, could you also add some tests to make coverage checks happy? :-)

miio/fan_miot.py
rnovatorov: Example added. Is the format OK?

miio/fan_miot.py
rnovatorov: Docstrings added.

miio/fan_miot.py
rnovatorov: It is the state of a built-in [air ionizer](https://en.wikipedia.org/wiki/Air_ioniser):
> Some commercial air purifiers are designed to generate negative ions.

miio/fan_miot.py
rnovatorov: I have checked the properties of the other fans and adapted some of the new methods to be similar to the existing ones. If you think that additional changes are required, then could you, please, be more specific about which properties you would like me to edit and how?

miio/fan_miot.py
rytilahti: Looks good to me, you can use that as an input to the unit tests of the status class, too!

miio/fan_miot.py
rytilahti: Maybe `is_ionizer_enabled` then?

rytilahti (commented):
> I have checked the properties of the other fans and adapted some of the new methods to be similar to the existing ones. If you think that additional changes are required, then could you, please, be more specific about which properties you would like me to edit and how?

I think it's fine, just add tests and we will consolidate it when there will be a common fan base class (#1114 lists some of the upcoming changes to make it easier to interact).

miio/fan_miot.py
rnovatorov: Renamed as suggested.

rnovatorov:
Added the tests. I guess everything is ready now.

miio/fan_miot.py
```diff
@@ -525,3 +547,312 @@ def delay_off(self, minutes: int):
...
+
+    @command(default_output=format_output("Turning ionizer on"))
+    def ionizer_on(self):
+        """Turn ionizer on."""
+        return self.set_property("anion", True)
+
+    @command(default_output=format_output("Turning ionizer off"))
+    def ionizer_off(self):
+        """Turn ionizer off."""
+        return self.set_property("anion", False)
```
rytilahti, line 751: Use `ionizer(self, on: bool)` just like for oscillate & buzzer to avoid separate command for toggling.

miio/fan_miot.py
```diff
@@ -525,3 +547,312 @@ def delay_off(self, minutes: int):
...
+            if oscillate
+            else "Turning off oscillate"
+        ),
+    )
+    def set_oscillate(self, oscillate: bool):
+        """Set oscillate on/off."""
+        if oscillate:
+            return self.set_property("swing_mode", True)
+        else:
+            return self.set_property("swing_mode", False)
```
rytilahti, line 791:
```suggestion
        return self.set_property("swing_mode", oscillate)
```

miio/fan_miot.py
```diff
@@ -525,3 +547,312 @@ def delay_off(self, minutes: int):
...
+        default_output=format_output(
+            lambda buzzer: "Turning on buzzer" if buzzer else "Turning off buzzer"
+        ),
+    )
+    def set_buzzer(self, buzzer: bool):
+        """Set buzzer on/off."""
+        if buzzer:
+            return self.set_property("buzzer", True)
+        else:
+            return self.set_property("buzzer", False)
```
rytilahti, line 804:
```suggestion
        return self.set_property("buzzer", buzzer)
```

miio/fan_miot.py
```diff
@@ -525,3 +547,312 @@ def delay_off(self, minutes: int):
...
+    )
+    def set_child_lock(self, lock: bool):
+        """Set child lock on/off."""
+        return self.set_property("child_lock", lock)
+
+    @command(
+        click.argument("light", type=int),
+        default_output=format_output("Setting light to {light}%"),
+    )
+    def set_light(self, light: int):
```
rytilahti, line 820:
```suggestion
    def set_led_brightness(self, light: int):
```

rytilahti (commented):
Just a couple minor things, and this is ready to be merged. Please also update the README.md to include mention about this device.

miio/fan_miot.py
rnovatorov: Done.

miio/fan_miot.py
rnovatorov: Done. This code was copy-pasted from `FanMiot`, so I have made the fixes there as well.

miio/fan_miot.py
rnovatorov: Done.

miio/fan_miot.py
rnovatorov: Done.

rnovatorov:
Fixed the minor things. Mentioned the device in `README.md`.

miio/fan_miot.py
rytilahti: Ah, thanks! Maybe there should be a generic base class for all those fans, but that can wait for now.

rytilahti (approved):

PR #1089, opened 2021-06-28
Add `water_level` and `water_tank_detached` property for humidifiers, deprecate `depth`

bieniu:
This PR:
- adds `water_level` property for AirHumidifier (value is calculated from `depth`)
- adds `water_tank_detached` for AirHumidifier (value is calculated from `depth`)
- fixes `water_level` calculation for AirHumidifierMiot
- adds `water_tank_detached` for AirHumidifierMiot (value is calculated from `water_level`)

### DIFF
--- a/miio/airhumidifier.py
+++ b/miio/airhumidifier.py
@@ -186,6 +186,9 @@ class AirHumidifierStatus(DeviceStatus):
     @property
     def depth(self) -> Optional[int]:
         """Return raw value of depth."""
+        _LOGGER.warning(
+            "The 'depth' property is deprecated and will be removed in the future. Use 'water_level' and 'water_tank_detached' properties instead."
+        )
         if "depth" in self.data:
             return self.data["depth"]
         return None
@@ -260,10 +263,6 @@ class AirHumidifier(Device):
 
         # TODO: convert to use generic device info in the future
         self.device_info: Optional[DeviceInfo] = None
-
-        _LOGGER.warning(
-            "The 'depth' property is deprecated and will be removed for 3 months (october 2021). Use 'water_level' and 'water_tank_detached' properties instead."
-        )
 
     @command(
         default_output=format_output(
--- a/miio/airhumidifier.py
+++ b/miio/airhumidifier.py
@@ -261,6 +261,8 @@ class AirHumidifier(Device):
         # TODO: convert to use generic device info in the future
         self.device_info: Optional[DeviceInfo] = None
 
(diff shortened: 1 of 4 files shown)


miio/airhumidifier.py
```diff
@@ -185,15 +185,29 @@ def motor_speed(self) -> Optional[int]:
 
     @property
     def depth(self) -> Optional[int]:
-        """The remaining amount of water in percent."""
+        """Return raw value of depth."""
```
rytilahti, line 188: So, this is a breaking change but I don't think it's a big deal in this case.

One question though, should this property be marked as deprecated/should there be a logger warning guiding to use `water_level` instead of this?

bieniu: Good question. HA integration will use only `water_level` value. I think we should marked `depth` value as deprecated and remove it in, I don't know, maybe 3 months. What do you think?

Tomorrow I'll add a logger warning about the deprecation of `depth`.

rytilahti: Sounds good to me! Those (if any) who want to access the raw value can always simply access it through the raw data. Let's merge this after adding a warning that the behavior has been changed and that using the new `water_level` and `water_tank_detached` is recommended.

bieniu: Done

miio/airhumidifier.py
```diff
@@ -247,6 +261,10 @@ def __init__(
         # TODO: convert to use generic device info in the future
         self.device_info: Optional[DeviceInfo] = None
 
+        _LOGGER.warning(
+            "The 'depth' property is deprecated and will be removed for 3 months (october 2021). Use 'water_level' and 'water_tank_detached' properties instead."
```
rytilahti, line 265:
```suggestion
            "The 'depth' property is deprecated and will in the future. Use 'water_level' and 'water_tank_detached' properties instead."
```
Move this to the `depth` property itself so it won't cause a warning if one is not using it. Also, no need to be that specific on when it gets removed, just that it's deprecated. There is also `deprecated` decorator available in utils that can be used (which uses warnings instead of logger for warning).

rytilahti (approved):

bieniu:
@rytilahti Are you planning to release a new version? The migration of humidifiers from the `fan` platform to the `humidifier` in HA was merged and now we need `water_level` property 😁

rytilahti:
Yes, I'll prepare a new release when I return from the holidays (i.e., there should be a new release available in a couple of weeks), if @syssi doesn't beat me with that. Sorry for the delay :/

bieniu:
> Sorry for the delay :/

No problem, enjoy the holidays.

Issue #1086, opened 2021-06-26
Unable to discover vacuum cleaner: Xiaomi Mi Robot Vacuum Mop (aka dreame.vacuum.mc1808)


StarterCraft:
**Bug Information: general description**
A Mi Robot Vacuum Mop (aka dreame.vacuum.mc1808) is connected to the same network as the computer running `miio` commands. Robot's connecting information is retrieved via [this](https://github.com/Maxmudjon/Get_MiHome_devices_token) program, and the token looks legit, according to the [docs](https://python-miio.readthedocs.io/en/latest/discovery.html#device-discovery) of `miio`. Various commands fail to connect to the robot, behaving as stated in the following:
```
G:\Work\Code>miiocli discover
INFO:miio.miioprotocol:Sending discovery to <broadcast> with timeout of 5s..
INFO:miio.miioprotocol:Discovery done
INFO:miio.discovery:Discovering devices with mDNS for 5 seconds...
//As you can see, discovering command does not seem to discover anything

G:\Work\Code>mirobo --ip "192.168.0.110" --token <myToken>
ERROR:miio.miioprotocol:Got error when receiving: timed out
Error: No response from the device
//That's the typical response while the 'mirobo' command is executed, however,
//I have got an unusual response once, stating the following:

G:\Work\Code>mirobo --ip "192.168.0.110" --token <myToken>
Error: Unable to discover the device 192.168.0.110
```
According to existing info in the `miio` documentation, the `unable to discover` error is appearing pretty often, but I didn't manage to find exact instructions for fixing this error.

**Version information:**
 - OS: Windows 10 version 1909
 - python-miio: 0.5.6
 
**Device information:**
Model: dreame.vacuum.mc1808
Hardware version: Linux
Firmware version: 3.5.8_1096

**To Reproduce:**
1. If you have the Mi Robot Vacuum Mop at your's, try to run `mirobo --ip {yourIP} --token {yourToken}`

**Expected behavior:**
I expected a standard `mirobo` information message, and some data about the device discovery while using `miiocli discover` command.

**Console output:**
```G:\Work\Code>miiocli --debug discover
INFO:miio.cli:Debug mode active
INFO:miio.miioprotocol:Sending discovery to <broadcast> with timeout of 5s..
INFO:miio.miioprotocol:Discovery done
INFO:miio.discovery:Discovering devices with mDNS for 5 seconds...

G:\Work\Code>mirobo --debug --ip "192.168.0.110" --token {myToken} info
[...]

Issue #1066, opened 2021-06-05
Vacuum 1C STYTJ01ZHM (dreame.vacuum.mc1808) - Many retries (and a potential fix)...


anonymous-one:
Before I begin, I am not very well versed in Python or the miio protocol... But here goes:

When ever I would try to execute any commands / pull any info from my STYTJ01ZHM miio would take quite some time to execute the request... At times, it would even fail out right.

When using the `-d` flag I noticed that there was quite a few retries:

`DEBUG:miio.miioprotocol:Retrying with incremented id, retries left: 1`

To fix this I modified `miioprotocol.py` with the following change:

`self.__id += 1`

to

`self.__id += random.randint(1,100)`

With this I am able to avoid retries and the replies to the commands happen on first request.

Not sure if this breaks things... But in my case it works.

StarterCraft:
This helped me to solve my connection problem, when the robot failed to answer 90% of the commands I sent him. Thank you!

rytilahti:
I just noticed this issue thanks to @StarterCraft's mention.

So, the id sequence behavior may differ between devices and maybe even whether you also use the official app or some other software to control the device simultaneously. That is also why the ID gets increment by 100 for retries (https://github.com/rytilahti/python-miio/blob/master/miio/miioprotocol.py#L231).

However, as you are not using the library (but the `miiocli` tool), this won't be saved anywhere so what happens is the following:
A: first miiocli execution (id=1)
B: second miiocli execution (id=1), gets ignored by the device as 1<2
B: automatic retry (id=101), this should work now
C: third miiocli execution (id=1), ignored again
C: miiocli retry (id=101), ignored again as the device now awaits >101 for the id
C: miiocli retry (id=201), this should work again
D: fourth miiocli execution (more retries needed now until the device resets its internal id counter)

The proper fix here is to modify miiocli to save the previously used sequence much like mirobo (for rockrobo only) does.

StarterCraft:
Suggested PR #1092 to resolve this issue as @anonymous-one did.

PR #1092, opened 2021-07-04
Update protocol core to fix SKV4093GL connection problem

StarterCraft:
SKV4093GL (aka STYTJ01ZHM aka dreame.vacuum.mc1808) connection problem was referenced first in #1066, later in #1086, and the working (as for me) fix was provided by @anonymous-one. However @rytilahti [suggested](https://github.com/rytilahti/python-miio/issues/1066#issuecomment-872536431) another way to fix this problem, I'd like to request to accept @anonymous-one's solution of the problem as the fastest one.

### DIFF
--- a/miio/miioprotocol.py
+++ b/miio/miioprotocol.py
@@ -258,11 +258,11 @@ class MiIOProtocol:
     @property
     def _id(self) -> int:
         """Increment and return the sequence id."""
-        self.__id += 1
+        self.__id += random.randint(1, 100)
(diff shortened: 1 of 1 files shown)


rytilahti:
As I mentioned earlier, I don't feel good about changing the id generation on that level as that might introduce problems on other devices that are currently working just fine.

One possible solution would be adding an interface to allow device implementations to define their own id generation logic, but it feels kludgy, too. :(

PR #1093, opened 2021-07-10
add Air Purifier Pro H support

dtbell91:


rytilahti:
Would you mind also adding a note to the README.md's supported devices, as well as to `airpurifier_miot.py` with the model information (from the info command, likely `zhimi.airpurifier.vb2`)?

PR #1098, opened 2021-07-14
Fix cct_max for ZNLDP12LM

mouth4war:


starkillerOG:
@rytilahti this PR is correct and can be merged.
I did not know the right min/max cct values of bulbs (I dont have one).

Thanks @mouth4war

syssi (approved):

Issue #1003, opened 2021-04-08
miiocli viomivacuum status not working for STYJ02YM (viomi.vacuum.v8, fw: 3.5.3_0017)


bkupidura:
**Describe the bug**
miiocli viomivacuum status is broken for STYJ02YM with firmware 3.5.3_0017.
Looks like vacuum is returning less properties than expected.

List of available properties can be found https://github.com/nqkdev/home-assistant-vacuum-styj02ym/blob/master/vacuum.py#L122, i see that nqkdev is not using all available - but this can be start point for further investigation.

**Version information (please complete the following information):**
 - OS: Home-Assistant container, tag 2021.4.0
 - python-miio: 0.5.5

**Device information:**
Model: viomi.vacuum.v8
Hardware version: Linux
Firmware version: 3.5.3_0017

**To Reproduce**
1. Run miiocli viomivacuum status

**Expected behavior**
Status command working

**Console output**
```
# miiocli --debug viomivacuum --ip 10.0.150.27 --token <TOKEN> status
INFO:miio.cli:Debug mode active

DEBUG:miio.protocol:Unable to decrypt, returning raw bytes: b''
DEBUG:miio.miioprotocol:Got a response: Container:
    data = Container:
        data = b'' (total 0)
        value = b'' (total 0)
        offset1 = 32
        offset2 = 32
        length = 0
    header = Container:
        data = b'!1\x00 \x00\x00\x00\x00\x13\x9f\x9c\xc8`n\xb1\x15' (total 16)
        value = Container:
            length = 32
            unknown = 0
            device_id = b'\x13\x9f\x9c\xc8' (total 4)
            ts = 2021-04-08 07:30:29
        offset1 = 0
        offset2 = 16
        length = 16
    checksum = b'\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff' (total 16)
DEBUG:miio.miioprotocol:Discovered 139f9cc8 with ts: 2021-04-08 07:30:29, token: b'ffffffffffffffffffffffffffffffff'
DEBUG:miio.miioprotocol:10.0.150.27:54321 >>: {'id': 1, 'method': 'get_prop', 'params': ['battary_life', 'box_type', 'cur_mapid', 'err_state', 'has_map', 'has_newmap', 'hw_info', 'is_charge', 'is_mop', 'is_work', 'light_state', 'm
[...]

PR #1099, opened 2021-07-14
Fix the issue with a wrong properties retrieval for Viomi Vacuum

nergal:
Fixes the issue that is described here https://github.com/rytilahti/python-miio/issues/1003. The cause is in the way how the device responds to a query of multiple properties. It just skips None values which makes other logic fails.


rytilahti:
I'm sorry, but this is not a proper fix for this issue and would cause a flood of requests which is not acceptable. The real solution would be requesting the supported properties based on the detected model.

PR #1124, opened 2021-08-30
Do not use deprecated `depth` property

bieniu:
This PR fixes my stupid mistake started here https://github.com/rytilahti/python-miio/pull/1089

We don't want to use deprecated `depth` property to calulate the `water_level` property.

### DIFF
--- a/miio/airhumidifier.py
+++ b/miio/airhumidifier.py
@@ -199,8 +199,9 @@ class AirHumidifierStatus(DeviceStatus):
 
         If water tank is full, depth is 125.
         """
-        if self.data.get("depth") is not None and self.data["depth"] <= 125:
-            return int(self.data["depth"] / 1.25)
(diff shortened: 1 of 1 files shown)


miio/airhumidifier.py
```diff
@@ -199,8 +199,8 @@ def water_level(self) -> Optional[int]:
 
         If water tank is full, depth is 125.
         """
-        if self.depth is not None and self.depth <= 125:
-            return int(self.depth / 1.25)
+        if self.data.get("depth") is not None and self.data["depth"] <= 125:
```
rytilahti, line 202: Introduce a new local variable `depth` to avoid typing `self.data` again:
```
depth = self.data.get("depth")
...
```

Otherwise looks ok for me :-)

rytilahti (approved):
Thanks! :+1:

PR #1125, opened 2021-08-31
add lumi.plug.mmeu01 - ZNCZ04LM

starkillerOG:
found from https://github.com/rytilahti/python-miio/discussions/1121#discussioncomment-1250771

### DIFF
--- a/miio/gateway/devices/subdevices.yaml
+++ b/miio/gateway/devices/subdevices.yaml
@@ -618,6 +618,22 @@
       unit: Watt
       get: get_property_exp
 
+- zigbee_id: lumi.plug.mmeu01
+  model: ZNCZ04LM
+  type_id: -2
+  name: Plug
+  type: Switch
+  class: Switch
+  getter: get_prop_plug
+  setter: toggle_plug
+  properties:
+    - property: neutral_0 # 'on' / 'off'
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti could you merge this?

miio/gateway/devices/subdevices.yaml
```diff
@@ -618,6 +618,22 @@
       unit: Watt
       get: get_property_exp
 
+- zigbee_id: lumi.plug.mmeu01
+  model: ZNCZ04LM
+  type_id: -2
```
rytilahti, line 623: Is it normal for type id to be negative? Otherwise LGTM :+1:

rytilahti (approved):

miio/gateway/devices/subdevices.yaml
starkillerOG: No but I don't know the type id and could not find it on the internet, so I set it to a negative value.
Someone who owns the device and a chinese version of the Hub schould get a warning about the device not beeing known and a message that he schould report it here.
Than we can change the -2 to the correct value.

rytilahti: Ok :+1:

PR #1131, opened 2021-09-04
Do not get battery status for mains powered devices

starkillerOG:
This schould fix the errors as reported here: https://github.com/rytilahti/python-miio/discussions/1121#discussioncomment-1267774
If the battery status is requested from a mains powered device like the `lumi.plug.mmeu01` it gives an error like:
`miio.exceptions.DeviceError: {'code': -4003, 'message': 'prop or action not exist'}`

### DIFF
--- a/miio/gateway/devices/subdevice.py
+++ b/miio/gateway/devices/subdevice.py
@@ -41,6 +41,7 @@ class SubDevice:
         if model_info is None:
             model_info = {}
         self._model_info = model_info
+        self._battery_powered = model_info.get("battery", True)
         self._battery = None
         self._voltage = None
         self._fw_ver = dev_info.fw_ver
@@ -208,6 +209,9 @@ class SubDevice:
     @command()
     def get_battery(self):
         """Update the battery level, if available."""
+        if not self._battery_powered:
+            return self._battery
+
         if self._gw.model not in [GATEWAY_MODEL_EU, GATEWAY_MODEL_ZIG3]:
             self._battery = self.send("get_battery").pop()
         else:
@@ -220,6 +224,9 @@ class SubDevice:
     @command()
     def get_voltage(self):
         """Update the battery voltage, if available."""
+        if not self._battery_powered:
+            return self._voltage
+
         if self._gw.model in [GATEWAY_MODEL_EU, GATEWAY_MODEL_ZIG3]:
             self._voltage = self.get_property("voltage").pop() / 1000
         else:
--- a/miio/gateway/devices/subdevice.py
+++ b/miio/gateway/devices/subdevice.py
@@ -207,9 +207,13 @@ class SubDevice:
         return self.send("remove_device")
 
     @command()
-    def get_battery(self):
+    def get_battery(self) -> Optional[int]:
         """Update the battery level, if available."""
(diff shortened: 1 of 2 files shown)


starkillerOG:
@rytilahti could you look at this and merge?

miio/gateway/devices/subdevice.py
```diff
@@ -208,6 +209,9 @@ def unpair(self):
     @command()
     def get_battery(self):
         """Update the battery level, if available."""
+        if not self._battery_powered:
```
syssi, line 212: The negation doesn't look correct here.

miio/gateway/devices/subdevice.py
```diff
@@ -220,6 +224,9 @@ def get_battery(self):
     @command()
     def get_voltage(self):
         """Update the battery voltage, if available."""
+        if not self._battery_powered:
```
syssi, line 227: The negation doesn't look correct here.

miio/gateway/devices/subdevice.py
```diff
@@ -220,6 +224,9 @@ def get_battery(self):
...
+            return self._voltage
```
syssi, line 228:
```suggestion
            return None
```

miio/gateway/devices/subdevice.py
```diff
@@ -208,6 +209,9 @@ def unpair(self):
     @command()
     def get_battery(self):
         """Update the battery level, if available."""
+        if not self._battery_powered:
+            return self._battery
```
syssi, line 213:
```suggestion
            return None
```

miio/gateway/devices/subdevice.py
```diff
@@ -41,6 +41,7 @@ def __init__(
         if model_info is None:
             model_info = {}
         self._model_info = model_info
+        self._battery_powered = model_info.get("battery", True)
```
syssi, line 44:
```suggestion
        self._battery_powered = model_info.get("battery_powered", True)
```

What about calling the model info property `battery_powered` too? Just for consistency.

starkillerOG: sure I will change it

starkillerOG: done

starkillerOG:
@syssi any more suggestions?

syssi (approved):

syssi:
I will merge this PR later the day. I want to give @rytilahti the chance to react.

miio/gateway/devices/subdevice.py
```diff
@@ -220,6 +224,9 @@ def get_battery(self):
     @command()
     def get_voltage(self):
         """Update the battery voltage, if available."""
+        if not self._battery_powered:
+            return None
```
rytilahti, line 228: I'm fine with this, but maybe add a debug log just in case someone starts to wonder why they are just getting None as a response. Also, adding type hints (here `Optional[int]`?) would be helpful.

starkillerOG: done, thanks for the tips

starkillerOG:
I think this can now be merged

syssi (approved):

Issue #1135, opened 2021-09-06
Incorrect water level of air humidifiers


OleksandrBerchenko:
The water level of air humidifiers is incorrect (assuming that the value in Mi Home is correct).

You have the following code (https://github.com/rytilahti/python-miio/blob/master/miio/airhumidifier_miot.py#L135):

`return int(self.data["water_level"] / 1.25)`

`1.25` here should be changed to `1.2`.  But you are right, `125` still means the full (actually, overflowing) tank.

**UPDATE:** The model of my air humidifiers is `zhimi.humidifier.ca1`.

rytilahti:
Ping @bieniu

bieniu:
> `1.25` here should be changed to `1.2`

Why? Do we have any specification for this?

OleksandrBerchenko:
We have the Mi Home application.  If you change `1.25` to `1.2`, you will get the identical value.

bieniu:
Not on my humidifier, I have 98-96% when depth is 120.

OleksandrBerchenko:
Oh, really?  What model do you have?  I have two `zhimi.humidifier.ca1` with the latest firmware.  They worked that way from the very beginning.

bieniu:
Really, my zhimi.humidifier.ca1 works that way.

OleksandrBerchenko:
Right now, I have the following values:
Humidifer #1: MI Home: 84%, python-miio: 81%
Humidifer #2: MI Home: 75%, python-miio: 72%

OleksandrBerchenko:
What is the easiest way to print raw `depth` values?  I'm using this library through Home Assistant.

OleksandrBerchenko:
Thanks, that works.

humidifier 1: depth: 68, water level: 54, mi home: 56
humidifier 2: depth: 95, water level: 76, mi home: 79

bieniu:
For your humidifiers `1,2` gives much better results. I'm curious why for mine `1,25` is better. I'll try to make more some research in this case but I don't know when. There was an attempt to break into my apartment, so my world is turned upside down now.

OleksandrBerchenko:
Oh, I'm really sorry to hear that.  That sounds awful!

No rush: I have already workarounded this issue by a correction formula in HA.

OleksandrBerchenko:
Just in case, one more observation:

[...]

PR #1140, opened 2021-09-17
Fix `water_level` calculation for humidifiers

bieniu:
This PR fixes the calculation of the `water_level` according to the source of Mi Home app [here](https://github.com/rytilahti/python-miio/issues/1135#issuecomment-921658587) and [here](https://github.com/rytilahti/python-miio/issues/1135#issuecomment-921660645).

Fix: https://github.com/rytilahti/python-miio/issues/1135


bieniu:
@rytilahti Can you help me? Docstring CI fails on multiple files. Should I improve them all in this PR?

rytilahti:
@bieniu yeah, it's better to fix all affected files in cases like this. I'm not sure why it wasn't fixed automatically (and why the doc8 problem got in in the first place), but I just fixed it by running `pre-commit run -a` & modifying the readme manually.

miio/airhumidifier.py
```diff
@@ -198,11 +198,11 @@ def depth(self) -> Optional[int]:
     def water_level(self) -> Optional[int]:
         """Return current water level in percent.
 
-        If water tank is full, depth is 125.
+        If water tank is full, depth is 120. If water tank is overfilled, depth is 125.
         """
         depth = self.data.get("depth")
         if depth is not None and depth <= 125:
-            return int(depth / 1.25)
+            return int(max(min(depth / 1.2, 100), 0))
```
rytilahti, line 205: Why use `max(val, 0)` here, avoiding negative values?

rytilahti (commented):
Do you mind adding some tests for the behavior, so that it won't break in the future?

miio/airhumidifier.py
bieniu: Because we have this
```
        else if (value === -1) {
          value = 0;
        }
```
in tha app source.

bieniu:
Yes, I'll try to add some tests.

miio/airhumidifier.py
rytilahti: Sorry for the delay, I see. I think it would be the best to inverse the condition and bail out early
```
if depth is None or depth < 0:
    return None

return <real value here>
```

miio/tests/test_airhumidifier.py
```diff
@@ -467,6 +467,25 @@ def dry():
         self.device.set_dry(False)
         assert dry() is False
 
+    def test_water_level(self):
```
rytilahti, line 470: I think it makes sense to use [pytest.parametrize](https://docs.pytest.org/en/6.2.x/example/parametrize.html) here:
```suggestion
    water_level_test_data = [(-1, 0), (0, 0), (60, 50), ...]
    @pytest.parametrize("depth, excepted", water_level_test_data)
    def test_water_level(self, depth, excepted):
        self.device.state["depth"] = depth
        assert self.state().water_level == expected
```

miio/tests/test_airhumidifier.py
```diff
@@ -694,3 +714,22 @@ def dry():
 
         self.device.set_dry(False)
         assert dry() is False
+
+    def test_water_level(self):
```
rytilahti, line 718: See my above comment about `parametrize`.

miio/tests/test_airhumidifier_miot.py
```diff
@@ -193,3 +193,22 @@ def clean_mode():
 
         self.device.set_clean_mode(False)
         assert clean_mode() is False
+
+    def test_water_level(self):
```
rytilahti, line 197: Here also :-)

rytilahti (changes requested):

miio/tests/test_airhumidifier.py
bieniu: Hmmm I think parametrization doesn't work with `unitteset`. Or I don't know how to use it properly.
```
self = <unittest.case._Outcome object at 0x7fe3287c4430>, test_case = <miio.tests.test_airhumidifier_miot.TestAirHumidifier testMethod=test_water_level>, isTest = True

    @contextlib.contextmanager
    def testPartExecutor(self, test_case, isTest=False):
        old_success = self.success
        self.success = True
        try:
>           yield

self = <miio.tests.test_airhumidifier_miot.TestAirHumidifier testMethod=test_water_level>
method = <bound method TestAirHumidifier.test_water_level of <miio.tests.test_airhumidifier_miot.TestAirHumidifier testMethod=test_water_level>>

    def _callTestMethod(self, method):
>       method()
E       TypeError: test_water_level() missing 2 required positional arguments: 'water_level' and 'expected'
```

rytilahti: How does the code look like at the moment? Something like this?
```
@pytest.mark.parametrize("water_level,expected", ...)
def test_water_level(test, water_level, expected)
```

For some reason it doesn't seem to be passing the parameters in on that code path.

rytilahti: Ah, I see, that's not possible when using `TestCase` classes: https://stackoverflow.com/questions/35561843/py-test-parametrize-test-cases-from-classes - maybe it's time to convert it to use regular test methods to avoid the whole class-based structure. Mind if I take a shot at it?

miio/airhumidifier.py
rytilahti: I did simplify the logic a bit. If it looks okay to you, I think this is ready to be merged after the tests are passing.

rytilahti (approved):

bieniu:
@rytilahti Wow, you were busy today :) Great work!

PR #1160, opened 2021-10-16
create separate directory for yeelight

Kirmas:
Based on https://github.com/rytilahti/python-miio/pull/1094#pullrequestreview-769683812 but I believe the "moving" PR could be separate. This PR only move the yeelight code to the folder. Next PR will be soon.

@rytilahti could you merge this one?

### DIFF
--- a/miio/__init__.py
+++ b/miio/__init__.py
@@ -41,6 +41,7 @@ from miio.gateway import Gateway
 from miio.heater import Heater
 from miio.heater_miot import HeaterMiot
 from miio.huizuo import Huizuo, HuizuoLampFan, HuizuoLampHeater, HuizuoLampScene
+from miio.integrations.yeelight import Yeelight
 from miio.miot_device import MiotDevice
 from miio.philips_bulb import PhilipsBulb, PhilipsWhiteBulb
 from miio.philips_eyecare import PhilipsEyecare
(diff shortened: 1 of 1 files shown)


Kirmas:
> Let's move it under `integrations/yeelight/` and use absolute imports, this way:
> 
> 1. the integrations are clearly separated from the main miio code-base, and
> 2. it will be easier to add support for out-of-tree integrations, when all integrations are self-contained

@rytilahti Did I do this correct? Could you provide what else I must change? Maybe move test some how.

rytilahti:
> @rytilahti Did I do this correct? Could you provide what else I must change? Maybe move test some how.

Looking good, but yeah, the test should also be moved under `tests/` inside the integration directory. Could you try if it still get executed fine after moving it? I think pytest should find it still just fine, and If not, we'll need to figure out how to adjust the test collection. The test itself should probably use a relative import for the module under test, and absolute imports just for modules that are contained under `miio` or `miio.tests`.

rytilahti:
@Kirmas the integrations should be inside the miio package so that it will be accessible. Now running `pytest miio` works fine at least for me :-)

rytilahti:
Yes, I think we can merge this now. Thanks for the PR and pushing this onward :-)

PR #1163, opened 2021-10-20
Add light specs for yeelight

Kirmas:
This is base support for spec file & uses it to provide information about supported models. Second PR based on https://github.com/rytilahti/python-miio/pull/1094#pullrequestreview-769683812


Kirmas:
> So, we could only initialize it only once inside the module. Think about a situation with 10 yeelight bulbs, calling the static method would cause file I/O & parsing for each and every instance. This would also skip those if the Yeelight module never gets initialized. Here's what I think could work:

I`m not satisfied but agree I/O for home assistant(or any other CD card user) more important than RAM. Hope in the future SSD will be cost nothing and all off us start to use it :) @rytilahti Please look current state.

PR #1173, opened 2021-11-02
Add more supported vacuum models

OGKevin:
Add supported models that have been reported by users of HA.

This PR is in favour of all the 3 small ones to add the models.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,7 +100,7 @@ To ease the process of setting up a development environment we have prepared `a 
 Supported devices
 -----------------
 
--  Xiaomi Mi Robot Vacuum V1, S4, S4 MAX, S5, S5 MAX, M1S, S7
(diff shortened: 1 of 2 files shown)


rytilahti:
@OGKevin do you happen to have a list of features that are only available on some models?

I have been looking into some potential flags. It's probably a good idea to start with something simple (so not adding all of these at once) and extend from that, but it would be really helpful to have some sort of list of different features available:

* Supports percentage based fan speeds
* Supports timers
* Supports DnD mode
* Supports mopping
* Supports mopping modes ("mop routing on s7", likely similar to viomi patterns)
* Supports carpet cleaning/avoidance mode
* Supports water flow controls
* Supports automatic bin cleaning

rytilahti (approved):
Let's get this merged now for the next release so that we can leverage it inside homeassistant instead of depending on the consts.

Issue #897, opened 2020-12-28
Support for pet water dispenser mmgg.pet_waterer (Xiaowan XWWF01MG)


gudvinr:
**Device information:**

  - Name(s) of the device: Xiaowan Smart Pet Water Dispenser
  - Link: [AliExpress](https://www.aliexpress.com/i/4001143786885.html)
  - Model: `mmgg.pet_waterer.s1` and `mmgg.pet_waterer.s4`
  - Hardware version: `esp8266` (s1)
  - Firmware version: `2.0.4_0006`

**Additional context**
s1 and s4 version almost identical according to miot spec except s4 has one additional property. s4 probably is an EU version of this water dispenser and s1 is Chinese one.

miot-spec:
* [mmgg-s1.json](https://github.com/rytilahti/python-miio/files/5748859/urn.miot-spec-v2.device.pet-drinking-fountain.0000A067.mmgg-s1.1.json.txt)
* [mmgg-s4.json](https://github.com/rytilahti/python-miio/files/5748858/urn.miot-spec-v2.device.pet-drinking-fountain.0000A067.mmgg-s4.1.json.txt)

HydrelioxGitHub:
I'm interested as I acquired this device a couple of months ago. If I could help in any way ..

rytilahti:
If you want to contribute, please take a look at other miot platforms (and `devtools/` directory for miottemplate to make spec files easier to comprehend).

HydrelioxGitHub:
I followed the guide provided by @gudvinr and I have been able to send commands to the device through command line.

I have take a quick look at other devices. I think I could do something but I'm not very skilled so it will take some time. Maybe waterpurifier.py is the most useful example I will take for example (work in a close way as water dispenser, simple stucture).

HydrelioxGitHub:
I do not get where python-miio do the match between property name written in the code and siid, piid and so on ...

HydrelioxGitHub:
I saw by reading some comment that @gudvinr is much more skilled than me. Is he still working on this issue ?

rytilahti:
@HydrelioxGitHub take a look at https://github.com/rytilahti/python-miio/blob/master/miio/heater_miot.py (or any other `_miot.py` file) for some ideas, I think the heater one is quite straightforward here so that's why I'm linking it.

[...]

PR #1174, opened 2021-11-02
add support for smart pet water dispenser mmgg.pet_waterer.s1

ofen:
https://github.com/rytilahti/python-miio/issues/897

Tested on following hardware
Model: mmgg.pet_waterer.s1
Hardware version: esp8266
Firmware version: 2.1.3_0013


miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,274 @@
...
+    @property
+    def filter_left_time(self) -> int:
+        '''Filter life time remaining in days'''
+        return self.data['filter_left_time']
+
+    @property
+    def power(self) -> str:
+        if self.data['on']:
+            return 'on'
+        return 'off'
```
rytilahti, line 66:
```suggestion
```

This is not needed as everyone should just use `is_on`.

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,274 @@
...
+                for prop in self.get_properties_for_mapping()
+            }
+        )
+
+    @command(default_output=format_output('Powering off ...'))
+    def off(self) -> Any:
+        '''Power off.'''
+        status = self.status()
+        if not status.is_on:
+            return 'Already off'
```
rytilahti, line 163:
```suggestion
```
It's simpler just to turn it off without checking the state beforehand.

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,274 @@
...
+        if resp[0]['code'] != 0:
+            raise PetWaterDispenserException(resp)
+        return 'ok'
+
+    @command(default_output=format_output('Powering on ...'))
+    def on(self) -> str:
+        '''Power on.'''
+        status = self.status()
+        if status.is_on:
+            return 'Already on'
```
rytilahti, line 174:
```suggestion
```
Same as above.

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,274 @@
...
+        resp = self.set_property('on', True)
+        if resp[0]['code'] != 0:
+            raise PetWaterDispenserException(resp)
+        return 'ok'
+
+    @command(default_output=format_output('Disabling indicator light ...'))
+    def disable_indicator_light(self) -> str:
```
rytilahti, line 181: Is indicate light just something like a led on some other devices? If yes, I think it's a good idea to use the same convention here.

Also, instead of creating a separate method for disabling and enabling the light, please use a boolean parameter like done for example here: https://github.com/rytilahti/python-miio/blob/master/miio/airfresh.py#L274

miio/miot_device.py
```diff
@@ -145,3 +145,10 @@ def set_property(self, property_key: str, value):
             "set_properties",
             [{"did": property_key, **self.mapping[property_key], "value": value}],
         )
+
+    def get_property(self, property_key: str, value):
+        """Gets property value using the existing mapping."""
+        return self.send(
+            "get_properties",
+            [{"did": property_key, **self.mapping[property_key], "value": value}],
+        )
```
rytilahti, line 154:
```suggestion
```
This is unnecessary and not used anywhere anyway.

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,274 @@
...
+    @command(default_output=format_output('Powering on ...'))
+    def on(self) -> str:
+        '''Power on.'''
+        status = self.status()
+        if status.is_on:
+            return 'Already on'
+        resp = self.set_property('on', True)
+        if resp[0]['code'] != 0:
+            raise PetWaterDispenserException(resp)
+        return 'ok'
```
rytilahti, line 178:
```suggestion
        return self.set_property('on', True)
```
Just return the return value of the `set_property` on all cases, it's up to the caller to handle the error cases (for now, at least).

rytilahti (changes requested):
Thanks for the PR, @ofen! I added a couple of comments inline. Btw, while developing, you can also call `pre-commit run -a` (or `tox -e lint`) to run all the linting tests locally.

ofen:
@rytilahti, thank you for feed back. Should we use bool switch for power on/off property too (same as https://github.com/rytilahti/python-miio/blob/master/miio/airfresh.py#L274)?

ofen:
Is there any guide how we can fix not-supported device warning in case of [MiotDevice class](https://github.com/rytilahti/python-miio/blob/master/miio/miot_device.py#L30)?

Warning due to empty `_supported_models` list https://github.com/rytilahti/python-miio/blob/master/miio/device.py#L152-L157

ofen:
Changes applied according to previous comments

rytilahti:
> Should we use boolean switch for power on/off property too

Turning on and off should be implemented by `on()` and `off()` by convention.

> Is there any guide how we can fix not-supported device warning in case of [MiotDevice class](https://github.com/rytilahti/python-miio/blob/master/miio/miot_device.py#L30)?

This is fixed by defining a class variable like here: https://github.com/rytilahti/python-miio/blob/master/miio/integrations/vacuum/roborock/vacuum.py#L146

ofen:
@rytilahti done

1. Reverted `on()` and `off()` methods back
2. `_supported_models` added to fix warning
3. More comments added
4. Status properties renamed according to return values

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+
+class PetWaterDispenserException(DeviceException):
+    pass
+
+
+class PetWaterDispenserStatus(DeviceStatus):
+    """Container for status reports from the Pet Water Dispenser."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
```
rytilahti, line 54: Could you add an example contents here, something like https://github.com/rytilahti/python-miio/blob/master/miio/airconditioner_miot.py#L157 that someone without a device can get an idea how the data looks like.

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+}
+
+
+class OperatingMode(enum.Enum):
+    common = 1
+    smart = 2
+
+
+class PetWaterDispenserException(DeviceException):
+    pass
```
rytilahti, line 47:
```suggestion
```
This is not used anymore.

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+    "remain_clean_time": {"siid": 6, "piid": 1},
+    "reset_filter_life": {"siid": 3, "aiid": 1},
+    "reset_device": {"siid": 8, "aiid": 1},
+    "timezone": {"siid": 9, "piid": 1},
+}
+
+
+class OperatingMode(enum.Enum):
+    common = 1
+    smart = 2
```
rytilahti, line 43:
```suggestion
    Normal = 1
    Smart = 2
```

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+from .click_common import EnumType, command, format_output
+from .exceptions import DeviceException
+from .miot_device import DeviceStatus, MiotDevice
+
+_LOGGER = logging.getLogger(__name__)
+
+MODEL_MMGG_PET_WATERER_S1 = "mmgg.pet_waterer.s1"
+MODEL_MMGG_PET_WATERER_S4 = "mmgg.pet_waterer.s4"
+
+SUPPORTED_MODELS: List[str] = [MODEL_MMGG_PET_WATERER_S1]
```
rytilahti, line 16:
```suggestion
SUPPORTED_MODELS: List[str] = [MODEL_MMGG_PET_WATERER_S1, MODEL_MMGG_PET_WATERER_S4]
```

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class PetWaterDispenserStatus(DeviceStatus):
+    """Container for status reports from the Pet Water Dispenser."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
+
+    @property
+    def sponge_filter_left_days(self) -> int:
```
rytilahti, line 57: All properties returning times remaining (or similar) should use `timedelta` to be consistent across the library.

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+
+    @property
+    def is_on(self) -> bool:
+        """True if device is on."""
+        return self.data["on"]
+
+    @property
+    def mode(self) -> str:
+        """OperatingMode."""
+        return OperatingMode(self.data["mode"]).name
```
rytilahti, line 69:
```suggestion
        return OperatingMode(self.data["mode"])
```
Better just return the mode here

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+
+    @property
+    def is_no_water(self) -> bool:
+        """True if there is no water left."""
+        if self.data["no_water_flag"]:
+            return False
+        return True
+
+    @property
+    def no_water_minutes(self) -> int:
```
rytilahti, line 94: Timedelta?

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
...
+        return self.set_property("on", False)
+
+    @command(
+        click.argument("led", type=bool),
+        default_output=format_output(
+            lambda led: "Turning LED on" if led else "Turning LED off"
+        ),
+    )
+    def set_led(self, led: bool) -> List[Dict[str, Any]]:
+        """Toggle idicator light on/off."""
```
rytilahti, line 175:
```suggestion
        """Toggle indicator light on/off."""
```

miio/pet_water_dispenser.py
```diff
@@ -0,0 +1,227 @@
+import enum
```
rytilahti, line 1: Could you move this file to `miio/integrations/petwaterdispenser/` directory?

rytilahti (commented):
Looking good, here another round of comments!

ofen:
@rytilahti all above done

rytilahti (commented):
Looking great!

One last wish, could you at least add some tests to the `PetWaterDispenserStatus`? Nothing fancy is needed here, simply verifying that that the example payload is parsed correctly:
```
data = (the example response)

def test_status():
    status = PetWaterDispenserStatus(data)

    assert status.is_on is  True
    assert status.sponge_filter_left_days == timedelta(days=10)
    # and so on
```

Also, instead of using `__init__.py` as the main code file, it'd be better to store the code elsewhere and just import the exported symbols (e.g., `PetWaterDispenser`) in side the `__init__.py`. I notice that this is not properly done in other integrations either, so sorry for that confusion.

ofen:
@rytilahti moved code from `__init__.py` to `device.py` and `status.py` files, status test also added.

rytilahti (approved):
LGTM, thanks! :tada:

PR #1180, opened 2021-11-06
Fix test_properties command logic

Zuz666:
This change is necessary because the current logic of test_properties command in some cases leads to an error on converting the exception object, as well as an inaccurate output if only one parameter is possible during the miIO request.


rytilahti (approved):
LGTM, thanks!

PR #1185, opened 2021-11-15
Add Air Purifier Pro H support

pvizeli:
Success of https://github.com/rytilahti/python-miio/pull/1093

The Pro H works like the 3H just a bit bigger.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -103,7 +103,7 @@ Supported devices
 -  Xiaomi Mi Robot Vacuum V1, S5, S5 MAX, M1S, S7
 -  Xiaomi Mi Home Air Conditioner Companion
 -  Xiaomi Mi Smart Air Conditioner A (xiaomi.aircondition.mc1, mc2, mc4, mc5)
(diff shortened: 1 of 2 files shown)


pvizeli:
According: https://github.com/rytilahti/python-miio/issues/1186

I did not notice that. How is the best way to address this?

rytilahti:
Considering it's working already, I would simply add it to the supported devices in the integration itself. The implementing class (`AirPurifierMiot`) does not currently have `_supported_devices` variable, but one should be added as the plan is to start using that information to allow initializing device classes based on the model information.

The mdns discovery information is not currently used at all, but it could be a part of extended metadata that will in the future get exposed over the `supported_devices()`. The aim is to move out device specific handling inside the lib so that downstreams do not have to keep their own mappings for things.

pvizeli:
What I do not understand is why it work as mb3 device but the linked URL https://github.com/rytilahti/python-miio/issues/1186 is complete different mapping

Issue #1107, opened 2021-08-05
Add support for Roborock Auto Empty Dock


andrey-yantsen:
**Is your feature request related to a problem? Please describe.**
Recently, Roborock released their Auto Empty Dock for the S7 robot vacuum. It would be great if it were possible to control the dock using the library (at least — to be able to trigger the emptying cycle manually).

**Describe the solution you'd like**
Ideally, I'd like to have a few extra methods in the vacuum-cli — `empty`, `set_auto_empty_settings` and `get_auto_empty_settings`. You can guess what they should do from their names :)

But to be honest, it would be just enough for me if somebody could advise how I can get the command the app uses to trigger emptying.

**Device information:**
If the enhancement is device-specific, please include also the following information.

  - Name(s) of the device: Auto Empty Dock
  - Link: https://global.roborock.com/pages/roborock-auto-empty-dock

Use `miiocli device --ip <ip address> --token <token>`.

  - Model: roborock.vacuum.a15
  - Hardware version: Linux
  - Firmware version: 4.1.5_1196

**Additional context**

The dock itself doesn't connect to WiFi, seems like all the logic is done via the vacuum itself. I'll be happy to help with any required debugging.

P.S. I tried running tcpdump on my router, but unfortunately, I don't see any traffic on the UDP port 54321 between the vacuum and my iOS device.

andrey-yantsen:
Found everything I needed in the Mi Home's plugin code.

First of all, to check whether the "auto empty dock" is available — `FeatureManager.robotNewFeatures & 0x2000000`. Then, to start the process — `app_start_collect_dust`, and to abort it `app_stop_collect_dust`.

To enable the automatic emptying call `set_dust_collection_switch_status` with `{"status": 1}`, and `{"status": 0}` to disable.

To change the emptying mode call `set_dust_collection_mode`, and `get_dust_collection_mode` to get the current one.

There're 5 modes defined in the code, but only 4 available in the app:
```
  var DustCollectionModeSettingMap = {
    DustCollectionModeSmart: 0,
    DustCollectionModeQuick: 1,
    DustCollectionModeDaily: 2,
    DustCollectionModeStrong: 3, // <---- this one is not available in the app
    DustCollectionModeMax: 4
  };
```

craigcabrey:
@andrey-yantsen thanks for finding all that, it was quite easy to add support

shred86:
> Found everything I needed in the Mi Home's plugin code.
> 
[...]

PR #1188, opened 2021-11-28
roborock: auto empty dustbin support

craigcabrey:
addresses #1107

This is just an initial pass at this, probably more needs to be done. Will also look at the home assistant component.

Example CLI run:

```
(miio-runtime) [craigcabrey@alto python-miio]$ python3 miio/cli.py vacuum --ip 192.168.10.222 --token XXXX dust_collection_mode
Running command dust_collection_mode
DustCollectionMode.Smart
```

### DIFF
--- a/miio/integrations/vacuum/roborock/vacuum.py
+++ b/miio/integrations/vacuum/roborock/vacuum.py
@@ -773,7 +773,7 @@ class Vacuum(Device):
 
     @command(click.argument("enabled", required=True, type=bool))
     def set_dust_collection(self, enabled: bool) -> bool:
-        """Turn automatic dust collection on or off"""
+        """Turn automatic dust collection on or off."""
         self._verify_auto_empty_support()
         return (
             self.send("set_dust_collection_switch_status", {"status": int(enabled)})[0]
@@ -784,10 +784,7 @@ class Vacuum(Device):
     def set_dust_collection_mode(self, mode: DustCollectionMode) -> bool:
         """Set dust collection mode setting."""
         self._verify_auto_empty_support()
-        return (
-            self.send("set_dust_collection_mode", {"mode": mode.value})[0]
-            == "ok"
-        )
+        return self.send("set_dust_collection_mode", {"mode": mode.value})[0] == "ok"
 
     @command()
     def start_dust_collection(self):
@@ -803,9 +800,7 @@ class Vacuum(Device):
 
     def _verify_auto_empty_support(self) -> None:
         if self.model not in self._auto_empty_models:
-            raise VacuumException(
-                "Device does not support auto emptying"
-            )
+            raise VacuumException("Device does not support auto emptying")
 
     @command()
     def stop_zoned_clean(self):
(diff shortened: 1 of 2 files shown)


miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -746,6 +756,49 @@ def set_carpet_cleaning_mode(self, mode: CarpetCleaningMode):
...
+            == "ok"
+        )
+
+    @command()
+    def disable_dust_collection(self):
+        """Disable automatic dust collection."""
+        return (
+            self.send("set_dust_collection_switch_status", {"status": 0})[0]
+            == "ok"
+        )
```
rytilahti, line 773: Please combine these to a single `set_dust_collection(self, enabled: bool)` method.

rytilahti (commented):
Thanks for the PR @craigcabrey!

As we want to have some sort of feature flags at some point to allow downstreams to decide which features are supported, could you please add a guard if clause to check for the models supporting this feature prior to calling any of the methods? You could raise a `VacuumException("Device does not support dustbin emptying")` or something similar when trying to run these commands on a non-supported devices.

That would be only the s7 and s7+ for the time being, I suppose?

craigcabrey:
> As we want to have some sort of feature flags at some point to allow downstreams to decide which features are supported, could you please add a guard if clause to check for the models supporting this feature prior to calling any of the methods? You could raise a `VacuumException("Device does not support dustbin emptying")` or something similar when trying to run these commands on a non-supported devices.

Yea, that works. It's a little ugly but over time with a feature flag framework we could clean it up.

> That would be only the s7 and s7+ for the time being, I suppose?

There is no S7+, just the S7. S7+ is a bundle of the S7 with the auto empty bin.

rytilahti:
> Yea, that works. It's a little ugly but over time with a feature flag framework we could clean it up.

Indeed, I have some code around locally, but I haven't yet figured out what's the best approach to make it generic among all supported devices. But this will do for now, it's fairly simple to replace that with something more generic when the time comes.

> There is no S7+, just the S7. S7+ is a bundle of the S7 with the auto empty bin.

Ah, thanks for the clarification. This looks good to go as soon as the checks pass :+1:

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -746,6 +761,52 @@ def set_carpet_cleaning_mode(self, mode: CarpetCleaningMode):
             == "ok"
         )
 
+    @command()
+    def dust_collection_mode(self) -> Optional[DustCollectionMode]:
+        """Get the dust collection mode setting."""
+        self._verify_auto_empty_support()
+        try:
+            return DustCollectionMode(self.send("get_dust_collection_mode")["mode"])
```
rytilahti, line 769: Just a side note, is there some other information delivered by that call? If yes, it's probably better to separate the request from accessing the property just in case.

rytilahti (approved):
This is ready to be merged as soon as the checks pass :+1:

Issue #1114, opened 2021-08-13
[meta] Plans for 0.6 series


rytilahti:
This meta task will track the progress and tasks to do for the next major release, which aims to simplify the use of the library for downstream users and reorganize the organically grown code base to be more maintainable. This will cause breaking changes all across the code base, so in order to allow releases to 0.5 series if needed, a new development branch (`next`) is created for this and alpha/beta releases are planned prior the final 0.6.0 release to make testing easier.

This issue will give an overview of the planned changes, which are discussed more in-depth under their corresponding issues linked below as they get written. This meta issue shall serve as a checklist for the changelog, and is open for discussions.

**If you are using python-miio in your project, your feedback on changes you would wish to see in this library are welcome. If you feel like it, feel free to create a PR against the `next` branch, too!**

### Repository structure
* **Motivation:** all source files are currently stored in the main miio directory for historical reasons, making it messy.
* **Goal:** by splitting each individual implementation (integration?) and files related to it (e.g., tests and some data files) will clean up the code base.

#### Tasks
* Separate integrations and their tests to their own directories under `miio/integrations/<name>`

### "Device manager"
* **Motivation:** there is currently no way to construct device instances just by knowing their host and the token, making it necessary for downstreams to import the classes implementing the support and initializing it separately based on model information (or simply supporting a single implementation, like done by homeassistant's vacuum platform).
* **Goal:** allow initializing device-derived objects by simply knowing their host and the token which simplifies downstream implementations.

#### Tasks
* Add manifest files to integrations that describe supported models (miIO.info-based and mDNS entries)
* Add facilities to initialize devices by using only the host and the token
* (potential) plugin support to allow off-tree integrations (e.g., to allow generic miot spec based integrations for those who don't mind the online/spec file requirement)

### Common APIs
* **Motivation:** adding support for devices of the same device class (e.g., vacuums) is cumbersome as there is no well-defined API among the different implementations.
* **Goal:** provide device-type based base classes to make adding support for new device integrations on downstream just a matter of upgrading the python-miio. 

#### Tasks

* Create base classes for individual device types where it makes sense to make it easier for downstreams to add support for a class of devices.
* Add model detection based on `miIO.info` to consolidate handling of device-specific differences inside a single integration.
* Add `supported_features` or similar feature flags to allow integrations specify which features is supported by the device model. This will allow moving these checks from downstream (homeassistant most prominently) to where this information belongs.

### Other changes
[...]

PR #1210, opened 2021-12-06
Add inital support for mmgg.feeder.petfeeder

Wh1terat:
I've tried to follow the style as best I can see.
Was 50/50 on calling the integration "PetfoodDispenser" (keeps with PetWaterDispenser) or simply calling it "PetFeeder".

This device does not follow it's own (somewhat limited) miot spec (annoying since the Manufacturer does make other compliant devices!) so this is all MiiO.

There's a number of features unavailable vs the mi home app:
- Fault status (seemingly only available via push to mihome)
- Feed total (seemingly stored in the plugin within mihome)

Added _most_ functions:

```
Usage: miiocli petfooddispenser [OPTIONS] COMMAND [ARGS]...

Options:
  --ip TEXT     [required]
  --token TEXT  [required]
  --model TEXT

Commands:
  dispense_food     Dispense food.
  info              Get (and cache) miIO protocol information from the...
  raw_command       Send a raw command to the device.
  reset_clean_time  Reset clean time.
  reset_dryer_time  Reset dryer time.
  set_feed_state    Enable / Disable the automatic feeding schedule.
  set_key_lock      Enable / Disable the key lock for the manual dispense...
  set_wifi_led      Enable / Disable the wifi status led.
  status            Retrieve properties.
  test_properties   Helper to test device properties.

Usage: miiocli petfooddispenser dispense_food [OPTIONS] AMOUNT

  Dispense food. :param amount: in units (1 unit ~= 5g)

Usage: miiocli petfooddispenser reset_clean_time [OPTIONS]

  Reset clean time.

Usage: miiocli petfooddispenser reset_dryer_time [OPTIONS]

  Reset dryer time.

Usage: miiocli petfooddispenser set_feed_state [OPTIONS] STATE

  Enable / Disable the automatic feeding schedule.

Usage: miiocli petfooddispenser set_key_lock [OPTIONS] STATE

  Enable / Disable the key lock for the manual dispense button.

Usage: miiocli petfooddispenser set_wifi_led [OPTIONS] STATE

  Enable / Disable the wifi status led.

```

Will look into adding support for feedplans soon.

Feedback welcome.

### DIFF
--- a/miio/integrations/petfooddispenser/device.py
+++ b/miio/integrations/petfooddispenser/device.py
@@ -46,7 +46,7 @@ class PetFoodDispenser(Device):
             "Food bin lid: {result.door_status}\n"
             "Dispense button lock: {result.key_lock}\n"
             "Days until clean: {result.clean_days}\n"
-            "Dessicant life: {result.dryer_days}\n"
+            "Desiccant life: {result.dryer_days}\n"
             "WiFi LED: {result.wifi_led}\n",
         )
     )
--- a/miio/integrations/petfooddispenser/device.py
+++ b/miio/integrations/petfooddispenser/device.py
@@ -7,7 +7,6 @@ import click
 from miio.click_common import EnumType, command, format_output
 from miio.device import Device
 
-from miio.miioprotocol import MiIOProtocol
 
 from .status import PetFoodDispenserStatus
 
--- a/miio/integrations/petfooddispenser/device.py
+++ b/miio/integrations/petfooddispenser/device.py
@@ -5,7 +5,7 @@ from collections import defaultdict
 import click
 
 from miio.click_common import EnumType, command, format_output
-from miio.device import Device
+from miio import Device
 
 
 from .status import PetFoodDispenserStatus
--- a/miio/integrations/petfooddispenser/device.py
+++ b/miio/integrations/petfooddispenser/device.py
@@ -1,0 +1,112 @@
+import logging
+from typing import Any, Dict, List
+
+from collections import defaultdict
+import click
+
+from miio.click_common import EnumType, command, format_output
+from miio.device import Device
+
+from miio.miioprotocol import MiIOProtocol
+
+from .status import PetFoodDispenserStatus
+
+_LOGGER = logging.getLogger(__name__)
+
+MODEL_MMGG_FEEDER_PETFEEDER = "mmgg.feeder.petfeeder"
+
+SUPPORTED_MODELS: List[str] = [MODEL_MMGG_FEEDER_PETFEEDER]
+
+AVAILABLE_PROPERTIES: Dict[str, List[str]] = {
+    MODEL_MMGG_FEEDER_PETFEEDER: [
+        "food_status",
+        "feed_plan",
+        "door_status",
+        "feed_today",
+        "clean_days",
+        "power_status",
+        "dryer_days",
+        "food_portion",
+        "wifi_led",
+        "key_lock",
+        "country_code",
+    ],
+}
+
+class PetFoodDispenser(Device):
+    """Main class representing the Pet Feeder / Smart Pet Food Dispenser. """
+
+    _supported_models = SUPPORTED_MODELS
+
+    @command(
+        default_output=format_output(
+            "",
+            "Power source: {result.power_status}\n"
+            "Food level: {result.food_status}\n"
+            "Automatic feeding: {result.feed_plan}\n"
+            "Food bin lid: {result.door_status}\n"
+            "Dispense button lock: {result.key_lock}\n"
+            "Days until clean: {result.clean_days}\n"
+            "Dessicant life: {result.dryer_days}\n"
+            "WiFi LED: {result.wifi_led}\n",
+        )
+    )
+    def status(self) -> PetFoodDispenserStatus:
+        """Retrieve properties."""
+        properties = AVAILABLE_PROPERTIES.get(
+            self.model, AVAILABLE_PROPERTIES[MODEL_MMGG_FEEDER_PETFEEDER]
+        )
+        values = self.send('getprops')
+        return PetFoodDispenserStatus(defaultdict(lambda: None, zip(properties, values)))
+
+    @command(
+        click.argument("amount", type=int),
+        default_output=format_output("Dispensing {amount} units)"),
+    )
+    def dispense_food(self, amount: int):
+        """Dispense food.
+        :param amount: in units (1 unit ~= 5g)
+        """
+        return self.send("outfood", [amount])
+
+    @command(default_output=format_output("Resetting clean time"))
+    def reset_clean_time(self) -> bool:
+        """Reset clean time."""
+        return self.send("resetclean")
+
+    @command(default_output=format_output("Resetting dryer time"))
+    def reset_dryer_time(self) -> bool:
+        """Reset dryer time."""
+        return self.send("resetdryer")
+
+    @command(
+        click.argument("state", type=int),
+        default_output=format_output(
+            lambda state: "Turning on WiFi LED" if state else "Turning off WiFi LED"
+        ),
+    )
+    def set_wifi_led(self, state: int):
+        """Enable / Disable the wifi status led."""
+        return self.send("wifiledon", [state])
+
+    @command(
+        click.argument("state", type=int),
+        default_output=format_output(
+            lambda state: "Enabling key lock for dispense button" if state else "Disabling key lock for dispense button"
+        ),
+    )
+    def set_key_lock(self, state: int):
+        """Enable / Disable the key lock for the manual dispense button."""
+        return self.send("keylock", [state ^ 1])
+
+    @command(
+        click.argument("state", type=int),
+        default_output=format_output(
+            lambda state: "Enabling automatic feeding schedule" if state else "Disabling automatic feeding schedule"
+        ),
+    )
+    def set_feed_state(self, state: int):
+        """Enable / Disable the automatic feeding schedule."""
+        return self.send("stopfeed", [state])
+
+
--- a/miio/integrations/petfooddispenser/status.py
+++ b/miio/integrations/petfooddispenser/status.py
@@ -1,7 +1,7 @@
 import enum
 from typing import Any, Dict
 
-from miio.miot_device import DeviceStatus
+from miio import DeviceStatus
 
 class FoodStatus(enum.Enum):
     Normal = 0
--- a/miio/integrations/petfooddispenser/status.py
+++ b/miio/integrations/petfooddispenser/status.py
@@ -27,7 +27,7 @@ class PetFoodDispenserStatus(DeviceStatus):
         self.data = data
 
     @property
-    def food_status(self) -> str:
+    def food_status(self) -> FoodStatus:
         """Current food status / level."""
         return FoodStatus(self.data["food_status"])
 
--- a/miio/integrations/petfooddispenser/status.py
+++ b/miio/integrations/petfooddispenser/status.py
@@ -1,0 +1,76 @@
+import enum
+from typing import Any, Dict
+
+from miio.miot_device import DeviceStatus
+
+class FoodStatus(enum.Enum):
+    Normal = 0
+    Low = 1
+    Empty = 2
+
+class PowerState(enum.Enum):
+    Mains = 0
+    Battery = 1
+
+class PetFoodDispenserStatus(DeviceStatus):
+    """Container for status reports from the Pet Feeder / Smart Pet Food Dispenser."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
+
+    @property
+    def food_status(self) -> str:
+        """Current food status / level."""
+        return FoodStatus(self.data["food_status"])
+
+    @property
+    def feed_plan(self) -> bool:
+        """Automatic feeding status."""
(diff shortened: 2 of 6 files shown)


miio/integrations/petfooddispenser/status.py
```diff
@@ -0,0 +1,76 @@
...
+
+class PowerState(enum.Enum):
+    Mains = 0
+    Battery = 1
+
+class PetFoodDispenserStatus(DeviceStatus):
+    """Container for status reports from the Pet Feeder / Smart Pet Food Dispenser."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        self.data = data
```
rytilahti, line 19: Please add an example of the data format here.

miio/integrations/petfooddispenser/status.py
```diff
@@ -0,0 +1,76 @@
+import enum
+from typing import Any, Dict
+
+from miio.miot_device import DeviceStatus
```
rytilahti, line 4:
```suggestion
from miio import DeviceStatus
```

miio/integrations/petfooddispenser/status.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @property
+    def door_status(self) -> bool:
+        """Food bin door status."""
+        return bool(self.data["door_status"])
+
+    @property
+    def feed_today(self) -> int:
+        """Units of food dispensed today."""
+        raise NotImplementedError("Device does not currently return this value correctly.")
```
rytilahti, line 39: I think it'd be better to either return the incorrect value (and log an error) or remove not implemented properties altogether.

miio/integrations/petfooddispenser/device.py
```diff
@@ -0,0 +1,112 @@
+import logging
+from typing import Any, Dict, List
+
+from collections import defaultdict
+import click
+
+from miio.click_common import EnumType, command, format_output
+from miio.device import Device
+
+from miio.miioprotocol import MiIOProtocol
```
rytilahti, line 10:
```suggestion
```

miio/integrations/petfooddispenser/device.py
```diff
@@ -0,0 +1,112 @@
+import logging
+from typing import Any, Dict, List
+
+from collections import defaultdict
+import click
+
+from miio.click_common import EnumType, command, format_output
+from miio.device import Device
```
rytilahti, line 8:
```suggestion
from miio import Device
```

rytilahti (changes requested):
A very brief review, looking great! Please do not forget to add the device to README.md, too

miio/integrations/petfooddispenser/status.py
Wh1terat: It's a tough one. Returning bad data seems worse than not returning the property at all.
I didn't want the implementation to look incomplete but nor did I want to return bad data.

It is your project though so given the 2 choices I'll go with remove :+1:

rytilahti: As the data is still available in the response payloads, anyone can go and check it out in case they start returning more proper values, right? Not having properties raising unexpected exceptions makes it possible to use the whole API without having to dig into the source, so that's why I prefer it that way.

Wh1terat:
Think that's everything fixed :smile_cat:

miio/integrations/petfooddispenser/device.py
```diff
@@ -0,0 +1,111 @@
...
+        default_output=format_output(
+            "",
+            "Power source: {result.power_status}\n"
+            "Food level: {result.food_status}\n"
+            "Automatic feeding: {result.feed_plan}\n"
+            "Food bin lid: {result.door_status}\n"
+            "Dispense button lock: {result.key_lock}\n"
+            "Days until clean: {result.clean_days}\n"
+            "Desiccant life: {result.dryer_days}\n"
+            "WiFi LED: {result.wifi_led}\n",
```
rytilahti, line 50: It would make sense to avoid using the property names from the protocol itself, but use more developer friendly names (similar to use for these printouts).

miio/integrations/petfooddispenser/device.py
```diff
@@ -0,0 +1,111 @@
...
+        """Reset dryer time."""
+        return self.send("resetdryer")
+
+    @command(
+        click.argument("state", type=int),
+        default_output=format_output(
+            lambda state: "Turning on WiFi LED" if state else "Turning off WiFi LED"
+        ),
+    )
+    def set_wifi_led(self, state: int):
```
rytilahti, line 87:
```suggestion
    def set_wifi_led(self, state: bool):
```

miio/integrations/petfooddispenser/device.py
```diff
@@ -0,0 +1,111 @@
...
+        """Reset clean time."""
+        return self.send("resetclean")
+
+    @command(default_output=format_output("Resetting dryer time"))
+    def reset_dryer_time(self) -> bool:
+        """Reset dryer time."""
+        return self.send("resetdryer")
+
+    @command(
+        click.argument("state", type=int),
```
rytilahti, line 82:
```suggestion
        click.argument("state", type=bool),
```
Use bool for boolean values across the board. This should work directly for the cli input values so no need for explicit conversions either.

miio/integrations/petfooddispenser/status.py
```diff
@@ -0,0 +1,67 @@
...
+ 
+        {'food_status': 0, 'feed_plan': 1, 'door_status': 0,
+          'feed_today': 0, 'clean_time': 7, 'power_status': 0,
+          'dryer_days': 6, 'food_portion': 0, 'wifi_led': 1,
+          'key_lock': 1, 'country_code': 255,}
+        """
+        self.data = data
+
+    @property
+    def food_status(self) -> str:
```
rytilahti, line 30:
```suggestion
    def food_status(self) -> FoodStatus:
```

miio/integrations/petfooddispenser/device.py
Wh1terat: Ah yeah I was going to ask about that.

Some names I've taken from miot spec for another feeder by mmgg (one that's actually compliant... :angry: ), but obviously some things don't translate overly well and wasn't sure on the preferred approach.

So you're suggesting (for example) "door_status" > "food_bin_lid" or something along those lines?

miio/integrations/petfooddispenser/device.py
```diff
@@ -0,0 +1,111 @@
...
+        default_output=format_output("Dispensing {amount} units)"),
+    )
+    def dispense_food(self, amount: int):
+        """Dispense food.
+        :param amount: in units (1 unit ~= 5g)
+        """
+        return self.send("outfood", [amount])
+
+    @command(default_output=format_output("Resetting clean time"))
+    def reset_clean_time(self) -> bool:
```
rytilahti, line 72: Are you sure this returns a bool? And same for `reset_dryer_time`?

miio/integrations/petfooddispenser/device.py
rytilahti: Yes, exactly. It makes sense to separate the property names from what is being internally used, as most of the times they are not clear in their meaning.

miio/integrations/petfooddispenser/device.py
Wh1terat: I used bool initially but seemingly then gets sent as literal "true" or "false" (not 1 or 0) which the device does not understand. miot spec for other mmgg products have these defined as ints rather than bools.

miio/integrations/petfooddispenser/device.py
Wh1terat: I used bool initially but seemingly then gets sent as literal "true" or "false" (not 1 or 0) which the device does not understand. miot spec for other mmgg products have these defined as ints rather than bools.

miio/integrations/petfooddispenser/device.py
Wh1terat: Good catch, it returns a list with a string of "ok".
That said, it returns "ok" literally no matter what you send to this device :facepalm:

miio/integrations/petfooddispenser/status.py
```diff
@@ -0,0 +1,67 @@
...
+        """Automatic feeding status."""
+        return bool(self.data["feed_plan"])
+
+    @property
+    def door_status(self) -> bool:
+        """Food bin door status."""
+        return bool(self.data["door_status"])
+
+    @property
+    def clean_days(self) -> int:
```
rytilahti, line 45: Use `timedelta` for this and the dryer one, this makes it consistent with other integrations.
Also, use naming scheme `<variable>_left`.

miio/integrations/petfooddispenser/device.py
rytilahti: That's common on commands, coming likely from the SDK. You could either check for the "ok" in the payload or simply return it as it is, there is no real consistency in the library at the moment on that.

miio/integrations/petfooddispenser/status.py
Wh1terat: Perfect, was just about to ask about this.
"dryer_days" > "desiccant_left" 

Can't quite think how to phrase "clean_days". 
It's the number of days until the unit requires cleaning.
Suggestions?

miio/integrations/petfooddispenser/device.py
rytilahti: Oh, then casting is needed; `int(boolvalue)` will fix that.

miio/integrations/petfooddispenser/device.py
```diff
@@ -0,0 +1,111 @@
...
+        """Enable / Disable the key lock for the manual dispense button."""
+        return self.send("keylock", [state ^ 1])
+
+    @command(
+        click.argument("state", type=int),
+        default_output=format_output(
+            lambda state: "Enabling automatic feeding schedule" if state else "Disabling automatic feeding schedule"
+        ),
+    )
+    def set_feed_state(self, state: int):
```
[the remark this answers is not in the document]
Wh1terat: Naming stuff is the hardest part of all this, honestly :laughing: 

In mihome the toggle is "automatic feeding schedule".

miio/integrations/petfooddispenser/status.py
rytilahti: I don't have a good term for that either, `time_until_clean_left` is sounds quite verbose.

miio/integrations/petfooddispenser/device.py
Wh1terat: :facepalm: 
Given I wrote it at 3am I'm amazed there's not more glaring issues. I'll sort that.

miio/integrations/petfooddispenser/status.py
Wh1terat: Yeah that's what I was trying to avoid, even though it's very pythonic to make it as long as is needed to convey it's use.

miio/integrations/petfooddispenser/device.py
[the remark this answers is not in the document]
rytilahti: Btw, you can run the linting tests locally by executing `tox -e lint` or `pre-commit run -a` (or alternatively, you can do `pre-commit install` and it will perform the checks prior commiting).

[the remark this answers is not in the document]
Wh1terat: I must admit I didn't with this commit, I will in future :angel:

Wh1terat:
Infact yeah, although this was an "initial support" pull - give me a day or so and I'll just refactor and add everything and make it fully supported from day one.

Wh1terat:
Work got a bit busy (chaos before Christmas change freeze!) but have started.

Few questions before we have another shot ;) 

**1. Implementing field maximums (i.e max dispense unit is 30)**
Should this just be a logged error and return false from the relevant function or should I raise an exception? And if exception is it ok to use standard exceptions (i.e ValueError ) or should I subclass DeviceException (as I have seen in other modules).

**2. Timezones!**
I've implemented the feedplan schedules now. The device stores these as CST timezone, which isn't a problem for displaying as I've just made them into datetime objects with the appropriate tzinfo. (Bit annoying that datetime.time doesn't support astimezone, so having to make them full datetime.datetime objects with 1970,1,1 date 😠 )
But the big question, for adding a feedplan - what format would you suggest for the user to input this towards my function? 

datetime.time with tzinfo?
datetime.datetime with tzinfo?
Hour, min as int and iana tz string?

**3. Boolean responses**
What's the general approach for boolean responses, just to return truthy falsy or to map these to string values?
Or create enums just for the 2 choices?
e.g:
On/Off
Locked/Unlocked
Open/Closed
Enabled/Disabled

rytilahti:
> Should this just be a logged error and return false from the relevant function or should I raise an exception? And if exception is it ok to use standard exceptions (i.e ValueError ) or should I subclass DeviceException (as I have seen in other modules).

Good catch, please use standard exceptions. I modified #1114 to list it as something that should be consolidated across the code base.

> What's the general approach for boolean responses, just to return truthy falsy or to map these to string values?

Use booleans where possible, no need to enumize.

On the topic of timezones, I'm not sure what's the best approach here. IIRC the vacuum schedules do convert the times to the local tz, maybe take a look how it's done and follow the same approach?

Wh1terat:
Closing in favour of https://github.com/rytilahti/python-miio/pull/1253

PR #1229, opened 2021-12-12
gateway: remove click support for gateway devices

starkillerOG:
see https://github.com/home-assistant/core/issues/61234#issuecomment-991856052
the new python-miio release broke the alarm and light entity of the gateway

(by rytilahti: the click support was never working for these "subdevices" as they require to be initialized with the gateway instance. this fixes the issue by avoiding calling `info()`/accessing `model` – which is done automatically since #1038 – on a device instance that has no host nor token.)

### DIFF
--- a/miio/gateway/gatewaydevice.py
+++ b/miio/gateway/gatewaydevice.py
@@ -32,4 +32,5 @@ class GatewayDevice(Device):
             )
 
         self._gateway = parent
-        super().__init__(ip, token, start_id, debug, lazy_discover)
+        self.ip = self._gateway.ip
+        self.token = self._gateway.token
--- a/miio/gateway/gatewaydevice.py
+++ b/miio/gateway/gatewaydevice.py
@@ -3,7 +3,6 @@
 import logging
 from typing import TYPE_CHECKING
 
-from ..device import Device
 from ..exceptions import DeviceException
 
 _LOGGER = logging.getLogger(__name__)
@@ -13,17 +12,12 @@ if TYPE_CHECKING:
     from .gateway import Gateway
 
 
-class GatewayDevice(Device):
+class GatewayDevice:
     """GatewayDevice class Specifies the init method for all gateway device
     functionalities."""
 
     def __init__(
         self,
-        ip: str = None,
-        token: str = None,
-        start_id: int = 0,
-        debug: int = 0,
-        lazy_discover: bool = True,
         parent: "Gateway" = None,
     ) -> None:
         if parent is None:
@@ -32,5 +26,3 @@ class GatewayDevice(Device):
             )
 
         self._gateway = parent
-        self.ip = self._gateway.ip
-        self.token = self._gateway.token
--- a/miio/gateway/alarm.py
+++ b/miio/gateway/alarm.py
@@ -2,77 +2,62 @@
 
 from datetime import datetime
 
-import click
-
-from ..click_common import command, format_output
 from .gatewaydevice import GatewayDevice
 
 
 class Alarm(GatewayDevice):
     """Class representing the Xiaomi Gateway Alarm."""
 
-    @command(default_output=format_output("[alarm_status]"))
     def status(self) -> str:
         """Return the alarm status from the device."""
         # Response: 'on', 'off', 'oning'
         return self._gateway.send("get_arming").pop()
 
-    @command(default_output=format_output("Turning alarm on"))
     def on(self):
         """Turn alarm on."""
         return self._gateway.send("set_arming", ["on"])
 
-    @command(default_output=format_output("Turning alarm off"))
     def off(self):
         """Turn alarm off."""
         return self._gateway.send("set_arming", ["off"])
 
-    @command()
     def arming_time(self) -> int:
         """Return time in seconds the alarm stays 'oning' before transitioning to
         'on'."""
         # Response: 5, 15, 30, 60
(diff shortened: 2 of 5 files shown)


starkillerOG:
@rytilahti could you have a look at this since the current HomeAssistant release the alarm and gateway light entities are broken

scxs535:
@rytilahti same problem here.
@starkillerOG thanks

rytilahti:
The real fix is to avoid doing model detection on those subdevices, who do not have a host. I created a PR that should fix this, but someone with a test device should give it a try before I merge it :-)

scxs535:
@rytilahti 
how can we test this exactly?

rytilahti:
@scxs535 you have to clone this PR and use that checkout in your setup. If you are not sure how to do that in your system, it's easier to wait for now.

@starkillerOG could you please create a very simple test script to see what happens if you call `info()` or try to access `model` on alarm or light? That should crash, but if you take the couple of assignments from my earlier PR, it should start working again.

miio/gateway/gatewaydevice.py
```diff
@@ -32,4 +32,5 @@ def __init__(
             )
 
         self._gateway = parent
-        super().__init__(ip, token, start_id, debug, lazy_discover)
```
rytilahti, line 35: See my last edit in my comment. However, I think it may make sense to cleanup this class to not to inherit from `Device` at all (as they are not independent devices) but rather create necessary wrappers to perform I/O through the `parent` reference?

starkillerOG: yea I agree, will convert it now

starkillerOG:
Tested again and it works.
@rytilahti Could you revieuw again/merge?

starkillerOG:
> @scxs535 you have to clone this PR and use that checkout in your setup. If you are not sure how to do that in your system, it's easier to wait for now.
> 
> @starkillerOG could you please create a very simple test script to see what happens if you call `info()` or try to access `model` on alarm or light? That should crash, but if you take the couple of assignments from my earlier PR, it should start working again.
> 
> edit: oh, there's another option that's possible: pass the host, token et. al. to the gateway device constructors, so that they can acquire the info as usual. The downside is that those queries will not be routed through the main instance and will cause I/O, if I'm not mistaken?

I don't really want to do any calls from the subdevice itself, but only invoke calls to the parent object (gateway). Also because the message id will not be properly incremented which could potentially cause issues. (otherwise there would be 2 messages ids that are incrementing at diffrent times and that could get messy).

rytilahti (approved):
LGTM, thanks for stepping up so quickly to fix this @starkillerOG!  :+1:

manu3b1:
I've installed the release 0.5.9.2 and this time the lumi.gateway.v3-MW300 Gateway does not contain any entity anymore.

manu3b1:
With 0.5.9.2 release, with a lumi.gateway.v3-MW300, Alarm and Luminance entities are present, but still no Light.

rytilahti:
@manu3b1 please open a new issue on that and provide some logs so that we have something that can be used to figure out why it isn't working. @starkillerOG do you have a v3? Are you also experiencing the missing light?

manu3b1:
Light is back with HA 2021.12.2.
Thx!

scxs535:
everything(alarm & light) works well as before after update to 2021.12.2 .
thanks @starkillerOG @rytilahti

rytilahti:
Glad to hear it's working and sorry for the inconvenience! It's rather hard to perform refactoring already on a code base size of this even when one tries to be careful.. Thanks for reporting back @scxs535 :-)

maury77:
for me no , I have removed the integration and now

Nuovo tentativo di configurazione: DeviceException during setup of xiaomi gateway with host {self._host}

starkillerOG:
I tested on my setup with gateway.v3 and both the alarm, illumination and light entities all work on the latest 2012.12.2 release.

@maury77 do you have anything else in the log?
Did you set it up using the cloud credentials and are you sure you have the correct credentials and selected the correct server?

maury77:
The cloud connectio si ok

image


configuration 

image

but

image

dic 16 09:42:19 sdomotica homeassistant[32525]: 2021-12-16 09:42:19 WARNING (MainThread) [homeassistant.config_entries] Config entry 'hub mare' for xiaomi_miio integration not ready yet: DeviceException during setup of xiaomi gateway with host {self._host}; Retrying in background

Seems a timeout but the device is reacheble

image

I upgrade the version of the device to  3.1.3_0004

starkillerOG:
@maury77 did you try power cycling the hub (plugin out, wait 15 seconds, plug back in)?
Did this hub work before for you?

maury77:
In this moment I can't turn it off because the gw is in other city, before 
it worked in 2012.11.X

but i also updated the firmware

therefore the combination of new firmware and new Ha has never been tested

starkillerOG:
@maury77 if possible could you try the old HA version 2021.11.X with the new firmware to make sure it is not a firmware problem?
(maybe in a dev enviroment if you have one).

starkillerOG:
I dont have a aqhm01 gateway (I have v3) so I cant test myself, the .v3 gateway works with HA 2021.12.2

maury77:
> I dont have a aqhm01 gateway (I have v3) so I cant test myself, the .v3 gateway works with HA 2021.12.2

With last version and reboot now is ok

PR #1232, opened 2021-12-12
philips_bulb: add philips.light.candle to supported devices

impankratov:
Add support for `philips.light.candle` aka "Xiaomi Philips Rui Chi Candle Light Bulb".

Device stopped working after 2021.12 Home Assistant update, causing this warning message:
```
Found an unsupported model 'philips.light.candle' for class 'Device'. If this is working for you, please open an issue at ...
```

It was working without any problems previously (brightness and cct control at least), not sure about `snm` (scenes) support (never used them), although this device plugin in Mihome app shows that scenes are supported.


rytilahti:
The warning itself should not be an issue, but I'm wondering why your log entry is marking `Device` as the implementation class that is missing the model, any ideas on that?

impankratov:
@rytilahti nice catch!

You're right, I've posted wrong log message in PR description, initially it should say "PhilipsBulb".

rytilahti:
I noticed that there are also other missing philips devices (among many other things) and created #1237. Would it be fine for you to close this PR in favor of that?

impankratov:
@rytilahti sure, no problem! 
Closing in favor of #1237

PR #1233, opened 2021-12-12
gateway: use main device model for subdevices

rytilahti:
This fixes the issue where invoking a method on gateway subdevice causes model discovery,
which fails as the subdevices do not have a host by themselves.

Related to downstream report https://github.com/home-assistant/core/issues/61234 . Replaces #1229

### DIFF
--- a/miio/gateway/gatewaydevice.py
+++ b/miio/gateway/gatewaydevice.py
@@ -33,5 +33,7 @@ class GatewayDevice(Device):
                 "This should never be initialized without gateway object."
             )
 
+        self._info = parent.info()
+        self._model = parent.model
         self._gateway = parent
         super().__init__(ip, token, start_id, debug, lazy_discover)


rytilahti:
@starkillerOG mind testing if this fix works for you?

starkillerOG:
@rytilahti I tested it, but it does not work, still getting:
`miio.exceptions.DeviceException: Unable to discover the device None`

starkillerOG:
https://github.com/rytilahti/python-miio/pull/1229 did work, but I agree that it is quick and dirty.

starkillerOG:
However due to these lines: 
https://github.com/rytilahti/python-miio/blob/4342e3df10af78a508796ac44ab9f04986e535dd/miio/gateway/gatewaydevice.py#L31-L34

Click will not work anyway I think.
You first need to initialize the gateway and then the subdevice can be used.
From the command line (click) that is inpossible anyway (as far as I am aware).

However note that I don't understand click very well and do not use it myself, so take my comments with a grain of salt regarding click.

rytilahti:
Okay, you are right, the click integration won't work so there's not much reason to hold those in the code base.. So let's go with your solution for the time being! It will still cause the same issue if someone for some reason tries to access the `info()`, though...

PR #1235, opened 2021-12-12
Fix Roborock S7 fan speed

shred86:
Small fix to return fan speeds for the Roborock S7. Also updated a broken link in the read me. Tested with latest Home Assistant 2021.12.0 and confirmed it resolved the issue of fan speeds not being displayed correctly.


rytilahti (approved):
Good catch, thanks! :+1:

Issue #1215, opened 2021-12-06
Please add mopping control


arthur-morgan-1:
Hey,
Please add an option to control mopping functions on Roborock S7
it has 5 levels on mopping + Deep/standard choosing 
It would be very useful to control that. especially to choose between mopping or vacuum only.

rytilahti:
You can use the `set_mop_mode` to control the mopping mode, but it supports currently only two modes (standard and deep). There are also controls for `waterflow` ([which has four options](https://github.com/rytilahti/python-miio/blob/0015a5fe2589a6c9236ee2a0af6ab05e7af54aa0/miio/integrations/vacuum/roborock/vacuum.py#L101)), is this what you mean by levels on mopping?

arthur-morgan-1:
oh. I wonder why none of these are implanted on the Home Assistant integration?

rytilahti:
No one has implemented them, I suppose. :-) To be fair, there was also a major rework a release or two ago to make it easier to add add separate sensors, inputs, buttons etc., and it wasn't that simple to add support for such features. So I hope that someone goes and implements more "exotic" features in the future.

shred86:
In addition to the “mop mode”, there’s also a “scrub intensity” setting with 4 options: close, mild, moderate, intense.

shred86:
Just made a PR #1236 for this with more information. Turns out the existing waterflow methods control the scrub intensity.

rytilahti:
As the PR from @shred86 has been merged, this can be closed now.

PR #1236, opened 2021-12-12
Add S7 mop scrub intensity

shred86:
The Roborock S7 has the option to control the scrub intensity with four different settings:

- Close (raised position)
- Mild
- Moderate
- Intense

The way to control these settings is actually identical to the current waterflow code using `self.send("get_water_box_custom_mode")` and `self.send("set_water_box_custom_mode")`. My initial thought was to just add a new `MopIntensity` class since the names of the settings are different and add a case to the existing `waterflow` and `set_waterflow` methods, so if it's an S7 it uses the new `MopIntensity`. However, the method names don't really make sense (which I'm assuming we don't want to change to prevent breaking apps relying on this library) so my thought is the better option is just to duplicate the methods with a new name.

Addresses #1215

### DIFF
--- a/miio/integrations/vacuum/roborock/vacuum.py
+++ b/miio/integrations/vacuum/roborock/vacuum.py
@@ -847,11 +847,17 @@ class RoborockVacuum(Device):
     @command()
     def mop_intensity(self) -> MopIntensity:
         """Get mop scrub intensity setting."""
+        if self.model != ROCKROBO_S7:
+            raise VacuumException("Mop scrub intensity not supported by %s", self.model)
+
         return MopIntensity(self.send("get_water_box_custom_mode")[0])
 
     @command(click.argument("mop_intensity", type=EnumType(MopIntensity)))
     def set_mop_intensity(self, mop_intensity: MopIntensity):
         """Set mop scrub intensity setting."""
+        if self.model != ROCKROBO_S7:
+            raise VacuumException("Mop scrub intensity not supported by %s", self.model)
+
         return self.send("set_water_box_custom_mode", [mop_intensity.value])
 
     @command()
--- a/miio/integrations/vacuum/roborock/vacuum.py
+++ b/miio/integrations/vacuum/roborock/vacuum.py
@@ -114,6 +114,15 @@ class MopMode(enum.Enum):
     Deep = 301
 
 
+class MopIntensity(enum.Enum):
+    """Mop scrub intensity on S7."""
+
+    Close = 200
+    Mild = 201
+    Moderate = 202
+    Intense = 203
+
+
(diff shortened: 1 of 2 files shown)


miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -835,6 +844,16 @@ def set_mop_mode(self, mop_mode: MopMode):
         """Set mop mode setting."""
         return self.send("set_mop_mode", [mop_mode.value])[0] == "ok"
 
+    @command()
+    def mop_intensity(self) -> MopIntensity:
+        """Get mop scrub intensity setting."""
+        return MopIntensity(self.send("get_water_box_custom_mode")[0])
```
rytilahti, line 850: Please add a check that the model (`self.model`) is the expected and raise an exception otherwise.

rytilahti:
> My initial thought was to just add a new MopIntensity class since the names of the settings are different and add a case to the existing waterflow and set_waterflow methods, so if it's an S7 it uses the new MopIntensity. However, the method names don't really make sense (which I'm assuming we don't want to change to prevent breaking apps relying on this library) so my thought is the better option is just to duplicate the methods with a new name.

I think it makes sense to have a separate method as it makes the API cleaner. That being said, there's nothing that prevents us changing the method names if necessary, we can always deprecate the old name and make it a wrapper to call the new functionality prior to removing it.

shred86:
Ah, good point. Well, I do think having it separate would probably be better for future considerations anyways. I wouldn't be surprised to see robot vacuums start being released with both a water and scrub intensity setting.

miio/integrations/vacuum/roborock/vacuum.py
shred86: Just added a simple check to make sure it's not anything besides the S7 since that's the only Roborock version with scrub intensity settings.

rytilahti:
Would you mind adding a couple of tests to make sure this also stays working in case this gets refactored in the future? This is otherwise good to go :+1:

shred86:
Will do! Any ideas how I can change `self._model` within a test method? It's been a while since I've messed with tests and after about an hour of searching, I'm a bit stumped. The `DummyVacuum` test class is initialized with `self._model = "missing.model.vacuum" but I basically need to change that to `"roborock.vacuum.a15"` to be able to test everything else besides raising an exception.

rytilahti:
For the time being, I'd just change the `_model` inside your test for the dummy. The test code is messy and should be refactored, but it's not a small effort so..

shred86:
Sorry for the delay - hopefully this will work.

rytilahti (approved):
LGTM, thanks @shred86 and happy holidays!

PR #1247, opened 2021-12-13
yeelight: use and expose the color temp range from specs

Kirmas:
This PR is next stage to integrate spec file. For more information look to https://github.com/rytilahti/python-miio/pull/1094 

@rytilahti I decided that it would be easier to go function by function and at the end try to add support for background lighting. It's slower, but honestly I don't see any other way to move at least somehow. 

Also fix the incorrect parameters while testing on my device.

### DIFF
--- a/miio/integrations/yeelight/__init__.py
+++ b/miio/integrations/yeelight/__init__.py
@@ -363,7 +363,9 @@ class Yeelight(Device):
     )
     def set_color_temp(self, level, transition=500):
         """Set color temp in kelvin."""
-        color_temp = self._model_info.lamps[YeelightSubLightType.Main].color_temp #for now only main light support
+        color_temp = self._model_info.lamps[
+            YeelightSubLightType.Main
+        ].color_temp  # for now only main light support
         if level > color_temp.max or level < color_temp.min:
             raise YeelightException("Invalid color temperature: %s" % level)
         if transition > 0:
--- a/miio/integrations/yeelight/__init__.py
+++ b/miio/integrations/yeelight/__init__.py
@@ -265,7 +265,6 @@ class Yeelight(Device):
         debug: int = 0,
         lazy_discover: bool = True,
         model: str = None,
-        light_type: YeelightSubLightType = YeelightSubLightType.Main,  # for now only Main support this is for future PR.
     ) -> None:
         super().__init__(ip, token, start_id, debug, lazy_discover, model=model)
         if Yeelight._spec_helper is None:
@@ -273,7 +272,7 @@ class Yeelight(Device):
             Yeelight._supported_models = Yeelight._spec_helper.supported_models
 
         self._model_info = Yeelight._spec_helper.get_model_info(self.model)
-        self._light_type = light_type
+        self._light_type = YeelightSubLightType.Main
         self._light_info = self._model_info.lamps[self._light_type]
         self._color_temp_range = self._light_info.color_temp
 
--- a/miio/integrations/yeelight/__init__.py
+++ b/miio/integrations/yeelight/__init__.py
@@ -8,7 +8,7 @@ from miio.device import Device, DeviceStatus
 from miio.exceptions import DeviceException
 from miio.utils import int_to_rgb, rgb_to_int
 
-from .spec_helper import YeelightSpecHelper, YeelightSubLightType
+from .spec_helper import ColorTempRange, YeelightSpecHelper, YeelightSubLightType
 
 
 class YeelightException(DeviceException):
@@ -265,6 +265,7 @@ class Yeelight(Device):
         debug: int = 0,
         lazy_discover: bool = True,
         model: str = None,
+        light_type: YeelightSubLightType = YeelightSubLightType.Main,  # for now only Main support this is for future PR.
     ) -> None:
         super().__init__(ip, token, start_id, debug, lazy_discover, model=model)
         if Yeelight._spec_helper is None:
@@ -272,6 +273,9 @@ class Yeelight(Device):
             Yeelight._supported_models = Yeelight._spec_helper.supported_models
 
         self._model_info = Yeelight._spec_helper.get_model_info(self.model)
+        self._light_type = light_type
+        self._light_info = self._model_info.lamps[self._light_type]
+        self._color_temp_range = self._light_info.color_temp
 
     @command(default_output=format_output("", "{result.cli_format}"))
     def status(self) -> YeelightStatus:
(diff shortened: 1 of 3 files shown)


Kirmas:
https://github.com/rytilahti/python-miio/pull/1247#issuecomment-992874615 Can anyone tell me why the coverage is less than before? I really do not understand.

miio/integrations/yeelight/specs.yaml
```diff
@@ -80,7 +80,7 @@ yeelink.light.ceiling20:
     supports_color: True
 yeelink.light.ceiling22:
   night_light: True
-  color_temp: [2700, 6500]
+  color_temp: [2600, 6100]
```
rytilahti, line 83: Are these changes on purpose in this PR?

miio/integrations/yeelight/tests/test_yeelight.py
```diff
@@ -25,6 +26,11 @@ def __init__(self, *args, **kwargs):
         }
 
         super().__init__(*args, **kwargs)
+        if Yeelight._spec_helper is None:
+            Yeelight._spec_helper = YeelightSpecHelper()
+            Yeelight._supported_models = Yeelight._spec_helper.supported_models
+
+        self._model_info = Yeelight._spec_helper.get_model_info(self.model)
```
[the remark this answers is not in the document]
rytilahti: Looks ok to me.

miio/integrations/yeelight/specs.yaml
Kirmas: Yes, during I tested on my devices i realized that standard(what I get from yeelight integration) params is wrong. And correct it. If it must be separate PR sorry.

rytilahti: It's fine, you could add a short note in the PR description about that though.

rytilahti:
Perhaps this PR should also include the getter for the valid range? I'm unsure about naming, but maybe something like this?
```
@property
def valid_temperature_range() -> ColorTempRange:
    passs
```

miio/integrations/yeelight/specs.yaml
Kirmas: Name | main.color_temp.min | main.color_temp.max | background.color_temp.min | background.color_temp.max
-- | -- | -- | -- | --
yeelink.light.color3 | 1700 | 6500 | - | -
yeelink.light.color4 | 1700 | 6500 | - | -
yeelink.light.color5 | 1700 | 6500 | - | -
yeelink.light.strip2 | 1700 | 6500 | - | -
yeelink.light.ceiling1 | 2700 | 6500 | - | -
yeelink.light.ceiling20 | 2700 | 6500 | 1700 | 6500
yeelink.light.ceiling4 | 2700 | 6500 | 1700 | 6500
yeelink.light.ceiling22 | 2600 | 6100 | - | -

Result of check.

miio/integrations/yeelight/__init__.py
```diff
@@ -363,7 +363,10 @@ def set_brightness(self, level, transition=0):
     )
     def set_color_temp(self, level, transition=500):
         """Set color temp in kelvin."""
-        if level > 6500 or level < 1700:
+        color_temp = self._model_info.lamps[
+            YeelightSubLightType.Main
+        ].color_temp  # for now only main light support
```
rytilahti, line 368: As this doesn't change after the initialization, it'd be cleaner to initialize this already inside `__init__`?
```
self._main_light = self._model_info.lamps[..]
self._color_temp_range = self._main_light.color_temp
```

Kirmas: @rytilahti Please look I have added light_type as constructor property and naw I can add something like `miiocli yeelight Background --ip 0.0.0.0 --token ======= set_color_temp 4000` in the next PR. What do you think?

miio/integrations/yeelight/specs.yaml
rytilahti: If you happen to know the names, we may want to add them to the yaml file some other point. I'm hoping that we could use the information derived from implementation classes (i.e., an extended version of `_supported_models`) to generate the list of supported devices to the readme/docs.

miio/integrations/yeelight/__init__.py
rytilahti: The code looks good to me, adding an option of sorts can be discussed later on :)

miio/integrations/yeelight/__init__.py
```diff
@@ -265,13 +265,17 @@ def __init__(
         debug: int = 0,
         lazy_discover: bool = True,
         model: str = None,
+        light_type: YeelightSubLightType = YeelightSubLightType.Main,  # for now only Main support this is for future PR.
```
rytilahti, line 268:
```suggestion
```
Please remove this for now, we don't want to extend the constructor signature for any more special cases.

miio/integrations/yeelight/__init__.py
```diff
@@ -265,13 +265,17 @@ def __init__(
...
     ) -> None:
         super().__init__(ip, token, start_id, debug, lazy_discover, model=model)
         if Yeelight._spec_helper is None:
             Yeelight._spec_helper = YeelightSpecHelper()
             Yeelight._supported_models = Yeelight._spec_helper.supported_models
 
         self._model_info = Yeelight._spec_helper.get_model_info(self.model)
+        self._light_type = light_type
```
rytilahti, line 276:
```suggestion
        self._light_type = YeelightSubLightType.Main
```

Kirmas: Ok, I am deleting this. But there are questions about this, can I do this in the future PR? 

As we said in the previous PR (https://github.com/rytilahti/python-miio/pull/1166#discussion_r741493032)
> * The main `Yeelight` instance controls the main light.
> * The sub lights (be it night light, or background one) use the same API as the main one, and they are exposed through the API in an easy-to-use manner. Potential downstreams like homeassistant should be able to have a single entity implementation, and the only difference between the main light and the sub lights should be based on the available features.

As I understand we should have the same API class for main and background lighting. And I don't understand how to do this without extending the constructor. Could you give me some advise (Then I`ll go correct way in the future PR)?

miio/integrations/yeelight/specs.yaml
Kirmas: > 

You mean added something like: 
```
yeelink.light.strip2:
          display_name: LED Light Strip 1S
          ...................
```

Kirmas:
> Thanks @Kirmas :+1:
@rytilahti Are you waiting for something additional from me in this PR? Or isn't time to merge? If second ok. But if first tell me.

rytilahti:
Ah, sorry, I got distracted with some other things and forgot to merge it. I think I was about to answer to your question regarding to how to do multiple lights without modifying the yeelight constructor. I'd say the simplest solution would be creating a separate class that handles the light-specific wrappings (while using the "parent" yeelight for communications).

PR #1254, opened 2021-12-15
Add support for Dreame Vacuum MB1808 1C

georgievgeorgi:
The full robot API is implemented in a dictionary which is interpreted
by helper functions in DreameVacuumBase. There is also an example which
allows to navigate the robot by the keyboard arrows.
Even though the coding conventions of the project are not followed, this
contribution might be helpful for the project.

### DIFF
--- a/examples/dreame_vacuum_mb1808.py
+++ b/examples/dreame_vacuum_mb1808.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+from getkey import getkey, keys
+from miio import DreameVacuumMB1808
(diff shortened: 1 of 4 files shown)


augustas2 (approved):

rytilahti:
Hi and sorry for the late response! Please check out my comment on another dreame model at https://github.com/rytilahti/python-miio/pull/1290#issuecomment-1009411484 – it would be better to consolidate this PR similarly to have a single dreame miot implementation :-)

PR #1276, opened 2021-12-27
Make zhimi.fan.za5 compatible to model autodetection

syssi:
Fixes https://github.com/syssi/xiaomi_fan/issues/176

### DIFF
--- a/miio/fan_miot.py
+++ b/miio/fan_miot.py
@@ -680,8 +680,7 @@ class FanZA5(MiotDevice):
         lazy_discover: bool = True,
         model: str = MODEL_FAN_ZA5,
     ) -> None:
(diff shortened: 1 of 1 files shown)


miio/fan_miot.py
```diff
@@ -680,8 +680,7 @@ def __init__(
         lazy_discover: bool = True,
         model: str = MODEL_FAN_ZA5,
     ) -> None:
-        super().__init__(ip, token, start_id, debug, lazy_discover)
```
rytilahti, line 683: Hmm, why not simply remove the whole constructor as it's not doing anything besides calling the base class one?

rytilahti:
Is this still relevant?

PR #1287, opened 2022-01-09
Add support for Air Purifier 4 Pro

ymj0424:
This PR adds support for Air Purifier 4 Pro (zhimi.airp.va2)
[https://home.miot-spec.com/spec/zhimi.airp.va2](https://home.miot-spec.com/spec/zhimi.airp.va2)


laz-:
I unboxed a 4 Pro yesterday, snagged the token, and was prepping to do this. Luckily, I looked in the PR section and found this. i pulled your PR down and ran it :)

My status output looks a little off:
```shell
(python-miio-dHNQd-W6-py3.10) ➜  python-miio git:(ymj0424/master) ✗ python3 ./miio/cli.py airpurifiermiot --ip $IP --token $TOKEN info
WARNING:miio.device:Found an unsupported model 'zhimi.airp.va2' for class 'AirPurifierMiot'. If this is working for you, please open an issue at https://github.com/rytilahti/python-miio/
Model: zhimi.airp.va2
Hardware version: esp32
Firmware version: 2.2.1
(python-miio-dHNQd-W6-py3.10) ➜  python-miio git:(ymj0424/master) ✗ python3 ./miio/cli.py airpurifiermiot --ip $IP --token $TOKEN status
WARNING:miio.device:Found an unsupported model 'zhimi.airp.va2' for class 'AirPurifierMiot'. If this is working for you, please open an issue at https://github.com/rytilahti/python-miio/
Power: off
AQI: None μg/m³
Average AQI: 2 μg/m³
Humidity: 22.4 %
Temperature: None °C
Fan Level: 3
Mode: OperationMode.Silent
LED: None
LED brightness: LedBrightness.Dim
Buzzer: None
Buzzer vol.: None
Child lock: None
Favorite level: None
Filter life remaining: 25 %
Filter hours used: None
Use time: 80:72:13:22:9b:dd:4 s
Purify volume: None m³
Motor speed: None rpm
Filter RFID product id: None
Filter RFID tag: None
Filter type: None
```
There's a lot in there that looks suspect. Have you tested the code against an actual device?

laz-:
Ahhh... disregard me. I was calling it with the wrong command, thus the WARNING :/

Much better:
```shell
(python-miio-dHNQd-W6-py3.10) ➜  miio git:(ymj0424/master) ✗ python3 ./cli.py airpurifierva2 --ip $IP --token $TOKEN status
Power: on
Anion: True
AQI: 11 μg/m³
Average AQI: 17 μg/m³
Purify volume: 1275 m³
Humidity: 29 %
Temperature: 22.4 °C
Fan Level: 1
Mode: OperationMode.Fan
LED brightness level: Brightness.Brightest
Buzzer: True
Child lock: False
Filter life remaining: 99 %
Filter hours used: 25
Filter left time: 207 days
Motor speed: 860 rpm
Favorite RPM: None rpm
Favorite level: 11
```

miio/airpurifier_miot.py
```diff
@@ -80,6 +82,37 @@
     "favorite_rpm": {"siid": 9, "piid": 3},
 }
 
+# https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:air-purifier:0000A007:zhimi-va2:2
+_MODEL_AIRPURIFIER_VA2 = {
```
rytilahti, line 86: Please start converting the `_MAPPING` to be a dict of mappings, so that we can avoid having multiple different interfaces in the future. This should allow dropping the separate classes for both AirpurifierMiot as well as for the status container, while giving us a common interface among models.

ymj0424: If all models use a common interface, how to set or get a model-specific property ?

rytilahti: The interface can be extended case-by-case basis. So for getters (so mostly related to status container), the way is to return `None` if the value is not available. And for setters, the model can be checked (`self.model`) to raise an exception on non-supported devices.

ymj0424: I met some problems during coverting the `_MAPPING` to be a dict of mappings.
Here's the dict of mappings
```python
_MODEL_MAPPING = {
    "zhimi.airpurifier.ma4": _MAPPING,  # airpurifier 3
    "zhimi.airpurifier.mb3": _MAPPING,  # airpurifier 3h
    "zhimi.airpurifier.va1": _MAPPING,  # airpurifier proh
    "zhimi.airpurifier.vb2": _MAPPING,  # airpurifier proh
    "zhimi.airpurifier.mb4": _MAPPING_MB4,  # airpurifier 3c
    "zhimi.airp.mb4a": _MAPPING_MB4,  # airpurifier 3c
    "zhimi.airp.va2": _MAPPING_VA2,  # airpurifier 4 pro
}
```
However, how to specify the mapping in `AirPurifierMiot` class ?
self.model is not valid here, and self.model is an optional parameter even if I specify the mapping in a `__init__` function
```python
class AirPurifierMiot(BasicAirPurifierMiot):
    """Main class representing the air purifier which uses MIoT protocol."""

    mapping = _MODEL_MAPPING[self.model]
    _supported_models = _MODEL_MAPPING.keys()
```

rytilahti: Since #1302, you just need to define `_mappings` and all relevant methods use the correct mapping in the background, see https://github.com/rytilahti/python-miio/blob/master/miio/heater_miot.py#L141 for an example.

miio/__init__.py
```diff
@@ -29,7 +29,7 @@
 from miio.airhumidifier_mjjsq import AirHumidifierMjjsq
 from miio.airpurifier import AirPurifier
 from miio.airpurifier_airdog import AirDogX3, AirDogX5, AirDogX7SM
-from miio.airpurifier_miot import AirPurifierMB4, AirPurifierMiot
+from miio.airpurifier_miot import AirPurifierMiot
```
rytilahti, line 32: Let's keep AirPurifierMB4 as a proxy with deprecated warning for a while, just in case someone is using that.

Please add a class to `miio.airpurifier_miot` module, something like this:
```
from miio.utils import deprecated

class AirPurifierMB4(AirPurifierMiot):
    @deprecated("Use AirPurifierMiot")
    def __init__(*args, **kwargs):
        super().__init__(*args, **kwargs)
```

rytilahti (commented):
Great cleanup on merging the implementations and tests! :1st_place_medal: 

I have just a very minor nitpick wrt deprecating the previously exported class, just in case someone happens to be using it.

miio/airpurifier_miot.py
```diff
@@ -134,102 +195,76 @@ def mode(self) -> OperationMode:
...
     def filter_life_remaining(self) -> int:
         """Time until the filter should be changed."""
-        return self.data["filter_life_remaining"]
+        return self.data.get("filter_life_remaining", None)
 
     @property
     def filter_hours_used(self) -> int:
         """How long the filter has been in use."""
-        return self.data["filter_hours_used"]
+        return self.data.get("filter_hours_used", None)
```
rytilahti, line 217:
```suggestion
        return self.data.get("filter_hours_used")
```
`get()` defaults to None, so no need to describe that separately.

Please update the type hint to use `Optional[]` for all the cases that are optional.

miio/airpurifier_miot.py
```diff
@@ -134,102 +195,76 @@ def mode(self) -> OperationMode:
     @property
     def buzzer(self) -> Optional[bool]:
         """Return True if buzzer is on."""
-        if self.data["buzzer"] is not None:
-            return self.data["buzzer"]
+        buzzer = self.data.get("buzzer", None)
+        if buzzer is not None:
+            return buzzer
```
rytilahti, line 200:
```suggestion
        return self.data.get("buzzer")
```

miio/airpurifier_miot.py
```diff
@@ -134,102 +195,76 @@ def mode(self) -> OperationMode:
...
     def child_lock(self) -> bool:
         """Return True if child lock is on."""
-        return self.data["child_lock"]
+        return self.data.get("child_lock", None)
 
     @property
     def filter_life_remaining(self) -> int:
         """Time until the filter should be changed."""
-        return self.data["filter_life_remaining"]
+        return self.data.get("filter_life_remaining", None)
```
rytilahti, line 212: All "remaining" properties should return `timedelta()`s, but considering this has been already like this it's not a big deal, but this API will likely get consolidated at latest when purifiers get a common interface.

miio/airpurifier_miot.py
```diff
@@ -134,102 +195,76 @@ def mode(self) -> OperationMode:
...
         """Return True if LED is on."""
-        return self.data["led"]
+        return self.data.get("led", None)
 
     @property
     def led_brightness(self) -> Optional[LedBrightness]:
         """Brightness of the LED."""
-        if self.data["led_brightness"] is not None:
+
+        value = self.data.get("led_brightness", None)
```
rytilahti, line 262:
```suggestion
        value = self.data.get("led_brightness")
```

rytilahti (approved):
Ready to be merged as soon as the tests pass, thanks! :1st_place_medal: 

I'm unsure what caused the CI not to use the pinned requirements for tests, likely causing those unrelated flake8 warnings, but could you fix those still in this PR?

LintHart:
Who do we need to beg to get this to show up in the home assitant integration now that it has been updated with 0.5.10?

rytilahti:
I have a local draft that will adapt the homeassistant integration to use the library provided supported models to avoid that in the future, but until that's done someone needs to create a PR to add it to the list of supported models.

Issue #1217, opened 2021-12-08
New device: Dreame Bot Z10 Pro (dreame.vacuum.p2028)


odelma:
**Device information:**

  - Dreame Bot Z10 Pro (robot vacuum cleaner)
  - Link: https://www.dreametech.com/products/dreame-z10-pro

Use `miiocli device --ip <ip address> --token <token>`.

   NAME:     Dreame Bot Z10 Pro
   ID:       447472770
   MODEL:    dreame.vacuum.p2028

**Additional context**

It seems there are working integrations that may prove to be helpful:

https://github.com/pooyashahidi/xiaomi_vacuum
https://github.com/openhab/openhab-addons/commit/b6836b8ef913a81c206521818f8d8795e98bd87c

FM84DEV:
I also have this robot. 
It will be great to add this device....
I can provide some help but someone has to tell me what to do because I have a basic knowledge in programming....

rytilahti:
Have you configured the rooms in the app? That response is coming from the device, so there isn't much we can do here..

Are other commands working fine?

FM84DEV:
Yes I configure them in the app but Z10 do not response.
Looking around I found a way to get the rooms ids using the timers in the app and get them via the miio via a specific function...
Then I added also the support for room cleaning in the last commit.
:-)

messyjoes1:
@FM84DEV can you share your code to get the room ids via schedule and the code to start cleaning the room by id ?

christianduddd:
> It's already shared. You can download it here: https://github.com/FM84DEV/python-miio/blob/feature/dreamef9_support/miio/integrations/vacuum/dreame/dreamevacuum_miot.py
> 
> To know the rooms ids you have to make a timer in the App at 12:00 selecting all the rooms in a specific order and disable it. Then you call the function "get_rooms_ids_from_timers" that return the ids of the rooms in the same order you have specified. So you can associate your rooms with your ids. I see that if a new room is discovered the IDs could change.... But this is very uncommon! ;-)
> 
> Then you can use: start_clean_rooms _rooms_params_ where the argument _rooms_params_ is something like this [[1,1,2,1]] for one room or like this [[1,1,2,1],[2,1,2,1],...] for more. The number in order are: roomid, repeats, powerlevel, waterlevel.

Hey Thank you very much for pointing that out.

I'm very new to python-miio so can you please explain how exactly I call the get_rooms_ids_from_timers functions from the cli?

Kind regards,
Chris

[...]

PR #1290, opened 2022-01-10
Dreame F9 Vacuum (dreame.vacuum.p2008) support

peleccom:
This PR adds support for Dreame F9 Vacuum (dreame.vacuum.p2008)
https://home.miot-spec.com/spec?type=urn%3Amiot-spec-v2%3Adevice%3Avacuum%3A0000A006%3Adreame-p2008%3A1


rytilahti:
Hi and thanks for the PR! Have you tried to check how much does this differ from the current dreame implementation? If it's just about different mappings (and some other extra features), it would be better to extend that instead of having a separate implementation for this.

peleccom:
@rytilahti Hi. It has different mapping and some different statuses.  Do you mean we should create subclass for Dreame F9 in that case?

rytilahti:
@peleccom hi! Okay, so it would be better to modify the existing implementation to have a dictionary of mappings (e.g., https://github.com/rytilahti/python-miio/blob/master/miio/fan_miot.py#L17) and using the same keys wherever possible.

I see that the current dreame implementation is not using `call_action` for the actions, likely because the wasn't available when the integration was created.. So, the way I see this should be done is to adjust the existing mc1808 mapping to include the necessary actions (as yours does) and consolidate the code to use `call_action` for everything.

If there are some functions that are available only on p2008, you can check for the model (`self.model`) and raise an exception to inform that the functionality is not available. The plan is to have some common interfaces to make it easier to add homeassistant integration notwithstanding that devices are produced by different vendors.

Now for the status container, we can extend that to cover new features of p2008, and it's fine to return `None` if the data is not available due to a different mapping.
About the status codes themselves, the best approach depends on whether we can re-use the existing mapping at all or not. Are the common codes (or statuses) being shared by both models, or do the differ so much that this would require a separate status code class?

peleccom:
Ok. Now it share the same implementation as Dreame 1C.

https://github.com/rytilahti/python-miio/pull/1254 as a rewrite of Dreame 1C implementation looks good and simplify code much. But it is very different from other implementation. Also current state doesn't allow to use commands from cli tool as I can see. 

Also I want to ask if it is some progress with  universal interface for all vacuum implementations? Can I help with something?

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -190,12 +260,189 @@ def volume(self) -> str:
...
+    @property
+    def total_clean_times(self) -> str:
+        return self.data["total_clean_times"]
+
+    @property
+    def total_clean_area(self) -> str:
+        return self.data["total_clean_area"]
+
+
+class Dreame1CVacuumStatus(DreameVacuumStatusBase):
```
rytilahti, line 292: Please use a single status container. For the not-supported information, you can mark them as optional:
```
@property
def water_flow(self) -> Optional[WaterFlow]:
    if "water_flow" not in self.data:
        return None
    ...
```

This way we will have a single API for all Dreame devices (much like we have for Viomi and Roborock) that'll make it easier to consolidate the APIs among vendors in the future.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -46,14 +115,22 @@ class ChargingState(Enum):
...
     Medium = 2
     Strong = 3
 
 
+class CleaningModeDreameF9(Enum):
+    Unknown = -1
+    Quiet = 0
+    Standart = 1
+    Strong = 2
+    Turbo = 3
```
rytilahti, line 131: This is similar to fan speed setting on other vacuum integrations, right? Following the API from roborock integration (`fan_speed_presets() -> Dict[str, int]` to return the available modes) is the way to go. You can check the model inside the class using `self.model` to find out which presets to return.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -82,7 +159,20 @@ class DeviceStatus(Enum):
...
+    Unknown = -1
+    Low = 1
+    Medium = 2
+    High = 3
+
+
+class WaterTankStatus(Enum):
+    Unknown = -1
+    NotAttached = 0
+    Attached = 1
```
rytilahti, line 172: I'd just use boolean here. Roborock is currently using `is_water_box_carriage_attached` which is not a great name for this, though.. :-P

rytilahti (changes requested):
A brief review round:
* Please avoid separate container & implementation classes, as this makes it just harder to prepare to move using common interfaces.
* `DreameVacuumMiot` would be the main class, which supports both models, you can adjust the behavior inside the class by leveraging the model information (`self.model`).

miio/integrations/vacuum/dreame/tests/test_dreamef9vacuum_miot.py
```diff
@@ -0,0 +1,97 @@
+from unittest import TestCase
```
rytilahti, line 1: Ah, forgot to add. As the implementation is the same, you don't need a separate test file but you can test only the extra features as you see fit.

rytilahti:
> Ok. Now it share the same implementation as Dreame 1C.

Looks already much better! :+1:

> #1254 as a rewrite of Dreame 1C implementation looks good and simplify code much. But it is very different from other implementation. Also current state doesn't allow to use commands from cli tool as I can see.

Ah, you are right. It's better to get your PR done first, and then consider adding support for other dreame devices to the integration.

> Also I want to ask if it is some progress with universal interface for all vacuum implementations? Can I help with something?

Alas, there hasn't been much going on on that front, I simply don't have time to work on it currently. The first step to make that happen would likely involve creating those interfaces (for `Vacuum` and `VacuumStatus`) and start porting all integrations over, starting from small with only very basic functionality and extending that over time.

rytilahti:
Hi @peleccom, I realized that there is no facilities to handle mappings dictionaries without fuss.. I created #1302 to improve the situation, that should make it simpler to implement the required changes, please let me know if that helps :-)

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -190,12 +256,115 @@ def volume(self) -> str:
...
+                return CleaningModeDreameF9.Unknown
 
-    mapping = _MAPPING
-    _supported_models = ["dreame.vacuum.mc1808"]
+    @property
+    def life_sieve(self) -> Optional[str]:
+        try:
+            return self.data["life_sieve"]
+        except KeyError:
+            return None
```
rytilahti, line 308:
```suggestion
    return self.data.get("life_sieve")
```
It's better to simply use `get` if there is no specific reason to log a warning or do something else.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -190,12 +256,115 @@ def volume(self) -> str:
...
+
+
+class DreameVacuumBase(MiotDevice):
+    _supported_models = [
+        DREAME_1C,
+        DREAME_D9,
+        DREAME_F9,
+    ]
+
+    mapping = MIOT_MAPPING[DREAME_1C]
```
rytilahti, line 352:
```suggestion
    mapping = MIOT_MAPPING
```
It is enough to define a dict of mappings, the correct mapping should be chosen now since #1302 is merged. If this isn't working, let me know and let's find a solution.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -190,12 +256,115 @@ def volume(self) -> str:
...
+    @property
+    def is_water_box_carriage_attached(self) -> Optional[bool]:
+        """Return True if water box carriage (mop) is installed, None if sensor not
+        present."""
+        if "water_box_carriage_status" in self.data:
+            return self.data["water_box_carriage_status"] == 1
+        return None
+
+
+class DreameVacuumBase(MiotDevice):
```
rytilahti, line 345:
```suggestion
class DreameVacuum(MiotDevice):
```

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -210,17 +379,33 @@ class DreameVacuumMiot(MiotDevice):
             "Filter left level: {result.filter_left_time}\n"
             "Filter life level: {result.filter_life_level}\n"
             "Life brush main: {result.life_brush_main}\n"
-            "Life brush side: {result.life_brush_side}\n"
-            "Life sieve: {result.life_sieve}\n"
+            if hasattr(result, "life_brush_main")
+            else "" "Life brush side: {result.life_brush_side}\n"
```
rytilahti, line 383: IMO, the container should always have this, and just return `None` if there is no value.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -233,51 +418,120 @@ def status(self) -> DreameVacuumStatus:
...
+        )
+
+
+class DreameVacuum(DreameVacuumBase):
+    """Interface for Vacuum 1C STYTJ01ZHM (dreame.vacuum.mc1808)"""
+
+
+Dreame1CVacuum = DreameVacuum
+DreameF9Vacuum = DreameVacuum
+DreameVacuumMiot = DreameVacuum
```
rytilahti, line 537: Please just rename `DreamVacuumBase` to `DreameVacuum`. You can leave `DreameVacuumMiot` or create a class that just inherits from the `DreameVacuum` and is decorated with `@deprecated()` just to let any potential users that it's been renamed, but it's not worth adding new aliases.

In practice, this means that using `miiocli dreamevacuum` should work directly on all supported vacuums.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -190,12 +256,115 @@ def volume(self) -> str:
...
+
+    mapping = MIOT_MAPPING[DREAME_1C]
+
+    def __init__(self, *args, **kwargs) -> None:
+        model = kwargs.get("model")
+        if model:
+            mapping = MIOT_MAPPING.get(model)
+        if mapping:
+            self.mapping = mapping
+        super().__init__(*args, **kwargs)
```
rytilahti, line 360: This is not needed, `MiotDevice` class takes care of using the correct mapping for methods since that recent PR.

peleccom: @rytilahti  yep. It is not 100% ready yet. It was changed after merge. I will write a message when this pr will be ready

peleccom:
@rytilahti Please check

README.rst
```diff
@@ -111,6 +111,8 @@ Supported devices
 -  Xiaomi Mijia 360 1080p
 -  Xiaomi Mijia STYJ02YM (Viomi)
 -  Xiaomi Mijia 1C STYTJ01ZHM (Dreame)
+-  Dreame F9
+-  Dreame D9
```
rytilahti, line 115:
```suggestion
-  Dreame F9, D9
```

miio/integrations/vacuum/dreame/dreamevacuum.py
```diff
@@ -0,0 +1,589 @@
+"""Dreame Vacuum."""
```
rytilahti, line 1: Please avoid renaming the file for now, as it makes it really hard to read the diffs.

miio/miot_device.py
```diff
@@ -79,10 +79,11 @@ def get_properties_for_mapping(self, *, max_properties=15) -> list:
     )
     def call_action(self, name: str, params=None):
         """Call an action by a name in the mapping."""
-        if name not in self.mapping:
+        mapping = self._get_mapping()
```
rytilahti, line 82: Please don't modify this file right now, but use the `status()` to access the information if necessary.

We can extend this (core) interface in the future, but that'll require some thought if it's really necessary, as well as adding unit tests for the new functionality.

peleccom: It looks like you have misplaced this comment. I haver removed `get_property` implementation and replaced usage with `self.status()` call. 

But with last changes with new `_mappings` field this change in `call_action` required. Otherwise it will try to access `self.mapping` which is not defined

rytilahti: Oh, I think I commented on a single commit as clicking on the notification link brought me there, my bad...

Thanks for spotting and fixing that!

miio/integrations/vacuum/dreame/tests/test_dreamef9vacuum_miot.py
rytilahti: Please remove this file and just test the extra features inside the already existing test file. I know that the tests are not easy to write and that the already existing tests are not up to par, but if you could just include a test for the `status()` to verify that it's working correctly for both implementations, that would be of great help!

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -190,37 +274,132 @@ def volume(self) -> str:
...
+    def water_flow(self) -> Optional[WaterFlow]:
+        try:
+            water_flow = self.data["water_flow"]
+        except KeyError:
+            return None
+        try:
+            return WaterFlow(water_flow)
+        except ValueError:
+            _LOGGER.error("Unknown WaterFlow (%s)", self.data["water_flow"])
+            return WaterFlow.Unknown
```
rytilahti, line 344:
```suggestion
        try:
            return WaterFlow(self.data["water_flow"])
        except KeyError:  # device does not report water flow
            return None
        except ValueError:
            _LOGGER.error("Unknown waterflow reported: %s", self.data["water_flow"])
            return None
```
I think it's better to log an error and return 'None` is better than having a separate unknown state. From the perspective of a library user, it is easier to check for `None` and the warning will give away that something unexpected has happened.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -190,37 +274,132 @@ def volume(self) -> str:
...
+        DREAME_F9,
+    ]
+    _mappings = MIOT_MAPPING
+
+    def get_fanspeeds(self):
+        """Return fanspeeds enum for model if found or None."""
+        if self.model == DREAME_1C:
+            return CleaningModeDreame1C
+        elif self.model in [DREAME_F9, DREAME_D9]:
+            return CleaningModeDreameF9
```
rytilahti, line 368:
```suggestion
    def fan_speed_presets(self) -> Dict[str, int]:
        """Return fanspeeds enum for model if found or None.""" 
        if self.model == DREAME_1C:
            return _enum_as_dict(CleaningModeDreame1C)
        elif self.model in [DREAME_F9, DREAME_D9]:
            return _enum_as_dict(CleaningModeDreameF9)
            
        _LOGGER.warning("Unknown model, falling back to %r", CleaningModeDreame1C)
        return _enum_as_dict(CleaningModeDreame1C)
```
I'm not sure what should be used here as a fallback, but renaming this method (as well as the `set_` one) will make it compatible with the roborock implementation that will simplify generalizing these.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -230,54 +409,179 @@ def status(self) -> DreameVacuumStatus:
...
+        :param int speed: Fan speed to set
+        """
+        fanspeeds_enum = self.get_fanspeeds()
+        fanspeed = None
+        if not fanspeeds_enum:
+            return
+        try:
+            fanspeed = fanspeeds_enum(speed)
+        except ValueError:
+            return None
```
rytilahti, line 485:
```suggestion
```
No need to check if the speed exists as trying to set the speed should error out if it errors.

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -230,54 +409,179 @@ def status(self) -> DreameVacuumStatus:
...
+        mapping = self._get_mapping()
+        if "water_flow" not in mapping:
+            return None
+        waterflow = None
+        try:
+            waterflow = WaterFlow(value)
+        except ValueError:
+            return None
+        click.echo(f"Setting waterflow to {waterflow.name}")
+        return self.set_property("water_flow", waterflow.value)
```
rytilahti, line 522: Do you think these are necessary when the information is already available when you always call `status()` and access it from there?

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
```diff
@@ -66,25 +157,38 @@ class OperatingMode(Enum):
...
+
+def _enum_as_dict(cls):
+    return {x.name: x.value for x in list(cls)}
+
+
 class DreameVacuumStatus(DeviceStatusContainer):
-    def __init__(self, data):
+    def __init__(self, data, model):
         self.data = data
+        self.model = model
```
rytilahti, line 191: Would you mind adding a docstring containing the contents of the input data?

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
peleccom: I've checked this and Dreame just return return code == 0.  So it is beeter just check that speed value exists in enum

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
peleccom: `de waterflow(self)`It is aligned with roborock implementation

rytilahti: Okay, let's keep it then for now :+1:

miio/integrations/vacuum/dreame/tests/test_dreamef9vacuum_miot.py
```diff
@@ -0,0 +1,98 @@
...
+    ChargingState,
+    CleaningModeDreameF9,
+    DeviceStatus,
+    FaultStatus,
+    OperatingMode,
+    WaterFlow,
+)
+
+_INITIAL_STATE = {
+    "battery_level": 42,
```
rytilahti, line 19: Would you mind combining all the tests under the existing test file? See #1287's `miio/tests/test_airpurifier_miot.py` for a nice example.

Also, the "state" should be a representation of the device response, i.e., there should be no enums in those values while the property getters should be checked to return the correct information.

miio/integrations/vacuum/dreame/tests/test_dreame1cvacuum_miot.py
```diff
@@ -25,7 +26,7 @@
     "brush_left_time2": 187,
     "brush_life_level2": 57,
     "operating_mode": OperatingMode.Cleaning,
-    "cleaning_mode": CleaningMode.Medium,
+    "cleaning_mode": CleaningModeDreame1C.Medium,
```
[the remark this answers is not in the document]
peleccom: I don't think it is possible. The two models have different names for the same mode values. So it is not posssible

```

CleaningModeDreame1C
    Quiet = 0
    Default = 1
    Medium = 2
    Strong = 3


CleaningModeDreameF9
    Quiet = 0
    Standart = 1
    Strong = 2
    Turbo = 3
```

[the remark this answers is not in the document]
rytilahti: Hmmh, yes, that will indeed require different handling... roborock integration currently handles this by only having a [percentage based setting](https://github.com/rytilahti/python-miio/blob/master/miio/integrations/vacuum/roborock/vacuum.py#L602) and exposes the [available fan speeds](https://github.com/rytilahti/python-miio/blob/master/miio/integrations/vacuum/roborock/vacuum.py#L62-L98) via [`fanspeed_presets`](https://github.com/rytilahti/python-miio/blob/master/miio/integrations/vacuum/roborock/vacuum.py#L616).

I don't have a direct answer how this should be done the best at this moment, but as we are slowly moving towards generic interfaces this should be handled differently to make it simple to use for the downstream users.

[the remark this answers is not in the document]
rytilahti: Ah, sorry, that's already handled properly in your code.. And thanks also for the test cleanup, that'll really make it simpler to add support for more dreame devices in the future!

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
rytilahti: Oh, I meant the data that gets passed in, to help the future readers to understand the contents. See https://github.com/rytilahti/python-miio/blob/master/miio/integrations/humidifier/deerma/airhumidifier_jsqs.py#L47 for an example.

rytilahti:
Btw, looks like another model (dreame.vacuum.p2028) is also using the same command set: https://github.com/rytilahti/python-miio/issues/1217#issuecomment-1027843395

peleccom:
@rytilahti Fine. I've added Dreame Z10 Pro  (dreame.vacuum.p2028) to the list of supported models.
Can we merge PR  now?

miio/integrations/vacuum/dreame/dreamevacuum_miot.py
rytilahti: Alternatively, as the tests contains the raw payloads, simply remove this docstring.

rytilahti:
@peleccom this is just pending on the status container docstring as I mentioned earlier: https://github.com/rytilahti/python-miio/pull/1290#discussion_r794078604 - would you please handle that? :-)

rytilahti (approved):
Looks great, thanks for the PR @peleccom! Please rebase on top of the current master and I'll merge it.

PR #1301, opened 2022-01-16
Add support for zhimi.heater.za2

PRO-2684:
This PR adds support to zhimi.heater.za2.
`heater_miot_za2.py`'s code is modified from `heater_miot.py`. Compared to already supported mc2 model, za2 can monitor relative humidity, so I added `relative_humidity` function.
P.S. Comment is wrong at `heater_miot.py`, line 21. Should be `# Physical Control Locked (siid=5)`.


rytilahti:
Please just extend the existing implementation as they are from the same manufacturer and they do not differ that much. You can do that by basing your PR on top of #1302, and introducing `_mappings` that is a dict containing model-keyed mappings. The correct mapping will then be chosen automatically run-time :-)

For the extra `relative_humidity`, you can check if the value is there and mark it as `Optional` in the container to keep the API for both types of models:
```
    @property
    def relative_humidity(self) -> int:
        """Current relative humidity."""
        return self.data.get("relative_humidity")
```

PRO-2684:
On za2, `LedBrightness` 0 means "On (Bright)", 1 means "On (Dim)", 2 means "Off". How should this conflict be solved?

PRO-2684:
@rytilahti This should work now. But as I modified the class `LedBrightness`, users may encounter compatibility issues.

miio/heater_miot.py
```diff
@@ -9,32 +9,66 @@
...
 
 
 class LedBrightness(enum.Enum):
-    On = 0
-    Off = 1
+    """Xiaomi Smart Space Heater S (zhimi.heater.mc2): 
+    0 = On, 1 = Off
+
+    Xiaomi Smart Space Heater 1S (zhimi.heater.za2): 
+    0 = On (Bright), 1 = On (Dim), 2 = Off
```
rytilahti, line 65: This is confusing API-wise.. How about just adding a `Dim = 2` to the original enum, and performing the conversion per-model basis (and warn about unsupported `Dim` for mc2?

miio/heater_miot.py
```diff
@@ -85,6 +119,11 @@ def temperature(self) -> float:
         """Current temperature."""
         return self.data["temperature"]
 
+    @property
+    def relative_humidity(self) -> Union[int, None]:
```
rytilahti, line 123:
```suggestion
    def relative_humidity(self) -> Optional[int]:
```

rytilahti (changes requested):
Just two more things and this is good to go!

PRO-2684:
I've run the tests and it shall be fine now.

rytilahti (approved):
Please fix the tests and update README.md, and then this is good to go! :tada:

rytilahti:
Oh, looks like the tests are still failing. Btw, you can also run the tests locally using executing `pytest` or `tox` :-)

PRO-2684:
The reason why the tests are not passed is that the dummy test cases are based on mc2, but I chose to fall back on the strictest restriction in times of unrecognized models, including dummy.

rytilahti:
This is good to go, thanks @PRO-2684!

Issue #1171, opened 2021-10-31
Support for `shuii.humidifier.jsq002` humidifier


iromeo:
Before submitting a new request, use the search to see if there is an existing issue for the device.

**Device information:**

  - Name(s) of the device: `shuii.humidifier.jsq002`
  - Link:

Use `miiocli device --ip <ip address> --token <token>`.

  - Model: `shuii.humidifier.jsq002`
  - Hardware version: `<I will provide it later>`
  - Firmware version: `<I will provide it later>`

**Additional context**
Some time ago I added `shuii.humidifier.jsq001` support. Now I'm going to make a PR for `shuii.humidifier.jsq002` device support. This issue is a first step for supporting it in HomeAssistant, as part of https://github.com/syssi/xiaomi_airpurifier/issues/50

wxy8866:
I have this device. Is there any chance to get it work?

iromeo:
@wxy8866 I'm working on a PR in my spare time, not ready yet.

MJeka:
Hi. I have the same problem.

iromeo:
@MJeka @wxy8866 

Hi guys, I implemented code to support this device, so e.g. it could be used for further integration with Home-Assistant (https://github.com/syssi/xiaomi_airpurifier/issues/50) or some other home automation hub. But before opening a pull request and merging it into `python-miio` we need to test my fix. I have only prev model `shuii.humidifier.jsq001`, so I cannot test `shuii.humidifier.jsq002`, they have different commands and response params.

In order to test my integration with device, you need to:

1) I assume you are familiar with command line and have git and python installed. It is easier to make it using Mac OS or Linux. But windows is also ok.
2) Also I assume you already have a token for your device, e.g you could use https://github.com/Squachen/micloud project or https://python-miio.readthedocs.io/en/latest/discovery.html
3) Clone my repo with fix and install miiocli tool from sources (with my changes). Here steps are similar to https://python-miio.readthedocs.io/en/latest/discovery.html#installation but using my repo.
```
git clone --branch features/1171_jsq002_support https://github.com/iromeo/python-miio.git python-miio-iromeo
cd python-miio-iromeo

poetry install
``` 
4) List of available commands:

```
$ poetry run miiocli airhumidifierjsq002 --help
Usage: miiocli airhumidifierjsq002 [OPTIONS] COMMAND [ARGS]...

Options:
  --ip TEXT     [required]
  --token TEXT  [required]
  --model TEXT
[...]

PR #1318, opened 2022-02-06
Support for Zero Fog DWZF(G)-4500Z (shuii.humidifier.jsq002)humidifier (#1171)

iromeo:
Hi, this is PR add support for Zero Fog DWZF(G)-4500Z (shuii.humidifier.jsq002)humidifier and fixes Issue #1171
New device support was tested by user on his device.

### DIFF
--- a/miio/airhumidifier_jsq.py
+++ b/miio/airhumidifier_jsq.py
@@ -12,19 +12,73 @@ _LOGGER = logging.getLogger(__name__)
 
 # Xiaomi Zero Fog Humidifier
 MODEL_HUMIDIFIER_JSQ001 = "shuii.humidifier.jsq001"
+# Xiaomi Zero Fog DWZF(G)-4500Z
+MODEL_HUMIDIFIER_JSQ002 = "shuii.humidifier.jsq002"
+
+# Array of several common commands
+SHARED_COMMANDS = {
+    MODEL_HUMIDIFIER_JSQ001: dict(
+        power_on_off="set_start",
+        buzzer_on_off="set_buzzer"
+    ),
+    MODEL_HUMIDIFIER_JSQ002: dict(
+        power_on_off="on_off",
+        buzzer_on_off="buzzer_on"
+    )
+}
 
 # Array of properties in same order as in humidifier response
 AVAILABLE_PROPERTIES = {
     MODEL_HUMIDIFIER_JSQ001: [
-        "temperature",  # (degrees, int)
-        "humidity",  # (percentage, int)
-        "mode",  # ( 0: Intelligent, 1: Level1, ..., 5:Level4)
-        "buzzer",  # (0: off, 1: on)
-        "child_lock",  # (0: off, 1: on)
-        "led_brightness",  # (0: off, 1: low, 2: high)
-        "power",  # (0: off, 1: on)
-        "no_water",  # (0: enough, 1: add water)
-        "lid_opened",  # (0: ok, 1: lid is opened)
+        # Example of 'Air Humidifier (shuii.humidifier.jsq001)' response:
+        # [24, 37, 3, 1, 0, 2, 0, 0, 0]
+        "temperature",  # 24 (degrees, int)
+        "humidity",  # 37 (percentage, int)
+        "mode",  # 3 ( 0: Intelligent, 1: Level1, ..., 5:Level4)
+        "buzzer",  # 1 (0: off, 1: on)
+        "child_lock",  # 0 (0: off, 1: on)
+        "led_brightness",  # 2 (0: off, 1: low, 2: high)
+        "power",  # 0 (0: off, 1: on)
+        "no_water",  # 0 (water level state  - 0: enough, 1: add water)
+        "lid_opened",  # 0 (0: ok, 1: lid is opened)
+    ],
+    MODEL_HUMIDIFIER_JSQ002: [
+        # Example of Xiaomi 'Zero Fog DWZF(G)-4500Z` (shuii.humidifier.jsq002)'
+        #   Model: shuii.humidifier.jsq002
+        #   Hardware version: ESP8266
+        #   Firmware version: 1.4.0
+        #
+        # > fast_set [bea, lock, temperature, humidity, led]
+        #           0/1,  0/1,          25,       50, 0/1]
+        #
+        #  Properties:
+        # > dev.send("get_props","")
+        #  [1, 2, 36, 2, 46, 4, 1, 1, 1, 50, 51, 0]
+
+        # res[0]=1 Values (0: off/sleep, 1: on);
+        "power",  # CMD: on_off [int]
+        # res[1]=2 Values (1, 2, 3); fan speed in UI: {Gear: level 1, level 2, level 3};
+        "mode",  # CMD: set_gear [int]
+        # res[2]=36 Value (is  % [int] );  Environment humidity;
+        "humidity",
+        # res[3]=2 Values (1, 2, 3) // Light 1 - Off; 2 - low; 3 - high
+        "led_brightness",  # CMD: set_led [int]
+        # res[4]=26 Values(is "ambient temp degrees int"+20, i.e. 46 corresponds to 26);
+        "temperature",
+        # res[5]=4 (0,1,2,3,4,5)
+        "water_level",  # Get cmd: corrected_water []
+        # res[6]=1, Water heater values (0: off, 1: on)
+        "heat",  # CMD: warm_on [int]
+        # res[7]=1 BeaPower values (0: off, 1: on)
+        'buzzer',  # CMD: buzzer_on [int]
+        # res[8]=1, Values (0: off, 1: on)
+        'child_lock',  # CMD: set_lock [int]
+        # res[9]=50 Values (water heater target temp in degrees, [int]: 30..60)
+        'target_temperature',  # CMD: set_temp [int]
+        # res[10]=51 Values (% [int])
+        'target_humidity',  # CMD: set_humidity [int]
+        # res[11]=0 Values: <Unknown> We failed to find when it changes, is not lid opened event
+        'reserved',  # XXX: cmd rst_clean [] ?
     ]
 }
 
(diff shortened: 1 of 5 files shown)


miio/airhumidifier_jsq.py
```diff
@@ -43,19 +90,21 @@ class LedBrightness(enum.Enum):
     High = 2
 
 
-class AirHumidifierStatus(DeviceStatus):
-    """Container for status reports from the air humidifier jsq."""
+class OperationModeJsq002(enum.Enum):
+    Level1 = 1
+    Level2 = 2
+    Level3 = 3
```
rytilahti, line 96: Let's re-use the existing enum for this, you can check for the model & raise an exception if an unsupported mode is tried to be set. We want to have a common interface in the future among different airhumidifers, so not having two different ones in this class will make it simpler.

miio/airhumidifier_jsq.py
```diff
@@ -43,19 +90,21 @@ class LedBrightness(enum.Enum):
...
 
-    def __init__(self, data: Dict[str, Any]) -> None:
-        """Status of an Air Humidifier (shuii.humidifier.jsq001):
 
-            [24, 30, 1, 1, 0, 2, 0, 0, 0]
+class LedBrightnessJsq002(enum.Enum):
+    Off = 1
+    Low = 2
+    High = 3
```
rytilahti, line 102: Reuse the existing & do the conversion when needed. You can create a separate dict mapping to map between the enum and the integer value to perform conversions based on `self.model`.

miio/airhumidifier_jsq.py
```diff
@@ -43,19 +90,21 @@ class LedBrightness(enum.Enum):
...
 
-        Parsed by AirHumidifierJsq device as:
-            {'temperature': 24, 'humidity': 29, 'mode': 1, 'buzzer': 1,
-             'child_lock': 0, 'led_brightness': 2, 'power': 0, 'no_water': 0,
-              'lid_opened': 0}
-        """
+
+class AirHumidifierStatusJsqCommon(DeviceStatus):
```
rytilahti, line 105: Please extend the existing status class, the values that are not available for a model should be returned as `None`.

rytilahti (commented):
Thanks for the PR @iromeo!

Please extend the existing classes to support this model instead of creating new ones: 1) it will be easier for downstreams to use it when there is a common interface and 2) later on, it will be easier to implement a common interface among different humidifer implementations.

miio/airhumidifier_jsq.py
iromeo: Hi, having some good API will be great. 

At the moment such enums are actually used as `string <-> response_value` mappings that are used for two purposes. First as response value (string or int) to presentable name mapping, the second as a list of available device modes names for UI. E.g. to generate names of available options in:
* python-miio CLI interface
* in projects like https://github.com/syssi/xiaomi_airpurifier

If I reuse the existing enum, we will have to hardcode device mode and brightness levels names in all projects based on python-miio. At the moment python-miio is a single place that captures all info about devices, e.g. their protocols and all options. So it isn't a good idea to force all dependent projects to figure out from the scratch what are valid names for different options for each device supported in the library. Better to do this work once in 'python-miio'. Moreover, even python-miio CLI will return confusing error message if a user passes the wrong mode argument via the command line:

I mean, that code like:
```
    @command(
        click.argument("mode", type=EnumType(OperationModeJsq002)),
        default_output=format_output("Setting mode to '{mode.value}'"),
    )
    def set_mode(self, mode: OperationModeJsq002):
        ...
```
results in message generated by click library (from argument type error), like ~ incorrect value 'foo', expected one of: 'level1', 'level2', 'level3'. So it doesn't call `set_mode`. If using `OperationModeJsq` error will be included mode names that are N/A for `jsq002` and that is confusing for user.

So how do you suggest fixing it correctly? As far as I understand better is do not use enums at all and hardcode the same info into a dictionary. It will be useful for API because we could use same method return types everywhere.

E.g.:
1) Humidifier device could have a `mode presentable name: str <-> response value: object` mapping. And a method that returns list of all supported modes names.  Then dependent projects could get list of supported option names to show them in UI.  And same for brightness modes names. So information about the mapping have to be hardcoded somewhere, the question is in what form.

```
class AirHumidifierStatusJsq002:
    @property
    def mode(self) -> str:
      # return presentable mode name
```
or 

```
class AirHumidifierStatusJsq002:
    @property
    def mode(self) -> obj:
      # return raw value from device
```

and :
```
class AirHumidifierJsq002(..):
   def __init__(..):
      self._mode_values_2_name = OrderDict([(1, "Level1"), .., (3, "Level3")])

   def mode_names() -> Tuple[str]:
      return tuple( self. _mode_values_2_name.keys())
```

2) As for :
```
 @command(
        click.argument("mode", type=EnumType(OperationMode)),
        default_output=format_output("Setting mode to '{mode.value}'"),
    )
    def set_mode(self, mode: OperationMode):
        ...
```
seems is better to use again 'string/int values/object' e.g.:

Using presentable names:
```
 @command(
        click.argument("mode", type=str)
        default_output=format_output("Setting mode to '{mode.value}'"),
    )
    def set_mode(self, mode: str):
       # if `mode` not in `self. mode_names()` than report an error
        ...
```
or use raw values reported by device as `mode`
```
 @command(
        click.argument("mode", type=obj)
        default_output=format_output("Setting mode to '{mode.value}'"),
    )
    def set_mode(self, mode: obj):
       # if `mode` not in `supported device response values` than report an error
        ...
```

---
So what do you think, and especially:
* Should I replace enums with str values?
* Is it better to use for Status `mode` device raw value (int or str) or better report some string presentable names? Maybe for dependent projects string names are better. Because users will use them in their automation, etc. code when comparing device state with some desired state.
* If do not use enums, should I extend `AirHumidifierStatus` and override signatures of `def mode(self) -> OperationMode:` and other properties? Or better to have an independent status class?
* I could modify `*jsq001` API also (I was a contributor). This will break compatibility with `https://github.com/syssi/xiaomi_airpurifier` project, but there I'm working on PR for `*jsq002` support, so I could update integrations with both devices. If we use presentable names for status mode values, user automations shouldn't be broken. `* jsq001` isn't supported by HA, so not changes in HA required. Maybe some other home automation solutions uses current `*jsq001` API, I'm not aware of this.

iromeo: @rytilahti Hi, any suggestion about my above thoughts?

rytilahti: Hey @iromeo, sorry, I haven't had time to formulate a proposal how I think this should be approached, but here a couple of thoughts:
* IMO, the same model/device family (like `shuii.humidifier.jsq*` in this case) should be implemented using a simple class that adapts to the model-specific differences. This means sharing the same data classes (for status) and raising exceptions when unavailable methods are called. 
* Different modes: this library really needs a common interface for the features (or settings) that apply to different devices. A first step could be something like adding `supported_modes` (or whatever naming fits the best) that returns a `dict[str, int]` and `set_mode_by_name(str)`. This would be similar to what `roborock.Vacuum` has with `fanspeed_presets` to inform homeassistant for the available fan speeds.
* For generic functionality, enums (or booleans where applicable) should be preferred, as those can be type-checked statically. Maybe `set_led(bool)` as a generic interface suffices until we find a proper solution to provide different "settings" in a common way? The implementations can keep having their separate, internal implementations, but it'd be great to have a stable interface to avoid imports like this https://github.com/syssi/xiaomi_airpurifier/blob/develop/custom_components/xiaomi_miio_airpurifier/fan.py#L31-L74 .

I'll see if I can find some time to implement a common interface for vacuums based using `typing.Protocol` that could be used as an inspiration for similar interface for humidifers, but I cannot promise currently anything.

tldr; my concern is that as there are currently no strict API contracts, it is awfully complicated to use this lib, e.g., in homeassistant as it is necessary to adapt parts of the code depending on the device and the implementation.

To briefly answer your explicit questions:

> Should I replace enums with str values?

Enums are better (as they can be consolidated far easier in the future) than strings.

> Is it better to use for Status mode device raw value (int or str) or better report some string presentable names? Maybe for dependent projects string names are better. Because users will use them in their automation, etc. code when comparing device state with some desired state.

Raw, device-reported values should not be used as-is as due to the reasons above.

> If do not use enums, should I extend AirHumidifierStatus and override signatures of def mode(self) -> OperationMode: and other properties? Or better to have an independent status class?

As long as the class is exchangeable with the existing one, it's fine. I would rather see the current class extended to cover new devices.

> I could modify *jsq001 API also (I was a contributor). This will break compatibility with https://github.com/syssi/xiaomi_airpurifier project, but there I'm working on PR for *jsq002 support, so I could update integrations with both devices. If we use presentable names for status mode values, user automations shouldn't be broken. * jsq001 isn't supported by HA, so not changes in HA required. Maybe some other home automation solutions uses current *jsq001 API, I'm not aware of this.

I think it's fine to break the compatibility here, having a common API will also make the implementation of `xiaomi_airpurifier` simpler :-)

iromeo: > For generic functionality, enums (or booleans where applicable) should be preferred, as those can be type-checked statically. 
I agree with bools

As for `enums`:
* Using single `technical` enum for a series/all device, where half of the values are N/A in the device `A` and other half ar N/A in the device `B` (and results in runtime exception) doesn't bring much profit in terms of type checking. No profit because correct types will result in run time error when using the unsupported value from enums. So just a set of supported string names (that could be requested in API for each device) doesn't differ form enum in such case.

* Maybe the main profit of enum is for internationalization, e.g. we could consider aggregated `enum` for all states of all `airpurifier` devices just as a list of all known states names, some of them are unique, some of them a reused in multiple devices. For new device `C` a contributor could reuse existing states and add new states if required. So we could implement translation for these states instead of doing it for each device personally, e.g. to avoid cases when `A` uses `"level #1"` and `B` uses `"Level1"` default names instead of `AllStates.LEVEL_1`. So each device should provide list of supported states from the enum and some mapping used for device response parsing from incoming object to one of supported enum states.


>`shuii.humidifier.jsq*`  should be implemented using a simple class that adapts to the model-specific differences. This means sharing the same data classes (for status) and raising exceptions when unavailable methods are called.

To my mind `shuii.humidifier.jsq001` and `shuii.humidifier.jsq002` are two very different devices and the most similar in them is just their device names. Their feature sets somewhat overlaps, but there is a difference on both sides, also there is a difference in communication protocols, command names, response format, state names, state ranges and their ids. Seems we need some generic humidifier/purifier API that could cover a wide set of features, and each device could report what features are supported from the all known features set.

>  fine to break the compatibility here,
Ok, I'll refactor this some way so we could discuss the result )

vkarpunin:
Hi! Could I help somehow to move this PR forward?

PR #1328, opened 2022-02-19
Add common API for constructing device instances, improve introspection

### LINKED ISSUE #1117: [meta] Add device registry & device factory facilities
There is no way to initialize an implementation for the given model, forcing downstreams to hardcode imports & initialization for specific devices. There is also currently no way to obtain a list of supported models, which can be useful for other tasks.

The goal of this issue is two-fold: 1) provide a registry that allows accessing the information about supported devices and 2) provide facilities to allow initializing device-derived implementations for a given host&token combination (with optional `model` to skip the `miIO.info` query). This issue depends on #1116.

Potential signature for the factory method:
```
class Device:
    @classmethod
    def create(cls, host, token, model=None):
        pass
```

rytilahti:
New class methods in Device:
* all_supported_models() to return a dictionary model=>implementation class
* class_for_model(model: str) to return implementation class for the given model
* create(host: str, token: str, model: Optional[str]) to create a device instance

New cli commands:
* `miiocli supported-models` will print a list of supported models

TBD:
- [x] Tests
- [ ] Extract the factory into its own class
- [ ] Update README and docs for changes

Fixes #1117

### DIFF
--- a/miio/airconditioningcompanion.py
+++ b/miio/airconditioningcompanion.py
@@ -14,7 +14,8 @@ MODEL_ACPARTNER_V1 = "lumi.acpartner.v1"
 MODEL_ACPARTNER_V2 = "lumi.acpartner.v2"
 MODEL_ACPARTNER_V3 = "lumi.acpartner.v3"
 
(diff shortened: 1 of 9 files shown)


rytilahti:
Superseded by #1556

PR #1373, opened 2022-03-23
Add Vacuum protocol for enforcing interface among vacuum integrations

rytilahti:
This is a first draft towards enforcing interfaces among vacuum integrations,
aimed to consolidate them to allow easier downstream support no matter implementation details.

Open questions:
- [ ] Is this even a sane approach, wrt. simply creating an abstract base-class to inherit from
- [ ] Status containers need their own interface
- [ ] Find out a better way than importing typing internals for metaclass handling
- [ ] Tests

### DIFF
--- a/miio/interfaces/vacuum.py
+++ b/miio/interfaces/vacuum.py
@@ -1,0 +1,32 @@
+from typing import Dict, Protocol, runtime_checkable
+
+from miio import DeviceStatus
+
+FanspeedPresets = Dict[str, int]
+
+
+@runtime_checkable
+class Vacuum(Protocol):
+    def start(self):
+        """Start vacuuming."""
+
+    def pause(self):
+        """Pause vacuuming."""
+
+    def stop(self):
+        """Stop vacuuming."""
+
+    def home(self):
+        """Return back to base."""
+
+    def find(self):
+        """Request vacuum to play a sound to reveal its location."""
+
+    def status(self) -> DeviceStatus:
+        """Return device status."""
+
+    def set_fan_speed(self, int):
+        """Set fan speed."""
+
+    def fan_speed_presets(self) -> FanspeedPresets:
+        """Return available fan speed presets."""
--- a/miio/device.py
+++ b/miio/device.py
@@ -3,6 +3,7 @@ import logging
 import warnings
 from enum import Enum
 from pprint import pformat as pf
+from typing import _ProtocolMeta  # type: ignore[attr-defined]
 from typing import Any, Dict, List, Optional  # noqa: F401
 
 import click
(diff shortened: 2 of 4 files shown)


miio/interfaces/vacuum.py
```diff
@@ -0,0 +1,32 @@
...
+FanspeedPresets = Dict[str, int]
+
+
+@runtime_checkable
+class Vacuum(Protocol):
+    def start(self):
+        """Start vacuuming."""
+
+    def pause(self):
+        """Pause vacuuming."""
```
2pirko, line 14: Some implementation does not contain pause. It would be good to describe, what should happen, if the method is not supported. Raise an exception?

miio/interfaces/vacuum.py
```diff
@@ -0,0 +1,32 @@
...
+        """Return back to base."""
+
+    def find(self):
+        """Request vacuum to play a sound to reveal its location."""
+
+    def status(self) -> DeviceStatus:
+        """Return device status."""
+
+    def set_fan_speed(self, int):
+        """Set fan speed."""
```
2pirko, line 29: It would be good to describe argument. Is argument in range 0-100 (percent)? Or can the argument be only one of the "presets"? What the method do if the argument does not match?

miio/interfaces/vacuum.py
```diff
@@ -0,0 +1,32 @@
+from typing import Dict, Protocol, runtime_checkable
+
+from miio import DeviceStatus
+
+FanspeedPresets = Dict[str, int]
+
+
+@runtime_checkable
+class Vacuum(Protocol):
```
2pirko, line 9: Consider rename interface to IVacuum. It is useful to understand from the name, whether it is a class or interface.

miio/interfaces/vacuum.py
```diff
@@ -0,0 +1,32 @@
...
+    def start(self):
+        """Start vacuuming."""
```
2pirko, line 11: Generally it is good to specify more details about behavior of the interface methods, format of the arguments and returning value. This helps to keep implementations more consistent.

2pirko (approved):

miio/interfaces/vacuum.py
rytilahti: On roborock, calling pause differs from stop in sense that resuming from a pause will continue using the existing internal map if the cleaning is resumed, in contrast to starting a completely new cleaning round.

miio/interfaces/vacuum.py
rytilahti: So this was just based on the current roborock implementation, but neither viomi nor g1vacuum seem to support percentage-based fan speed setting.

The solution would be to have a `set_fan_speed_preset(x)` that accepts "something" that is returned by the implementations' `fan_speed_presets()`. A feature flag & a separate `set_fan_speed_percentage(int)` could be introduced for devices that support arbitrary fan speeds.

I think this is more clearly visible on some fan/air(de)humidifer implementations where the "levels" or "presets" are more common than continuous values between 0 and 100.

miio/interfaces/vacuum.py
rytilahti: Agreed. I just made this PR to show the idea, but your earlier proposal feels like a better solution than this.

rytilahti (commented):
Thanks for the comments @2pirko, I added a couple of notes how I was thinking about this. In hindsight, I think that your proposal on using abstract base classes is a better approach, at least for the very basic common interface.

rytilahti:
Obsoleted in favor of #1404 - it is better to have some common API that is enforced through the abstract base class.

PR #1384, opened 2022-04-04
Add device_id property to Device class

starkillerOG:
This PR adds new `device_id` property to `Device` class.
The `device_id` is part of the handshake reply and thus available also on devices with unknown tokens (i.e., it can be useful for use cases where the token is extract from the cloud or other external sources).

This is also needed for push server implementation: https://github.com/rytilahti/python-miio/pull/1288

### DIFF
--- a/miio/device.py
+++ b/miio/device.py
@@ -169,7 +169,7 @@ class Device(metaclass=DeviceGroupMeta):
             ) from ex
 
     @property
-    def device_id(self):
+    def device_id(self) -> int:
         """Return device id (did), if available."""
         if not self._protocol._device_id:
             self._info = self.info()
--- a/miio/device.py
+++ b/miio/device.py
@@ -171,7 +171,7 @@ class Device(metaclass=DeviceGroupMeta):
     @property
     def device_id(self):
         """Return device id (did), if available."""
-        if self._protocol._device_id == b"":
+        if not self._protocol._device_id:
             self._info = self.info()
         return int.from_bytes(self._protocol._device_id, byteorder="big")
 
--- a/miio/device.py
+++ b/miio/device.py
@@ -169,6 +169,13 @@ class Device(metaclass=DeviceGroupMeta):
             ) from ex
 
     @property
+    def device_id(self):
+        """Return device id (did), if available."""
+        if self._protocol._device_id == b"":
+            self._info = self.info()
+        return int.from_bytes(self._protocol._device_id, byteorder="big")
+
+    @property
     def raw_id(self) -> int:
         """Return the last used protocol sequence id."""
         return self._protocol.raw_id
--- a/miio/device.py
+++ b/miio/device.py
@@ -171,9 +171,9 @@ class Device(metaclass=DeviceGroupMeta):
     @property
     def device_id(self):
         """Return device id (did), if available."""
-        if self._protocol._device_id == b"":
+        if self._protocol.device_id == b"":
             self._info = self.info()
-        return int.from_bytes(self._protocol._device_id, byteorder="big")
+        return int.from_bytes(self._protocol.device_id, byteorder="big")
 
     @property
     def raw_id(self) -> int:
--- a/miio/device.py
+++ b/miio/device.py
@@ -172,7 +172,7 @@ class Device(metaclass=DeviceGroupMeta):
     def device_id(self) -> int:
         """Return device id (did), if available."""
         if not self._protocol._device_id:
-            self._info = self.info()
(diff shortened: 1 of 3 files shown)


starkillerOG:
@rytilahti could you merge this?

miio/device.py
```diff
@@ -168,6 +168,13 @@ def _fetch_info(self) -> DeviceInfo:
                 "Unable to request miIO.info from the device"
             ) from ex
 
+    @property
+    def device_id(self):
+        """Return device id (did), if available."""
+        if self._protocol.device_id == b"":
+            self._info = self.info()
```
rytilahti, line 175: When can this happen? The device_id is always reported in the handshake responses even when the token is not known, so shouldn't it be always available?

starkillerOG: If you initialize the class but did not communicate with the device (not even once) and already acces the device_id property, it will not yet be set.

miio/miioprotocol.py
```diff
@@ -47,8 +47,8 @@ def __init__(
 
         self._discovered = False
         # these come from the device, but we initialize them here to make mypy happy
+        self.device_id = bytes()
         self._device_ts: datetime = datetime.utcnow()
-        self._device_id = bytes()
```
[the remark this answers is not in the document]
starkillerOG: alright, I will revert this.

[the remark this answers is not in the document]
starkillerOG: done

starkillerOG:
@rytilahti processed the feedback

miio/device.py
starkillerOG: Which is not actually that crazy, since once you initialize the class one of the first things you want to do is register the push callbacks of the device. When creating the request to register the callbacks you already need the device_id, but since registering the callbacks could very well be the first thing you do, there may not have been any communication to the device yet.

starkillerOG:
@rytilahti could you merge?

miio/device.py
```diff
@@ -168,6 +168,13 @@ def _fetch_info(self) -> DeviceInfo:
                 "Unable to request miIO.info from the device"
             ) from ex
 
+    @property
+    def device_id(self):
+        """Return device id (did), if available."""
+        if self._protocol._device_id == b"":
```
rytilahti, line 174:
```suggestion
        if not self._protocol._device_id:
```

miio/device.py
```diff
@@ -168,6 +168,13 @@ def _fetch_info(self) -> DeviceInfo:
...
+            self._info = self.info()
```
rytilahti, line 175: I think this should rather perform only the handshake here, as it will deliver the device id already. Would that be enough?

This would allow an easy way to obtain the device id, even if the token is not known, e.g., for leveraging the cloud data to fetch the token.

miio/device.py
```diff
@@ -168,6 +168,13 @@ def _fetch_info(self) -> DeviceInfo:
                 "Unable to request miIO.info from the device"
             ) from ex
 
+    @property
+    def device_id(self):
```
rytilahti, line 172:
```suggestion
    def device_id(self) -> int:
```
As this is part of the base class, please also add a simple test to `test_device.py` that it functions (and will keep functioning in the future) with some example values.

miio/device.py
starkillerOG: Yes I think indeed only doing the handshake is enough, I will change that

starkillerOG: done

miio/device.py
starkillerOG: I think I have added a working test, could you check if this is what you ment?

rytilahti (commented):
This will cause I/O if the device hasn't been initialize prior trying to access the property, but I think the pros outweigh the cons. If needed, this can be revisited in some later point.

Thanks @starkillerOG, this is good to go as soon as some tests are added to check the both code paths! :+1:

starkillerOG:
@rytilahti I just added a test, could you check if that is what you ment.

miio/tests/test_device.py
```diff
@@ -59,6 +59,18 @@ def test_unavailable_device_info_raises(mocker):
     assert send.call_count == 1
 
 
+def test_device_id_handshake(mocker):
+    """Make sure send_handshake() gets called if did is unknown."""
```
rytilahti, line 63: This looks good to me, please also add a test for the code path where the did is known.

starkillerOG: I just added the second test

rytilahti (approved):

PR #1408, opened 2022-05-02
Mijia vacuum v2

neturmel:


neturmel:
Ups, that was only a mistake

PR #1414, opened 2022-05-09
Update troubleshooting to note discovery issues with roborock.vacuum.a27

golddragon007:
Added new model


rytilahti:
Thanks @golddragon007!

PR #1456, opened 2022-07-14
Improve gateway get_devices_from_dict

starkillerOG:
Improve the get_devices_from_dict function of the gateway:
Now the device_id from direct communication is checked against the device_id received from the cloud.
Besides the already presend mac matching.
The _did is removed from the gateway since it is the same as the device_id which was already presend.

### DIFF
--- a/miio/gateway/gateway.py
+++ b/miio/gateway/gateway.py
@@ -209,6 +209,14 @@ class Gateway(Device):
 
         return self._devices
 
+    def _get_device_by_did(self, device_dict, device_did):
+        """Get a device by its did from a device dict."""
+        for device in device_dict:
+            if device["did"] == device_did:
+                return device
+
+        return None
+
     @command()
     def get_devices_from_dict(self, device_dict):
         """Get SubDevices from a dict containing at least "mac", "did", "parent_id" and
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti could you review and merge?
(this was discovered during my work on the push server, but actually has nothing to do with the push server so I split it out).

miio/gateway/gateway.py
```diff
@@ -222,24 +222,34 @@ def get_devices_from_dict(self, device_dict):
         self._devices = {}
 
         # find the gateway
+        found_gateway = False
         for device in device_dict:
-            if device["mac"] == self.mac:
-                self._did = device["did"]
-                break
+            if device["did"] == str(self.device_id):
```
rytilahti, line 227: Please refactor this into an internal `_get_device_by_did()` method which returns either None or the device object.
This way you can avoid deeper nesting inside this method, and make it much simpler by doing something like this:
```
gateway = self._get_device_by_did(self.device_id)`
if gateway is None:
    # log error
    return self._devices

if gateway["mac"] != self.mac:
    # log error about incorrect mac
    return self._devices

# here the rest of the code
```

miio/gateway/gateway.py
```diff
@@ -222,24 +222,34 @@ def get_devices_from_dict(self, device_dict):
...
+                self.device_id,
                 self.model,
             )
             return self._devices
 
         # find the subdevices belonging to this gateway
         for device in device_dict:
-            if device.get("parent_id") == self._did:
+            if device.get("parent_id") == str(self.device_id):
                 # Match 'model' to get the type_id
```
rytilahti, line 253: Reverse t he check to avoid one level of nesting:
```
    if device.get("parent_id") != str(self.device_id):
        continue

    # Match 'model' ..
```

miio/gateway/gateway.py
starkillerOG: Done

miio/gateway/gateway.py
starkillerOG: done

starkillerOG:
@rytilahti thanks for the review, I have processed your feedback.
Could you have a look again?

starkillerOG:
I think the test failure is unrelated

rytilahti (commented):
LGTM, please add a note to description on why/how this improves it, then this is ready to go.

starkillerOG:
@rytilahti I updated the description of the PR, ready to go.

rytilahti (approved):

PR #1467, opened 2022-07-18
Add dmaker.fan.p33.

dainnilsson:
This adds support for the Xiaomi Smart Standing Fan 2 Pro. I've tested it and everything seems functional. I compared https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:fan:0000A005:dmaker-p33:1 to the already supported https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:fan:0000A005:dmaker-p11:1 and from what I can tell they are functionally identical.

I noticed I get a warning: `Found an unsupported model 'dmaker.fan.p33' for class 'FanMiot'. If this is working for you, please open an issue at https://github.com/rytilahti/python-miio/`, but I'm not sure what constitutes a model being "supported", so I didn't do anything there. From looking at the source code, it doesn't look like any of the `FanMiot` devices are being added to `_supported_models`. Please let me know if I should change anything there!

Thanks for a great easy to use project!


dainnilsson:
Rebased on the latest commit from master.

PR #1477, opened 2022-07-24
fix bright level in set_led_brightness for miot purifiers

borky:
For MIOT air purifiers the led brightness value is 0 for the Bright level and due to wrong check the Bright level could not be set. This PR fixes the check and adds tests to cover this device type.


rytilahti (approved):
Good catch, thanks for the PR @borky! :+1:

PR #1482, opened 2022-08-06
Add smb share feature for Chuangmi Camera

0x5e:
The Xiaomi App have limitation in setting NAS location, it has to be in the local network. It discover the NAS then list them then let me choose. I have a NAS across the VPN and actually it's accessible, but not able to be discovered and set.
This function can help people config NAS outside the local network.

Example:
```shell
miiocli chuangmicamera --ip 192.168.60.11 --token xxxxxxxxxx set_nas_config on "smb://user:pass@192.168.40.2/camera" realtime quarter
```

If the `share` param left empty, It will keep the last config.

Tested on `chuangmi.camera.ipc013`

### DIFF
--- a/miio/chuangmi_camera.py
+++ b/miio/chuangmi_camera.py
@@ -379,7 +379,7 @@ class ChuangmiCamera(Device):
     ):
         """Set NAS configuration."""
 
-        params: dict[str, Any] = {
+        params: Dict[str, Any] = {
             "state": state,
             "sync_interval": sync_interval,
             "video_retention_time": video_retention_time,
--- a/miio/chuangmi_camera.py
+++ b/miio/chuangmi_camera.py
@@ -1,11 +1,10 @@
 """Xiaomi Chuangmi camera (chuangmi.camera.ipc009, ipc013, ipc019, 038a2) support."""
 
 import enum
-import logging
-from typing import Any, Dict
-
 import ipaddress
+import logging
 import socket
+from typing import Any, Dict
 from urllib.parse import urlparse
 
 import click
--- a/miio/chuangmi_camera.py
+++ b/miio/chuangmi_camera.py
@@ -396,7 +396,7 @@ class ChuangmiCamera(Device):
                 "type": 1,
                 "name": share.hostname,
                 "addr": addr,
-                "dir": share.path.lstrip('/'),
+                "dir": share.path.lstrip("/"),
                 "group": "WORKGROUP",
                 "user": share.username,
                 "pass": share.password,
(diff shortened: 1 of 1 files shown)


one-leaf:
Tested on chuangmi.camera.ip029a, working fine.

miio/chuangmi_camera.py
```diff
@@ -375,13 +379,29 @@ def set_nas_config(
...
+                "pass": share.password,
+            }
+        
+        params = {
+            "state": state,
+            "sync_interval": sync_interval,
+            "video_retention_time": video_retention_time,
+        }
+        if share_obj != None:
+            params["share"] = share_obj
```
rytilahti, line 405: It would be better to move the `params` definition above the urlparse check, this way the smb code path just creates the share object and sets it to the already existing params dictionary.

miio/chuangmi_camera.py
```diff
@@ -375,13 +379,29 @@ def set_nas_config(
...
-                "sync_interval": sync_interval,
-                "video_retention_time": video_retention_time,
-            },
-        )
+        share_obj = None
+        share = urlparse(share)
+        if share.scheme == "smb":
+            ip = socket.gethostbyname(share.hostname)
+            reversed_ip = ".".join(reversed(ip.split(".")))
+            addr = int(ipaddress.ip_address(reversed_ip))
```
rytilahti, line 387: Is reversing the address really necessary?

0x5e: Yes, I use MiHome App set the smb config, then get config by miiocli, and I see the address is stored reversed.

miio/chuangmi_camera.py
0x5e: OK

rytilahti (approved):
LGTM, can be merged after the linting checks pass. Thanks @0x5e! 👍

0x5e:
Hi @rytilahti , I updated the code for black and isort checks, still failed in mypy. I'm not familiar with pyhon3 syntax, could you help me solve this error? Thanks.
```
miio/chuangmi_camera.py:394: error: Incompatible types in assignment (expression has type "Dict[str, Any]", target has type "IntEnum")
```

miio/chuangmi_camera.py
```diff
@@ -375,13 +378,27 @@ def set_nas_config(
...
-        return self.send(
-            "nas_set_config",
-            {
-                "state": state,
-                "sync_interval": sync_interval,
-                "video_retention_time": video_retention_time,
-            },
-        )
+
+        params: dict[str, Any] = {
```
rytilahti, line 382:
```suggestion
        params: Dict[str, Any] = {
```

cnlnlhb:
I executed this command on the 029a02 camera, it returned ['OK'] but did not actually write. How can I adapt 029a02?
https://home.miot-spec.com/s/chuangmi.camera.029a02

Issue #1495, opened 2022-08-13
[meta] Add introspectable interfaces


rytilahti:
This is a meta task to keep track on what interfaces needs to be added to allow defining all the necessary information inside python-miio to make interfacing with it from homeassistant (and other downstreams) as straightforward as possible.

This issue serves three purposes:
1. As a single point check list to keep track of the progress.
2. As a call for help as I have access only to very limited set of devices.
3. As something that can be linked to homeassistant contributors submitting PRs to enhance support for individual devices.

If you are reading this, and you have a Xiaomi device that is supported in homeassistant by this library, please consider contributing to this project by adding some simple metadata to the status containers of the devices you own (see "call for help" below for details).

The end goal of this journey is that this library provides enough information directly to avoid hard-coding any device-specific details–available features, actions, sensors, settings etc.–inside the downstream applications like homeassistant.
This would make this library the single source of truth when it comes to the devices supported by it, and relieves downstream maintainers from the painstaking process of adapting their implementation whenever this library gets extended.

## Call for help
If you possess any devices besides the 1st gen Roborock vacuum and a zhimi powerstrip which I have, please consider creating a pull request to add the `@sensor`, `@setting` and `@action` descriptor decorators so that homeassistant (and other potential downstreams) can create entities for your device automatically in the future. https://python-miio.readthedocs.io/en/latest/contributing.html#status-containers contains some documentation on the topic, and the issues linked below in the checklist show an example of how they look like in homeassistant.

The homeassistant fork at https://github.com/rytilahti/home-assistant/tree/xiaomi_miio/feat/entities_from_upstream allows testing your changes. When we get feature parity (or close to it) with the existing homeassistant support, the plan is to create smaller PRs (per platform, most likely) based on the fork to first enable the additional entities, and sunset the hard-coded ones afterward.

When adding the descriptors to the status classes, you can provide the following extra kwargs that are currently used by the fork:
* `device_class` would be great to have besides the name and the unit
* `icon` for defining custom icons when no `device_class` is set
*  `state_class` is not a priority at the moment, the default value should be chosen inside homeassistant and overridden only when needed

If you want to contribute information on several devices, please create a single pull request for each separate integration.
Feel free to ping me on the homeassistant discord or via e-mail if you have any questions!

## Checklist

* [x] Device initialization: #1556
* [x] Sensors: #1488
  * `sensor` and `binary_sensor` platforms on homeassistant
* [x] Switches: #1494
  * `switch` platform on homeassistant
[...]

PR #1496, opened 2022-08-13
Fix supported angles for mapped fans

iMicknl:
In https://github.com/rytilahti/python-miio/pull/1362, supported for P15 and P18 has been added. However, both devices support changing the oscillation angle, but don't have a key present in `SUPPORTED_ANGLES`.

This PR adds the model keys to SUPPORTED_ANGLES. In a follow-up PR, I am happy to see if we can add full blown support for the P15 to Home Assistant. (since currently it won't be automatically mapped from P15 to P11 and I had to figure this out myself by reading GitHub issues).

```
  File "/usr/local/lib/python3.10/site-packages/miio/integrations/fan/dmaker/fan_miot.py", line 321, in set_angle
    if angle not in SUPPORTED_ANGLES[self.model]:
KeyError: 'dmaker.fan.p15'
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -129,7 +129,7 @@ Supported devices
 -  Xiaomi Philips Zhirui Bedroom Smart Lamp
 -  Huayi Huizuo Lamps
 -  Xiaomi Universal IR Remote Controller (Chuangmi IR)
--  Xiaomi Mi Smart Pedestal Fan V2, V3, SA1, ZA1, ZA3, ZA4, ZA5 1C, P5, P9, P10, P11, P33
+-  Xiaomi Mi Smart Pedestal Fan V2, V3, SA1, ZA1, ZA3, ZA4, ZA5 1C, P5, P9, P10, P11, P15, P18, P33
 -  Xiaomi Rosou SS4 Ventilator (leshow.fan.ss4)
 -  Xiaomi Mi Air Humidifier V1, CA1, CA4, CB1, MJJSQ, JSQ, JSQ1, JSQ001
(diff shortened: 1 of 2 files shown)


rytilahti:
How is the oscillation angle displayed in homeassistant? I'm asking as I'm currently working on to move much of the device-specific information back to this library to avoid making changes to home-assistant integration when adding new models or features. This involves defining what is to be exposed to homeassistant directly inside this code base (see https://github.com/rytilahti/python-miio/issues/1495 for a quick overview).

I'll happily merge this PR, as every bit that comes closer to upstream helps. That being said, if you want to contribute further, you could check out that meta issue and see what features available in homeassistant could be defined already inside this lib.

iMicknl:
> How is the oscillation angle displayed in homeassistant? I'm asking as I'm currently working on to move much of the device-specific information back to this library to avoid making changes to home-assistant integration when adding new models or features. This involves defining what is to be exposed to homeassistant directly inside this code base (see #1495 for a quick overview).
> 
> I'll happily merge this PR, as every bit that comes closer to upstream helps. That being said, if you want to contribute further, you could check out that meta issue and see what features available in homeassistant could be defined already inside this lib.

Currently it is exposed as a number entity in Home Assistant. However, when setting a new value the error above will happen since the mapping is not here.

image

I see this as a short-term fix indeed, which is not just Home Assistant related. Everyone that uses a P15 or P18 will benefit from this fix.

Anything I did wrong with the docs, or is CI/CD broken?

rytilahti (approved):
CI is indeed broken for the docs, this is good to go, thanks! :+1:

PR #1508, opened 2022-08-17
Expose sensors, switches, and settings for zhimi.airhumidifier

Kirmas:
-sensor decorator 
-switch decorator 
-setting decorator 

for correct working off this https://github.com/rytilahti/home-assistant/pull/3 or similar changes must be merged

### DIFF
--- a/miio/integrations/humidifier/zhimi/airhumidifier.py
+++ b/miio/integrations/humidifier/zhimi/airhumidifier.py
@@ -131,7 +131,6 @@ class AirHumidifierStatus(DeviceStatus):
     @setting(
         name="Led Brightness",
         icon="mdi:brightness-6",
-        unit="something",
         setter_name="set_led_brightness",
         choices=LedBrightness,
         entity_category="config",
--- a/miio/integrations/humidifier/zhimi/airhumidifier.py
+++ b/miio/integrations/humidifier/zhimi/airhumidifier.py
@@ -143,7 +143,7 @@ class AirHumidifierStatus(DeviceStatus):
 
     @property
     @switch(
-        name="Child lock",
+        name="Child Lock",
         icon="mdi:lock",
         setter_name="set_child_lock",
         device_class="switch",
@@ -208,14 +208,14 @@ class AirHumidifierStatus(DeviceStatus):
 
     @property
     @sensor(
-        "Motor speed",
+        "Motor Speed",
         unit="rpm",
         device_class="measurement",
         icon="mdi:fast-forward",
         entity_category="diagnostic",
     )
     def motor_speed(self) -> Optional[int]:
-        """Current fan speed."""
+        """Current motor speed."""
         if "speed" in self.data and self.data["speed"] is not None:
             return self.data["speed"]
         return None
@@ -232,7 +232,7 @@ class AirHumidifierStatus(DeviceStatus):
 
     @property
     @sensor(
-        "Water level",
+        "Water Level",
         unit="%",
         device_class="measurement",
         icon="mdi:water-check",
@@ -254,7 +254,7 @@ class AirHumidifierStatus(DeviceStatus):
 
     @property
     @sensor(
-        "Water tank",
+        "Water Tank Attached",
         device_class="connectivity",
         icon="mdi:car-coolant-level",
         entity_category="diagnostic",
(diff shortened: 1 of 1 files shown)


miio/integrations/humidifier/zhimi/airhumidifier.py
```diff
@@ -108,23 +110,46 @@ def temperature(self) -> Optional[float]:
...
+    )
     def buzzer(self) -> bool:
         """True if buzzer is turned on."""
         return self.data["buzzer"] == "on"
 
     @property
+    @setting(
+        name="Led Brightness",
+        icon="mdi:brightness-6",
+        unit="something",
```
rytilahti, line 134: Instead of doing this, please change the setting definition to default to `None` (or whatever is the expected unit for these per default).

miio/devicestatus.py
```diff
@@ -2,7 +2,16 @@
 import logging
 import warnings
 from enum import Enum
-from typing import Callable, Dict, Optional, Union, get_args, get_origin, get_type_hints
+from typing import (
```
rytilahti, line 5: Do you mind extracting the devicestatus changes into a separate PR? This will make using `git annotate` nicer in case someone in the future wonders why the API got changed.

miio/integrations/humidifier/zhimi/airhumidifier.py
```diff
@@ -99,6 +100,7 @@ def mode(self) -> OperationMode:
         return OperationMode(self.data["mode"])
 
     @property
+    @sensor("Temperature", unit="°C", device_class="temperature")
```
rytilahti, line 103:
```suggestion
    @sensor("Temperature", unit="C", device_class="temperature")
```
I'm starting having second thoughts on how to define the unit information, and I'm currently wondering if it should be an enum to leave the conversion to string to the downstream users... I have no clear answer for this right now, but we should revisit this at some point later on.

miio/integrations/humidifier/zhimi/airhumidifier.py
```diff
@@ -108,23 +110,45 @@ def temperature(self) -> Optional[float]:
...
+    )
     def led_brightness(self) -> Optional[LedBrightness]:
         """LED brightness if available."""
         if self.data["led_b"] is not None:
             return LedBrightness(self.data["led_b"])
         return None
 
     @property
+    @switch(
+        name="Child lock",
```
rytilahti, line 146:
```suggestion
        name="Child Lock",
```
Let's use title case for the names throughout the file, please correct the other instances too.

miio/integrations/humidifier/zhimi/airhumidifier.py
```diff
@@ -214,17 +252,44 @@ def water_level(self) -> Optional[int]:
 
         return int(min(depth / 1.2, 100))
 
+    @property
+    @sensor(
+        "Water tank",
```
rytilahti, line 257: "Water Tank Attached", maybe?

miio/integrations/humidifier/zhimi/airhumidifier.py
```diff
@@ -183,6 +207,13 @@ def firmware_version_minor(self) -> int:
...
     @property
+    @sensor(
+        "Motor speed",
+        unit="rpm",
+        device_class="measurement",
+        icon="mdi:fast-forward",
+        entity_category="diagnostic",
+    )
     def motor_speed(self) -> Optional[int]:
         """Current fan speed."""
```
rytilahti, line 218: The docstring talks about fan speed, maybe the naming should be consolidated?

miio/devicestatus.py
Kirmas: https://github.com/rytilahti/python-miio/pull/1519

miio/integrations/humidifier/zhimi/airhumidifier.py
Kirmas: home-assistant use °C for TEMP_CELSIUS. https://github.com/home-assistant/core/blob/a641bbc352b7559c9491971fa6fc06437f4ab964/homeassistant/const.py#L507 

Will we have some step to convert from C to TEMP_CELSIUS?

rytilahti: I think we may need to create some sort of mapping. I'm wondering if there are only so few units in the end that we could simply have an enum inside python-miio to expose those, otherwise there are [libraries like pint](https://pint.readthedocs.io/en/stable/) that could allow doing this in a more standardized manner.

In any case, we don't want to directly tie this library to the internal presentations as there are no guarantees that their API (which is an internal one nonetheless) will remain stable in the future.

PR #1519, opened 2022-09-05
Make unit optional for @setting, fix type hint for choices

Kirmas:
unit is optional string
fix choices type


rytilahti (approved):
Thanks! :+1:

PR #1528, opened 2022-09-15
Roborock timer: add find_next_schedule to replace next_schedule

phil9909:
This is a follow up for: https://github.com/rytilahti/python-miio/pull/1520#discussion_r970052935

In PR #1520 we discussed, if `next_schedule` should return the next schedule relative to (A) the time of access or (B) the time of object creation. While alternative B has the advantage of "fresh" data, even if a reference to a Timer is kept for an extended period, it may be surprising to a user that a property of an object can change over time (without any external influence). In other words: I would be surprised, if in a single threaded program the following assertion could ever fail `assert timer.next_schedule == timer.next_schedule`.
If an API behaves in a surprising way, it may lead to bugs as developers assume a certain behavior while using the API (see [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)).

As an alternative I suggested deprecating the `next_schedule` property and replace it with a method `find_next_schedule`. A method returning a different value based on the current time is far less surprising (IMHO).

To be honest: I'm not sure if it's worth the change. Maybe programs just should not keep the Timer objects in memory anyhow, as they might get changed by the user or another program at any point in time.

### DIFF
--- a/miio/integrations/vacuum/roborock/vacuumcontainers.py
+++ b/miio/integrations/vacuum/roborock/vacuumcontainers.py
@@ -7,7 +7,7 @@ from pytz import BaseTzInfo
 
 from miio.device import DeviceStatus
 from miio.devicestatus import sensor, setting
-from miio.utils import pretty_seconds, pretty_time
+from miio.utils import deprecated, pretty_seconds, pretty_time
 
 
 def pretty_area(x: float) -> float:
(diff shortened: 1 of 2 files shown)


rytilahti:
So, I don't think downstream users should cache the timer objects due to the reasons you already mentioned. The saved I/O is unlikely worth it and the use cases for it are rather uncommon, and if someone really needs that it is not so difficult to implement this on your own.

Issue #1385, opened 2022-04-05
Mi Robot Vacuum-Mop 2 Pro (ijai.vacuum.v3)


evenywchen:
**Device information:**

  - Name(s) of the device: Mi Robot Vacuum-Mop 2 Pro | MJST1SHW
  - Link: https://home.miot-spec.com/s/ijai.vacuum.v3
  - Model: ijai.vacuum.v3
  - Hardware version: 
  - Firmware version: 4.1.5_0041

jeffrey-zammit:
Was there ever any update to this?

rytilahti:
#1497 adds support for this, but I haven't had time to review it yet. Anyone with a device could contribute by testing that PR (and/or doing code review on it).

jeffrey-zammit:
Can you advise how I can test this out?

redQs:
how do i get this PR? i got 1 and can test it

rytilahti:
I think the simplest way is to use the [github client](https://cli.github.com/) to check out the PR:
```
gh repo clone rytilahti/python-miio
cd python-miio
gh pr checkout 1497
poetry install
poetry run miiocli pro2vacuum
```

rytilahti:
Sorry, that is not currently possible. Adding support for non-roborock vacuums requires some very heavy refactoring which will take some time. #1495 is the issue where the refactoring work is being tracked.

NEoKhajitt:
Hi @rytilahti, i have a `ijai.vacuum.v3` and would like to contribute by testing it.
Followed the steps you mentioned above (cloning repo, checkout PR) had to setup poetry etc.

I am running `poetry 1.2.0` with `Python 3.8.0` i tried your command above `poetry run miiocli pro2vacuum` but get an error message `    from .pcapparser import parse_pcap
  File "C:\Users\nielm\Downloads\python-miio\miio\devtools\pcapparser.py", line 10, in <module>
    def read_payloads_from_file(file, tokens: list[str]):
TypeError: 'type' object is not subscriptable`

I also ran it with the ip and token args `poetry -v run miiocli pro2vacuum --ip 192.168.1.70 --token MYTOKEN`
but i get the exact same error message.
Is there something else i'm missing?

rytilahti:
@NEoKhajitt oops, looks like I accidentally broke python 3.8 support when adding pcapparser... You could modify that exact line locally to have `List[str]` (uppercase L) instead of `list[str]` (lowercase l) to fix that for testing.

[...]

PR #1530, opened 2022-09-18
Use typing.List for devtools/pcapparser

rytilahti:
As pcapparser is now shipped and exposed through the cli, and we still support python 3.8, we need to use the generics from `typing`.

Thanks to @NEoKhajitt for reporting this in #1385!


NEoKhajitt:
CI / Perform linting checks (3.10) (pull_request)  is failing due to `Could not parse version constraint: ` there is a problem in the `poetry 1.2.1` package when testing local I had the same issue and you have to downgrade to `poetry 1.2.0` on Windows env.

rytilahti:
The CI error is probably described in https://github.com/python-poetry/poetry/issues/6519 but I wasn't able to confirm that the linked PR works for it. As I don't like the idea of pinning the build system version, let's see if it gets fixed when a new poetry-core with that PR is released.

PR #1531, opened 2022-09-19
Make pushserver more generic

rytilahti:
This generalizes the push server to allow using it for other uses besides event handling, e.g., to create device simulators.
When receiving a datagram from an unregistered device, the implementation will now try to decrypt the contents using a token full of 0s. If the decryption succeeds, the list of previously registered methods (using `add_method(name: str, response: Union[Dict, Callable)`) is used to obtain a response to be send back to the requester. An error is send to the client if no method with the given name is found.

The main motivation behind this change is to make it easier for developers to test miiocli & the library against devices they do not possess by providing easy-to-use device simulators for both miio and miot protocols. The simulator implementations will follow in separate PRs.

@starkillerOG could you please take a look at this PR to make sure I'm not breaking anything event related? I'll add some tests prior getting this merged so that we don't break it accidentally in the future.

TBD:
- [x] Add tests

### DIFF
--- a/miio/push_server/server.py
+++ b/miio/push_server/server.py
@@ -3,7 +3,7 @@ import logging
 import socket
 from json import dumps
 from random import randint
-from typing import Callable, Optional
+from typing import Callable, Dict, Optional, Union
 
 from ..device import Device
 from ..protocol import Utils
@@ -53,7 +53,7 @@ class PushServer:
         await push_server.stop()
     """
 
-    def __init__(self, device_ip):
+    def __init__(self, device_ip=None):
         """Initialize the class."""
         self._device_ip = device_ip
 
@@ -66,6 +66,8 @@ class PushServer:
         self._listen_couroutine = None
         self._registered_devices = {}
 
+        self._methods = {}
+
         self._event_id = 1000000
 
     async def start(self):
@@ -76,7 +78,9 @@ class PushServer:
 
         self._loop = asyncio.get_event_loop()
 
-        _, self._listen_couroutine = await self._create_udp_server()
+        transport, self._listen_couroutine = await self._create_udp_server()
+
+        return transport, self._listen_couroutine
 
     async def stop(self):
         """Stop Miio push server."""
@@ -89,6 +93,13 @@ class PushServer:
         self._listen_couroutine.close()
         self._listen_couroutine = None
         self._loop = None
+
+    def add_method(self, name: str, response: Union[Dict, Callable]):
+        """Add a method to server.
+
+        The response can be either a callable or a dictionary to send back as response.
+        """
+        self._methods[name] = response
 
     def register_miio_device(self, device: Device, callback: PushServerCallback):
         """Register a miio device to this push server."""
@@ -208,7 +219,8 @@ class PushServer:
 
     async def _create_udp_server(self):
         """Create the UDP socket and protocol."""
-        self._server_ip = await self._get_server_ip()
+        if self._device_ip is not None:
+            self._server_ip = await self._get_server_ip()
 
         # Create a fresh socket that will be used for the push server
         udp_socket = socket.socket(family=socket.AF_INET, type=socket.SOCK_DGRAM)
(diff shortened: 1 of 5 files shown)


miio/push_server/server.py
```diff
@@ -90,6 +94,13 @@ async def stop(self):
...
         self._loop = None
 
+    def add_method(self, name: str, response: Union[Dict, Callable]):
+        """Add a method to server.
+
+        The response can be either a callable or a dictionary to send back as response.
+        """
+        self._methods[name] = response
+
     def register_miio_device(self, device: Device, callback: PushServerCallback):
```
starkillerOG, line 104: I would add a check here:
```
if self._device_ip is None:
   warning
   return
```

To make sure this is a proper server that can work with real devices

miio/push_server/serverprotocol.py
```diff
@@ -52,57 +52,101 @@ def send_ping_ACK(self, host, port):
...
         response = Message.build(msg, token=token)
-        self.transport.sendto(response, (host, port))
+
+        return response
+
+    def send_response(self, host, port, msg_id, token, payload=None):
+        if payload is None:
+            payload = {}
+
+        result = {**payload, "id": msg_id}
```
starkillerOG, line 76: I would call this data instead of result

miio/push_server/serverprotocol.py
```diff
@@ -52,57 +52,101 @@ def send_ping_ACK(self, host, port):
...
+        """Handle received messages."""
+        try:
+            (host, port) = addr
+            if data == HELO_BYTES:
+                return self.send_ping_ACK(host, port)
 
-            callback(source_device_id, action, msg_value.get("params"))
+            if host in self.server._registered_devices:
+                return self._handle_datagram_from_registered_device(host, port, data)
+            else:
```
starkillerOG, line 148: I am slightly worried that we might exedently send incorrect messages to a real device if that device is for instance not properly disconnected and the server is restarted or other edge cases.
Maybe we schould make a `register_client` function and check against that instead of just handeling everything which is not registerd as a miio_device as a client.

miio/push_server/serverprotocol.py
```diff
@@ -52,57 +52,101 @@ def send_ping_ACK(self, host, port):
...
+        """Handle requests from registered eventing devices."""
+        token = self.server._registered_devices[host]["token"]
+        callback = self.server._registered_devices[host]["callback"]
+
+        msg = Message.parse(data, token=token)
+        msg_value = msg.data.value
+        msg_id = msg_value["id"]
+        _LOGGER.debug("<< %s:%s: %s", host, port, msg_value)
 
-            if host not in self.server._registered_devices:
```
starkillerOG, line 82: I would like to have this error back when the host is not registered as a miio_device or a client.
See comment below.

starkillerOG (commented):
Looks good to me.
Sorry for the late review, I was quite bussy.
I really like we now have some tests for the serverprotocol, thank you very much!

I have some small feedback, maybe you could adress that in a follow up PR?

I will try to test this new code with my devices tonight and let you know!

starkillerOG:
@rytilahti I just tested this code with gateway devices, and the event pushes still work just fine.
Would you be so kind to have a look at the feedback I left here?
Sorry for beeing late.

miio/push_server/serverprotocol.py
rytilahti: Done :+1:

miio/push_server/server.py
rytilahti: I'm not sure I'm following what you mean, the `device_ip` is checked right below?

miio/push_server/serverprotocol.py
rytilahti: This is now checked inside `datagram_received`, for registered devices. But you are right, we are responding to all non-registered devices as if they were real clients..

miio/push_server/serverprotocol.py
rytilahti: The incorrect messages would be encrypted with a hardcoded token that should be different from a real device, so sending a message to the server with some other will simply cause the device not to respond. That is why I don't think it's a non-issue.

rytilahti:
@starkillerOG thanks for taking a look, and sorry for taking so long to respond... I changed the variable naming, but I'm not sure if we should do the other changes, see my commentary inline.

PR #1538, opened 2022-09-26
Allow custom methods for miio simulator

rytilahti:
This will allow simulating devices that do use other methods besides "get_prop" for fetching the properties.

The result can be defined either as regular yaml object (`result`) or as json string (`result_json`).

Added simulated roborock s7 as an example case, looks like the status container embedding is working as expected (@starkillerOG):
image


starkillerOG:
@rytilahti I now also tested the embeded status container with a real vacuum, it seems to work, but there is some issue with the getatrributes see https://github.com/rytilahti/python-miio/pull/1542

PR #1542, opened 2022-09-30
Make embedded statuses directly accessible attributes

starkillerOG:
Consider this test script with a Roborock vacuum:
```
from miio import RoborockVacuum
r = RoborockVacuum(ip = "192.168.1.IP", token = "TokenTokenToken")
status = r.status()
```

`print(status["CleaningSummary:count"])`
Will give
`TypeError: 'VacuumStatus' object is not subscriptable`

`print(status.CleaningSummary:count)`
will give
`SyntaxError: invalid syntax`
This is a invalid syntax since variable names can only contain (A-Z, a-z), (0-9), and (_), see https://realpython.com/python-variables/#variable-names

~Therefore overwriding __getattribute__ will not work as intended.~
Alright so it does work when using getattr(status, "CleaningSummary:count") which is used inside HomeAssistant for conveniance. So lets keep it.

This PR will make the embeded classes a attribute of the main class, so you can acess the embedded classes using:

`print(status.CleaningSummary.count)`
Which will work

(This is still needed since using normal python scripts there needs to be an easy way to acces the embeded containers)

### DIFF
--- a/miio/devicestatus.py
+++ b/miio/devicestatus.py
@@ -119,6 +119,7 @@ class DeviceStatus(metaclass=_StatusMeta):
         other_name = str(other.__class__.__name__)
 
         self._embedded[other_name] = other
+        setattr(self, other_name, other)
 
         for name, sensor in other.sensors().items():
             final_name = f"{other_name}:{name}"
@@ -133,14 +134,6 @@ class DeviceStatus(metaclass=_StatusMeta):
         for name, setting in other.settings().items():
             final_name = f"{other_name}:{name}"
             self._settings[final_name] = attr.evolve(setting, property=final_name)
-
-    def __getattribute__(self, item):
-        """Overridden to lookup properties from embedded containers."""
-        if ":" not in item:
-            return super().__getattribute__(item)
-
-        embed, prop = item.split(":")
-        return getattr(self._embedded[embed], prop)
 
 
 def sensor(name: str, *, unit: str = "", **kwargs):
--- a/miio/devicestatus.py
+++ b/miio/devicestatus.py
@@ -135,6 +135,13 @@ class DeviceStatus(metaclass=_StatusMeta):
             final_name = f"{other_name}:{name}"
             self._settings[final_name] = attr.evolve(setting, property=final_name)
 
+    def __getattribute__(self, item):
+        """Overridden to lookup properties from embedded containers."""
+        if ":" not in item:
+            return super().__getattribute__(item)
+
+        embed, prop = item.split(":")
+        return getattr(self._embedded[embed], prop)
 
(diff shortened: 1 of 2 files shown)


starkillerOG:
@rytilahti I don't know why mypy is complaining about the ._sensors etc.
But in any case that is unrelated to this PR.

starkillerOG:
@rytilahti could you have a look at this PR, it is very small and simple.

rytilahti:
Yeah, I was thinking about the naming and I knew it would not be a valid identifier for direct accesses. So, I'm wondering if we should just use double-underscore instead of `:` as a separator, this would also allow accesses without going through the subcontainer name.

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -403,11 +403,16 @@ def manual_control(
...
-        status = VacuumStatus(self.send("get_status")[0])
+        status = self.vacuum_status()
         status.embed(self.consumable_status())
         status.embed(self.clean_history())
         return status
 
+    @command()
+    def vacuum_status(self) -> VacuumStatus:
+        """Return status of the vacuum."""
+        return VacuumStatus(self.send("get_status")[0])
```
rytilahti, line 414: This change is separate and should be in your other PR, shouldn't it?

miio/devicestatus.py
```diff
@@ -119,6 +119,7 @@ def embed(self, other: "DeviceStatus"):
         other_name = str(other.__class__.__name__)
 
         self._embedded[other_name] = other
+        setattr(self, other_name, other)
```
rytilahti, line 122: We can add this (and update the docstring to inform anyone reading the docs about the fact), and also change the separator to `__` instead of `:`, in this PR?

starkillerOG: No `__` can not be used as seperator, I tried it but very bad things start happening.
Python uses __ for python variables like __main__, __name__ and even worse __getattribute__ itself is also an attribute to the class, so you get into an infinite loop of __getattribute__ getting itself using another __getattribute__ and you crash the program.

miio/integrations/vacuum/roborock/vacuum.py
starkillerOG: I will move it

starkillerOG:
@rytilahti ready for another review/merge

miio/devicestatus.py
rytilahti: I don't think having a double undescore should be a problem as a separator in the middle of the name, as far as I know the dunder methods need to both start and end with `__`.

starkillerOG:
Closing this as it has been resolved by https://github.com/rytilahti/python-miio/pull/1573

PR #1549, opened 2022-10-11
Update README.rst

sahilunagar00:
How do I install Python-Miio?
How does The Miio module work?


rytilahti:
Copy&paste from the existing docs, looks like hacktoberfest spamming so I'll just close this.

PR #1557, opened 2022-10-24
Add descriptors for yeelight

rytilahti:
Adds descriptor decorators for the yeelight integration, so that they can be automatically detected by downstreams.
This also adds a minimal LightInterface, and converts Yeelight to use it.
This interface is expected to change, but this will enable development for the time being.

Related to #1495 - ping @Kirmas


Kirmas:
Hello @rytilahti 
Code part looks good for me. I've been sick lately and I'm still not feeling well. As soon as there is a little more energy, I will immediately check this code with my lamps. Thank you.

PS. Thanks that merged my PR( https://github.com/rytilahti/python-miio/pull/1508 ) it was unexpected :)

PR #1573, opened 2022-11-06
Use __ as delimiter for embedded statuses

rytilahti:
Leaves ":" free for other uses,
converts to use `__getattr__` instead of `__getattribute__` to avoid checking on every attribute access.


rytilahti:
Ping @starkillerOG – looks like the previous implementation used wrong dunder method for accessing the custom attributes, this should fix the issues you had when you tried to use `__` earlier in names in your PR.

starkillerOG:
Thanks for looking into this, I will test this as soon as possible

starkillerOG:
@rytilahti I just tested it and it works as advertised, Thanks!

PR #1578, opened 2022-11-07
Add interface to obtain miot schemas

rytilahti:
Adds classes to download and parse miot schemas.
Also, converts the miot simulator to allow specifying the model in place of a schema file.

TBD:
- [ ] Externalize cloud requests into a separate lib


rytilahti:
Freezegun is oddly suddenly misbehaving, merging this though to get a PR created for genericmiot.

PR #1579, opened 2022-11-07
Less verbose reprs for descriptors

rytilahti:
Makes reading logs easier.

### DIFF
--- a/miio/descriptors.py
+++ b/miio/descriptors.py
@@ -22,9 +22,9 @@ class ActionDescriptor:
 
     id: str
     name: str
-    method_name: Optional[str] = None
-    method: Optional[Callable] = None
-    extras: Dict = attr.ib(factory=dict)
+    method_name: Optional[str] = attr.ib(default=None, repr=False)
+    method: Optional[Callable] = attr.ib(default=None, repr=False)
+    extras: Dict = attr.ib(factory=dict, repr=False)
 
 
 @attr.s(auto_attribs=True)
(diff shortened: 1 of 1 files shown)


rytilahti:
For some reason `timedelta(hours=1)` on `datetime.now()` does not change the date properly, so the tests will fail if run between 12am and 1am...

```
self = <DummyViomiDishwasher: 192.0.2.1 (token: ffffffffffffffffffffffffffffffff)>, time = FakeDatetime(2022, 11, 8, 0, 57), program = <Program.Quick: 2>

    @command(
        click.argument("time", type=click.DateTime(formats=["%H:%M"])),
        click.argument("program", type=EnumType(Program)),
    )
    def schedule(self, time: datetime, program: Program) -> str:
        """Schedule a program run.
    
        *time* defines the time when the program should finish.
        """
    
        if program == Program.Unknown:
            ValueError(f"Program {program.name} is not valid for this function.")
    
        now = datetime.now()
        scheduled_finish_date = now.replace(
            hour=time.hour, minute=time.minute, second=0, microsecond=0
        )
        scheduled_start_date = scheduled_finish_date - timedelta(
            seconds=program.run_time
        )
        if scheduled_start_date < now:
>           raise ValueError(
                f"Proposed time ({scheduled_start_date} < {now}) is in the past (the proposed time is the finishing time, not the start time)."
            )
E           ValueError: Proposed time (2022-11-07 00:29:05 < 2022-11-07 23:57:53.324466) is in the past (the proposed time is the finishing time, not the start time).

miio/integrations/viomidishwasher/viomidishwasher.py:336: ValueError
```

PR #1585, opened 2022-11-09
Add additional sensors and settings to Roborock vacuums

starkillerOG:

Add additional sensors and improve the already implemented sensors for the Roborock S7 MaxV.

Split vacuum enums to a separte file to prevent circular imports

Controls added:
- Mop routing (enum)
- Mop scrub intensity (enum)
- Auto dust collection (switch)


miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -163,33 +184,76 @@ def error(self) -> str:
...
     @property
-    @setting(
-        "Fanspeed",
-        unit="%",
-        setter_name="set_fan_speed",
-        min_value=0,
-        max_value=100,
-        step=1,
-        icon="mdi:fan",
-    )
```
starkillerOG, line 180: I removed this because it does not work with my Roborock S7 MaxV+.
That only excepts fan_speed values of 101 to 108 (the same settings as in the app), not 0 to 100.
Those fan speeds are already exposed through the vacuum entity in HomeAssistant.

We could add this back for other models that might support it, but we schould then make some model decorator or something to specifiy which models support this.
I would want to leave that to a future PR.

starkillerOG:
@rytilahti can you have a look an merge this?
The test failure is just some unrelated error in the server.

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -964,7 +879,7 @@ def set_mop_mode(self, mop_mode: MopMode):
     @command()
     def mop_intensity(self) -> MopIntensity:
         """Get mop scrub intensity setting."""
-        if self.model != ROCKROBO_S7:
+        if self.model not in [ROCKROBO_S7, ROCKROBO_S7_MAXV]:
             raise UnsupportedFeatureException(
                 "Mop scrub intensity not supported by %s", self.model
```
rytilahti, line 884: An idea: we could remove the model check and embed this also to the returned status. This can be done later in a separate PR though, that should also convert the `embed()` to take a partial to execute instead of a prepared status container. The reasoning is that we can then make error handling inside the embed (and potentially skip unsupported methods from future requests).

miio/integrations/vacuum/roborock/vacuumcontainers.py
rytilahti: I think we should probably keep this but return `None` if the value > 100, assuming this could be used as a proxy to check if the device supports defining the fan speed in percentage?

rytilahti (commented):
A couple of comments, but all in all this looks good :+1:

miio/integrations/vacuum/roborock/vacuumcontainers.py
starkillerOG: Sounds good, I will do that.

starkillerOG: Done

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -180,16 +226,46 @@ def battery(self) -> int:
     )
     def fanspeed(self) -> int:
         """Current fan speed."""
-        return int(self.data["fan_power"])
+        fan_power = int(self.data["fan_power"])
+        if fan_power > 100:
```
rytilahti, line 230:
```suggestion
        # values 100+ are reserved for presets
        if fan_power > 100:
```

miio/integrations/vacuum/roborock/vacuum.py
starkillerOG: It is already in the returned status (water_box_mode is included in the get_status command).
(So this seperate method is a bit redundant).

But you are totally right that there schould be a general way to handle commands that are only supported by a subset of the models.
But lets do that in a future PR.

miio/integrations/vacuum/roborock/vacuumcontainers.py
starkillerOG: done

miio/integrations/vacuum/roborock/vacuum.py
rytilahti: Ah, sorry, I didn't notice that. Sounds good to me! :+1:

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -163,7 +184,32 @@ def error(self) -> str:
...
-    @sensor("Battery", unit="%", device_class="battery")
+    @sensor(
+        "Dock error code",
+        icon="mdi:alert",
+        entity_category="diagnostic",
+        enabled_default=False,
+    )
+    def dock_error_code(self) -> int:
+        """Dock error status as returned by the device."""
+        return int(self.data["dock_error_status"])
```
rytilahti, line 195: This causes an exception as the key is not available on my v1 vacuum.

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -178,18 +224,49 @@ def battery(self) -> int:
...
-    @sensor("Clean Duration", unit="s", icon="mdi:timer-sand")
+    @setting(
+        "Mop scrub intensity",
+        choices=MopIntensity,
+        setter_name="set_mop_intensity",
+        icon="mdi:checkbox-multiple-blank-circle-outline",
+    )
+    def mop_intensity(self) -> int:
+        """Current mop intensity."""
+        return int(self.data["water_box_mode"])
```
rytilahti, line 244: This needs to be changed to be optional, too.

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -178,18 +224,49 @@ def battery(self) -> int:
...
+    @property
+    @setting(
+        "Mop route",
+        choices=MopMode,
+        setter_name="set_mop_mode",
+        icon="mdi:swap-horizontal-variant",
+    )
+    def mop_route(self) -> int:
+        """Current mop route."""
+        return int(self.data["mop_mode"])
```
rytilahti, line 255: See my commentary above. All new properties that may not exist on other vacuums (at least dust collection, mopping, .. are not available on mine) need to be guarded and made optional, so please check all of them as I won't go now through all of them while testing whether they are supported on my v1. 

When you are ready, I will retry it against my v1 :-)

starkillerOG: @rytilahti could you post the raw output of send("get_status") for the v1 so I can check against that?
Or is that simply the first one in the discription:
```
# {'result': [{'state': 8, 'dnd_enabled': 1, 'clean_time': 0,
#  'msg_ver': 4, 'map_present': 1, 'error_code': 0, 'in_cleaning': 0,
#  'clean_area': 0, 'battery': 100, 'fan_power': 20, 'msg_seq': 320}],
#  'id': 1}
```

rytilahti: Here's a recent one:
```
 'result': [{'battery': 100,
             'clean_area': 19087500,
             'clean_time': 1222,
             'dnd_enabled': 1,
             'error_code': 0,
             'fan_power': 60,
             'in_cleaning': 0,
             'map_present': 1,
             'msg_seq': 552,
             'msg_ver': 8,
             'state': 8}]}
```
Here's the container (incl. KeyErrors):
```
<VacuumStatus auto_dust_collection=None battery=100 clean_area=19.0875 clean_time=0:20:22 dock_error=Definition missing for dock error None dock_error_code=None error=No error error_code=0 fanspeed=60 got_error=False in_segment_cleaning=False in_zone_cleaning=False is_on=False is_paused=False is_water_box_attached=None is_water_box_carriage_attached=None is_water_shortage=None map=True mop_intensity=None mop_route=KeyError state=Charging state_code=8 ConsumableStatus=<ConsumableStatus dustbin_auto_empty_used=None filter=4:21:58 filter_left=6 days, 1:38:02 main_brush=9 days, 8:23:56 main_brush_left=3 days, 3:36:04 sensor_dirty=4:21:58 sensor_dirty_left=1 day, 1:38:02 side_brush=5:01:55 side_brush_left=8 days, 2:58:05> CleaningSummary=<CleaningSummary count=674 dust_collection_count=None ids=[1667940243, 1667594218, 1667409403, 1667232470, 1667065794, 1666800328, 1666642567, 1666030897, 1660593167, 1660158015, 1659987490, 1659640516, 1659465433, 1659359824, 1659123168, 1658941529, 1658775771, 1658602278, 1658434044, 1658161285] total_area=11630.995 total_duration=9 days, 7:52:39> DNDStatus=<DNDStatus enabled=True end=08:00:00 start=22:00:00>>
```

starkillerOG: Thanks!
I pushed a commit and I think I now check for all KeyErrors for the newly added sensors/settings.

starkillerOG: Could you retry?

miio/integrations/vacuum/roborock/vacuumcontainers.py
starkillerOG: done

miio/integrations/vacuum/roborock/vacuumcontainers.py
starkillerOG: done

starkillerOG:
@rytilahti could you test again/merge?
(test failures are unrelated)

miio/integrations/vacuum/roborock/vacuumcontainers.py
rytilahti: Seems to work now (after making adjusting the homeassistant fork to use entitycategory properly)! :1st_place_medal: 
image

rytilahti: I'm wondering if I'm misunderstanding something, but shouldn't the entity category be always "config" for settings that can be configured and "diagnostic" for read-onlys? If yes, we don't need to hardcode that in each and every setting but let the homeassistant integration decide on it, see my commit: https://github.com/rytilahti/home-assistant/commit/1365ed79bce065930b94bda910dda89d60b8abcf ?

starkillerOG: So there are 4 categories (I think):
- Controls (settable key features)
- Sensors (non-settable key features)
- Configuraration (settable non essential feature)
- Diagnostics (non-settable non essential feature)

starkillerOG: https://github.com/home-assistant/core/blob/0941ed076ccafe36e2e8ffbdc480cea4f8fa5650/homeassistant/helpers/entity.py#L180

> An entity with a category will:
>     - Not be exposed to cloud, Alexa, or Google Assistant components
>     - Not be included in indirect service calls to devices or areas

starkillerOG: > Getting this exception though, but I'm not sure how it should be handled:
> 
> ```
> ValueError: Invalid datetime: sensor.rockrobo_vacuum_v1_do_not_disturb_start provides state '22:00:00', which is missing timezone information
> ```

Yes I had that to, It is a problem in the miio HomeAssistant component, not this library.
I have a fix for it on my local repo for testing.
(But I don't think I can make a PR to your forked HomeAssistant branch or can I?)

rytilahti: Yeah, okay, so let's remove the category extras completely from python-miio and take care of selecting the correct category inside home assistant. 

We need to have a separate standardized set of platform specific main controls, to get them synchronized with the genericmiot that will be used for this. This will likely be based on the id, e.g., vacuum:power for vacuums etc. but I'm not that far yet.

starkillerOG: Maybe we schould not add entities that have None values (such as the scrub intensity) because they are not supported.
But that is in the HomeAssistant branch code, not this library.

rytilahti: > I have a fix for it on my local repo for testing.
> (But I don't think I can make a PR to your forked HomeAssistant branch or can I?)

It is just a regular branch on a regular git repository (that just happens to be a fork from another repository), so there should be no issues creating a PR on it. Feel free to try it out and if it doesn't work, we'll figure a solution :-)

> Maybe we schould not add entities that have None values (such as the scrub intensity) because they are not supported.

That's a problem we need to tackle separately, I'm afraid. The problem is, that at least on viomi.v8 the availability depends on whether the water or the dust container is attached to the device, so we cannot trust that these remain `None` forever. Maybe the homeassistant integration needs to be be adapted to create new sensors during regular update cycles (if that's allowed) instead of just creating them during the config flow.

starkillerOG:
@rytilahti I think this can be merged right?

miio/integrations/vacuum/roborock/vacuumcontainers.py
starkillerOG: Please leave the entity_category untill we have found a better way to handle it.
We cannot set a default since for instance the Battery schould not have a entity category (schould not be diagnostic but a sensor).

rytilahti:
How about the concern regarding removing the entity_category for the time being? We can discuss that later after we have a better view of how that's going to work with the genericmiot

starkillerOG:
I made a PR to your HomeAssistant fork here: https://github.com/rytilahti/home-assistant/pull/4
That has the fixes I made to prevent the 
`ValueError: Invalid datetime: sensor.rockrobo_vacuum_v1_do_not_disturb_start provides state '22:00:00', which is missing timezone information`
And some other errors

starkillerOG:
> How about the concern regarding removing the entity_category for the time being? We can discuss that later after we have a better view of how that's going to work with the genericmiot

We can not just set a default since for instance the Battery schould not have a entity category (schould not be diagnostic but a sensor), so we schould somehow properly implement this.

I would rather keep it in this PR, because I now figured out which ones schould have which catogory, otherwise I need to figure it out again later.
In any case something schould be done about it in a future PR, but that can then also change/remove them from the vacuum implementation.

starkillerOG:
@rytilahti could you agree to that?
(If you really insist I will remove the entity category).

starkillerOG:
@rytilahti I updated the PR description, can you merge?

PR #1586, opened 2022-11-09
Update decorators

starkillerOG:
- set the id to the qualified_name instead of only the function name (in order to prevent unique_id collisions in HomeAssistant for example with DNDStatus.start and CleaningDetails.start)
- fix sensor unit beeing set to "" instead of None (HomeAssistant default)
- remove setter input from setting decorator, since it will not work to put in a function directly in the decorator since the class object has not yet been created. Therefore always the setter_name schould be used and the binding of the setter methods schould take place when the class object gets created.

### DIFF
--- a/miio/devicestatus.py
+++ b/miio/devicestatus.py
@@ -2,16 +2,7 @@ import inspect
 import logging
 import warnings
 from enum import Enum
-from typing import (
-    Callable,
-    Dict,
-    Optional,
-    Type,
-    Union,
-    get_args,
-    get_origin,
-    get_type_hints,
-)
+from typing import Dict, Optional, Type, Union, get_args, get_origin, get_type_hints
 
 from .descriptors import (
     BooleanSettingDescriptor,
@@ -131,7 +122,7 @@ class DeviceStatus(metaclass=_StatusMeta):
         return getattr(self._embedded[embed], prop)
 
 
-def sensor(name: str, *, unit: str = "", **kwargs):
+def sensor(name: str, *, unit: Optional[str] = None, **kwargs):
     """Syntactic sugar to create SensorDescriptor objects.
 
     The information can be used by users of the library to programmatically find out what
@@ -143,7 +134,8 @@ def sensor(name: str, *, unit: str = "", **kwargs):
     """
 
     def decorator_sensor(func):
-        property_name = func.__name__
+        property_name = str(func.__name__)
+        qualified_name = str(func.__qualname__)
 
         def _sensor_type_for_return_type(func):
             rtype = get_type_hints(func).get("return")
@@ -157,8 +149,8 @@ def sensor(name: str, *, unit: str = "", **kwargs):
 
         sensor_type = _sensor_type_for_return_type(func)
         descriptor = SensorDescriptor(
-            id=str(property_name),
-            property=str(property_name),
+            id=qualified_name,
+            property=property_name,
             name=name,
             unit=unit,
             type=sensor_type,
@@ -174,7 +166,6 @@ def sensor(name: str, *, unit: str = "", **kwargs):
 def setting(
     name: str,
     *,
-    setter: Optional[Callable] = None,
     setter_name: Optional[str] = None,
     unit: Optional[str] = None,
     min_value: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


starkillerOG:
@rytilahti this is split out from https://github.com/rytilahti/python-miio/pull/1543
Could you merge this?

miio/devicestatus.py
```diff
@@ -196,17 +187,17 @@ def setting(
...
 
     def decorator_setting(func):
-        property_name = func.__name__
+        property_name = str(func.__name__)
+        qualified_name = str(func.__qualname__)
 
-        if setter is None and setter_name is None:
-            raise Exception("Either setter or setter_name needs to be defined")
+        if setter_name is None:
+            raise Exception("Setter_name needs to be defined")
```
rytilahti, line 194: Why is this parameter being removed?

miio/devicestatus.py
```diff
@@ -143,7 +134,8 @@ def sensor(name: str, *, unit: str = "", **kwargs):
     """
 
     def decorator_sensor(func):
-        property_name = func.__name__
+        property_name = str(func.__name__)
+        qualified_name = str(func.__qualname__)
```
rytilahti, line 138: Should this change also be done for other decorators for the sake of consistency?

miio/devicestatus.py
starkillerOG: Since it will not work to put in a function directly in the decorator since the class object has not yet been created. Therefore always the setter_name schould be used and the binding of the setter methods schould take place when the class object gets created.

I tried to use the setter parameter in my code directly in the decorator, but I could not get it to work (I think it will never work).

miio/devicestatus.py
starkillerOG: yes it schould be done for all of them, I think I did.

starkillerOG: I did it for setting and sensor and those are the only 2 there are now (I think)

starkillerOG: @rytilahti are there any others?

rytilahti: Oh, sorry, I think that's right. Let's get this merged and take care of any downfall in followup PRs :+1:

miio/devicestatus.py
rytilahti: One way would be to give it a reference to a callable (like `DeviceClass.set_something`), and binding that to the instance object much like what is done for the `setter_name` one, no?`.

But that is not important at the moment, let's however keep the code in place just in case for the time being.

rytilahti:
> fix sensor unit beeing set to "" instead of None (HomeAssistant default)

It's better to keep one type of change in one PR, this also makes it easier to write more clear PR titles than "update X".

Could you move this into a separate PR? I'm wondering if we should remove the `unit` altogether from the decorator signature, making it go into the extras...

miio/devicestatus.py
starkillerOG: I don't think that works since how would you get the instance object from the `DeviceClass.set_something` callable. But I am quite new to all these decorators, still figuring it all out a bit.
Learning a lot from your coding skills!

Schould I just keep the setter parameter and add a Exception that it is not yet implemented instead?

rytilahti: Yes, just add an exception there. I'm also learning, so it's all good :-)

I think something like this could work for the binding, but I haven't tried it out:
```
from functools import partial
method = partial(setter, self)  # now the `self` is passed as the first parameter to the method calls
```

starkillerOG:
@rytilahti I think we schould leave the unit for sensors since I think it is very important for almost all sensors and this streches the importance of the unit.

Split out into these PRs:
https://github.com/rytilahti/python-miio/pull/1589
https://github.com/rytilahti/python-miio/pull/1590
https://github.com/rytilahti/python-miio/pull/1591

miio/devicestatus.py
starkillerOG: I made the exception in a seperate PR.

rytilahti:
> I think we schould leave the unit for sensors since I think it is very important for almost all sensors and this streches the importance of the unit.

Fair enough, let's keep it for now, especially as we want to create standardized set of units (in a form of enum) to make the unit information consistent across the integrations.

PR #1587, opened 2022-11-09
Implement choices_attribute for setting decorator

starkillerOG:
Will be used for Multi Map Enum select.


rytilahti (approved):
LGTM, thanks :+1:

PR #1588, opened 2022-11-09
Implement introspectable actions

starkillerOG:
Adds the last missing introspectable properties (#1495), support for actions and implements them for RoborockVacuum.
The home assistant fork using this information is available at [https://github.com/rytilahti/home-assistant/tree/xiaomi_miio/feat/entities_from_upstream](https://github.com/rytilahti/home-assistant/tree/xiaomi_miio/feat/entities_from_upstream?rgh-link-date=2022-08-13T19%3A00%3A59Z)


miio/device.py
```diff
@@ -57,6 +58,7 @@ def __init__(
         self.token: Optional[str] = token
         self._model: Optional[str] = model
         self._info: Optional[DeviceInfo] = None
+        self._actions: Optional[List[ActionDescriptor]] = None
```
rytilahti, line 61: This should be `Dict[str, ActionDescriptor]` to be consistent with to-be-added `_settings` and `_sensors` which should be taken care of in a separate PR. Then we have the descriptors nicely cached after the initial status/info call! :-)

Let me know if you want to prepare a PR for that, otherwise I will do it when I find some free cycles (I'm currently busy with the genericmiot stuff).

miio/devicestatus.py
```diff
@@ -236,3 +237,32 @@ def decorator_setting(func):
...
+
+def action(name: str, **kwargs):
+    """Syntactic sugar to create ActionDescriptor objects.
+
+    The information can be used by users of the library to programmatically find out what
+    types of actions are available for the device.
+
+    The interface is kept minimal, but you can pass any extra keyword arguments.
+    These extras are made accessible over :attr:`~miio.descriptors.ActionDescriptor.extras`,
+    and can be interpreted downstream users as they wish.
```
rytilahti, line 250: Please also update the docs in `contributing.rst` accordingly.

miio/device.py
starkillerOG: I will make a sperate PR for caching, I am now realizing how messy and big the original PR was :)

starkillerOG: Is it correct that the key schould be the `method_name`?

starkillerOG: I checked it and I do think the key schould indeed be the method_name

miio/devicestatus.py
starkillerOG: Done

starkillerOG:
@rytilahti done processing feedback, could you have another look/merge?

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -866,12 +867,14 @@ def set_dust_collection_mode(self, mode: DustCollectionMode) -> bool:
...
 
     @command()
+    @action(name="Start dust collection", icon="mdi:turbine")
     def start_dust_collection(self):
         """Activate automatic dust collection."""
         self._verify_auto_empty_support()
         return self.send("app_start_collect_dust")
 
     @command()
+    @action(name="Stop dust collection", icon="mdi:turbine")
```
rytilahti, line 877: Are you planning to add action decorators for the rest of the actions in a separate PR? It would be fine for `find()` and other actions that do not require any inputs, for those that require input parameters we need to figure out a way to deliver that information to homeassistant.

starkillerOG: HomeAssistant buttons can only do "press" action, so it is not possible to supply any inputs.
Functions that require inputs need to be implemented diffrently (select, input_number, switch etc.).

starkillerOG:
I think this can be merged.

miio/integrations/vacuum/roborock/vacuum.py
rytilahti: How about `find()`, `start()`, `stop()`, `pause()`, `home()`, `test_sound_volume()`, the actions for segment & zoned cleaning?

Basically, all actions that can be useful to be called from homeassistant should be marked as such. The homeassistant integration can then ignore creating separate "standard" actions if they are implemented by a specific platform, like vacuum.

starkillerOG: Are we not making our lives difficult by first marking them as actions that need to be added and then in HomeAssistant hardcoding to ignore those actions (because they are already implemented in the base vacuum entity)?

rytilahti: > find() is already included in the vacuum.roborock entitiy, so I don't think we need to add a seperate button.

The idea behind `VacuumInterface` was not probably a suitable one and I have been thinking that it would probably make more sense to codify this using the descriptors. This will make it simpler to support all integrations (and genericmiot devices), as there is no need to inherit from a specific base class (or implement specific methods) to let homeassistant know about the device type. It might be that we will have `genericvacuum` or something similar later on in the future, but I'd concentrate now on getting it done inside homeassistant using the descriptors.

So, at the end, the vacuum platform for homeassistant should look up if there exists an action with `vacuum:start`, `vacuum:pause` or `vacuum:find` (or whatever naming scheme is chosen) and use that to inform both 1) whether the feature is supported, and 2) provide an interface to execute the wanted action.

starkillerOG: Alright I think this is a good idea, but maybe we schould make a seperate decorator for this like "entity_action" (need to think about a good name).
Because otherwise we will end up inside the button_ng.py file of HomeAssistant with a whole list of exceptions that schould not be added (to prevent doubles).

If we make a seperate decorator for actions that schould not be added as a separate button but be part of a genric entity (like the vacuum entity or fan entity or media_player entity), I think we will make our live easier.

rytilahti: The problem is that we still need to handle it somehow based on the IDs (or similar information), as the genericmiot actions are created dynamically from JSON files. One potential way could be keeping a state of the "initialized" descriptors and initialize the "main" platforms first to let them handle the actions they want, leaving the rest uninitialized ones to the "common" platforms to handle.

starkillerOG: Normally there is no order in the initialization of the diffrent platforms inside HomeAssistant.
For now I will add the action with a extra key "type": "vacuum", that can be used to seperate them in HomeAssistant. We can figure out a better or more generic way in a future PR, but that is out of scope of this PR.

starkillerOG:
@rytilahti I added the missing actions, could you merge this PR?

miio/integrations/vacuum/roborock/vacuum.py
rytilahti: Okay, I thought we could have controlled the initialization as the platforms are normally initialized using `async_forward_entry_setups()`. But let's merge this now and take care of that later.

rytilahti (approved):
Tested to work as expected, thanks @starkillerOG! :+1:

image

PR #1589, opened 2022-11-09
Use __qualname__ to make ids unique for settings and sensors

starkillerOG:
Split out from https://github.com/rytilahti/python-miio/pull/1586

set the id to the qualified_name instead of only the function name (in order to prevent unique_id collisions in HomeAssistant for example with DNDStatus.start and CleaningDetails.start)

### DIFF
--- a/miio/devicestatus.py
+++ b/miio/devicestatus.py
@@ -143,7 +143,8 @@ def sensor(name: str, *, unit: str = "", **kwargs):
     """
 
     def decorator_sensor(func):
-        property_name = func.__name__
+        property_name = str(func.__name__)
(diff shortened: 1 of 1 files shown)


rytilahti (approved):
LGTM, I edited the title a bit :+1:
@starkillerOG On those CI errors, I'm wondering if you are basing your branches on master that does not have #1565?

rytilahti:
On those CI errors, I'm wondering if you are basing your branches on master that does not have #1565?

starkillerOG:
I schould be basing on the default https://github.com/rytilahti/python-miio/ (master) branch when I make new branches...
(I use Github Desktop)

Issue #1580, opened 2022-11-08
Test failures when run between 12am and 1am (datetime/timedelta issue)


rytilahti:
For some reason `timedelta(hours=1)` on `datetime.now()` does not change the date properly, so the tests will fail if run between 12am and 1am...

```
self = <DummyViomiDishwasher: 192.0.2.1 (token: ffffffffffffffffffffffffffffffff)>, time = FakeDatetime(2022, 11, 8, 0, 57), program = <Program.Quick: 2>

    @command(
        click.argument("time", type=click.DateTime(formats=["%H:%M"])),
        click.argument("program", type=EnumType(Program)),
    )
    def schedule(self, time: datetime, program: Program) -> str:
        """Schedule a program run.
    
        *time* defines the time when the program should finish.
        """
    
        if program == Program.Unknown:
            ValueError(f"Program {program.name} is not valid for this function.")
    
        now = datetime.now()
        scheduled_finish_date = now.replace(
            hour=time.hour, minute=time.minute, second=0, microsecond=0
        )
        scheduled_start_date = scheduled_finish_date - timedelta(
            seconds=program.run_time
        )
        if scheduled_start_date < now:
>           raise ValueError(
                f"Proposed time ({scheduled_start_date} < {now}) is in the past (the proposed time is the finishing time, not the start time)."
            )
E           ValueError: Proposed time (2022-11-07 00:29:05 < 2022-11-07 23:57:53.324466) is in the past (the proposed time is the finishing time, not the start time).

miio/integrations/viomidishwasher/viomidishwasher.py:336: ValueError
```

_Originally posted by @rytilahti in https://github.com/rytilahti/python-miio/issues/1579#issuecomment-1306397653_

PR #1591, opened 2022-11-09
Raise exception on not-implemented @setting(setter)

starkillerOG:
Split out from https://github.com/rytilahti/python-miio/pull/1586
warn for setter input from setting decorator, since it will not work to put in a function directly in the decorator since the class object has not yet been created. Therefore always the setter_name schould be used and the binding of the setter methods schould take place when the class object gets created.

### DIFF
--- a/miio/devicestatus.py
+++ b/miio/devicestatus.py
@@ -199,7 +199,9 @@ def setting(
         property_name = func.__name__
 
         if setter is None and setter_name is None:
-            raise Exception("Either setter or setter_name needs to be defined")
+            raise Exception("setter_name needs to be defined")
+        if setter_name is None:
+            raise Exception("setter not yet implemented, use setter_name instead")
 
         common_values = {
             "id": str(property_name),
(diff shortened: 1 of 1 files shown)


miio/devicestatus.py
```diff
@@ -199,7 +199,9 @@ def decorator_setting(func):
         property_name = func.__name__
 
         if setter is None and setter_name is None:
-            raise Exception("Either setter or setter_name needs to be defined")
+            raise Exception("setter_name needs to be defined")
+        if setter_name is None:
+            raise Exception("setter not yet implemented, use setter_name instead")
```
rytilahti, line 204:
```suggestion
            raise NotImplementedError("setter not yet implemented, use setter_name instead")
```

starkillerOG:
@rytilahti ready to be merged.

rytilahti:
Can be merged after the linting issue is fixed.

starkillerOG:
@rytilahti I think the linting issue is resolved.
No idea why my tests are beeing so flaiky.

I am calling it a day, will continue tommorow.
Thanks for the help and quick response today!

rytilahti:
Hehe, my pleasure! Looks like we were quite productive with short feedback cycles :-)

The CI flakiness is something that has happened before and I thought using the codecov token would help, but this time you were also encountering this issues: #1580

PR #1592, opened 2022-11-09
Initialize and cache descriptors only once

starkillerOG:
Add caching to the descriptors to prevent unnessesary IO.

Also needs to happen for action descriptors which are added in this PR: https://github.com/rytilahti/python-miio/pull/1588 (which PR gets merged first does not matter).


miio/device.py
```diff
@@ -246,14 +249,26 @@ def status(self) -> DeviceStatus:
...
         """Return device settings."""
-        settings = self.status().settings()
-        for setting in settings.values():
+        if self._settings is not None:
+            return self._settings
+
+        self._settings = (
+            self.cached_status().settings()
+        )  # NOTE that this already does IO so schould be run in executer job in HA
+        for setting in self._settings.values():
```
rytilahti, line 271: Instead of handling the initialization inside the methods returning the descriptors, maybe there should be a separate `_initialize_descriptors` that handles the initialization, leaving these getters just to return the existing, cached values?

Calling the initialization procedure could be done, e.g., as part of the initial `info()` request after the model is known?

starkillerOG: Sounds very good, I implemented it that way!

starkillerOG:
@rytilahti something is wrong with the linting check, something about a username and gitlab???
That is unrelated to this PR I think.

starkillerOG:
@rytilahti ready to be reviewd/merged

miio/device.py
```diff
@@ -126,12 +128,16 @@ def info(self, *, skip_cache=False) -> DeviceInfo:
 
         This includes information about connected wlan network, and hardware and
         software versions.
+        This also caches the descriptors for sensors, settings and actions
+        which makes additional IO calls.
```
rytilahti, line 132:
```suggestion
        This method implicitly populates the internal data structures needed prior accessing `sensors()`, `settings()`, or `actions()`.
```
I don't think it is necessary to point it out here that this might do some extra requests, considering this is an external API. It is more noteworthy to mention that this needs to be called in most of the cases :-)

miio/device.py
```diff
@@ -155,6 +161,32 @@ def _fetch_info(self) -> DeviceInfo:
...
+        self._settings = status.settings()
+        for setting in self._settings.values():
+            if setting.setter is None:
+                if setting.setter_name is None:
+                    raise Exception(
+                        f"Neither setter or setter_name was defined for {setting}"
+                    )
+                setting.setter = getattr(self, setting.setter_name)
+            if (
+                isinstance(setting, EnumSettingDescriptor)
```
rytilahti, line 184: This should check if `setting.type == SettingType.Enum` instead of using the `isinstance`.

miio/device.py
```diff
@@ -155,6 +161,32 @@ def _fetch_info(self) -> DeviceInfo:
...
+        status = self.status()
+
+        # Sensors
+        self._sensors = status.sensors()
+
+        # Settings
+        self._settings = status.settings()
+        for setting in self._settings.values():
+            if setting.setter is None:
+                if setting.setter_name is None:
```
rytilahti, line 178: This is rather deep conditional, any way to simplify?

miio/device.py
```diff
@@ -252,31 +284,21 @@ def actions(self) -> Dict[str, ActionDescriptor]:
...
+        return self._settings
 
     def sensors(self) -> Dict[str, SensorDescriptor]:
         """Return device sensors."""
-        # TODO: the latest status should be cached and re-used by all meta information getters
-        sensors = self.status().sensors()
-        return sensors
+        if self._sensors is None:
+            raise Exception(
+                "sensor descriptors schould first be cached using the info() call"
```
rytilahti, line 298: I think it would be better just to call `_initialize_descriptors` considering it's just a single method call that will return fast if already initialized.

miio/device.py
starkillerOG: thanks, done

miio/device.py
starkillerOG: Well I was doing it this way to make sure accesing the descriptors will not do IO.
For HomeAssistant this is important because we can not do IO inside the event loop.
So we need to schedule IO calls in the executer.

In this way we do not need to worry about this going wrong.

But if we are carefull in HomeAssistant about always calling .info() before using any of the descriptors we could indeed do what you suggest and just call the initialization, but I suspect this might cause a bug in the future if someone forgets.

starkillerOG: What do you think?

miio/device.py
starkillerOG: I managed to simplify it a bit to be one level less.

starkillerOG:
@rytilahti thanks for the review!
I processed all feedback :)

miio/device.py
rytilahti: So, I think it's a common design pattern (at least among homeassistant folks?) to avoid doing I/O on property accesses to make it clear it's fine to access these anywhere. As these are currently just regular methods, I think we should try to keep the API simple for now. We can discuss about the pros and cons on converting these to properties another time, but let's keep it as simple as possible for now.

miio/device.py
```diff
@@ -155,6 +161,32 @@ def _fetch_info(self) -> DeviceInfo:
...
+        for setting in self._settings.values():
+            if setting.setter_name is not None:
+                setting.setter = getattr(self, setting.setter_name)
+            if setting.setter is None:
+                raise Exception(
+                    f"Neither setter or setter_name was defined for {setting}"
+                )
+            if (
+                setting.type == SettingType.Enum
+                and setting.choices_attribute is not None
```
rytilahti, line 185: Hmmh, I suppose we don't currently have tests for this case, but wouldn't we just fail silently if only `choices` if defined?

miio/device.py
```diff
@@ -252,31 +284,21 @@ def actions(self) -> Dict[str, ActionDescriptor]:
...
-                setting.setter = getattr(self, setting.setter_name)
-            if (
-                isinstance(setting, EnumSettingDescriptor)
-                and setting.choices_attribute is not None
-            ):
-                retrieve_choices_function = getattr(self, setting.choices_attribute)
-                setting.choices = retrieve_choices_function()  # This can do IO
+        if self._settings is None:
+            raise Exception(
+                "setting descriptors schould first be cached using the info() call"
```
rytilahti, line 289:
```suggestion
                "setting descriptors should first be cached using the info() call"
```
Btw, you often type `schould` instead of `should` :-)

miio/device.py
starkillerOG: done

miio/device.py
starkillerOG: Well ether `choices` or `choices_attribute` needs to be defined in order to get the `SettingType.Enum` in the decorator. We only need `choices` so if `choices` is defined (as a static enum) everything will work just fine.
If `choices_attribute` is defined (for dynamic enums) the `choices` will be set from a call to the `choices_attribute` function.

So no we will not fail silently

starkillerOG:
@rytilahti thanks for the feedback!
Ready for another round of comments.

rytilahti (commented):
Considering this is the main device class, please add some tests for the expected functionalities and this is then ready to be merged.

starkillerOG:
@rytilahti I rebased on top of the main branch.
What kind of tests do you imagine?
I am not so good in tests ;)
I don't see any previous tests for the descriptors...

miio/device.py
```diff
@@ -248,45 +290,26 @@ def status(self) -> DeviceStatus:
         """Return device status."""
         raise NotImplementedError()
 
-    def actions(self) -> Dict[str, ActionDescriptor]:
+    def actions(self) -> Optional[Dict[str, ActionDescriptor]]:
```
rytilahti, line 293: I don't think these should be marked as `Optional`: if the device has no actions defined, we should just have an empty list. This will make the API cleaner as the downstreams do not need to check against noneness but can simply iterate over the list directly.

miio/device.py
```diff
@@ -157,6 +164,41 @@ def _fetch_info(self) -> DeviceInfo:
...
+                raise Exception(
+                    f"Neither setter or setter_name was defined for {setting}"
+                )
+            setting = cast(EnumSettingDescriptor, setting)
+            if (
+                setting.type == SettingType.Enum
+                and setting.choices_attribute is not None
+            ):
+                retrieve_choices_function = getattr(self, setting.choices_attribute)
+                setting.choices = retrieve_choices_function()
```
rytilahti, line 192: I'd extract this into its own internal method, and then use:
```
self._settings = self._setting_descriptors_from_status(status)
self._sensors_ = self._sensor_descriptors_from_status(status)
```
etc. This will allow testing them separately and just make the code nicer.

miio/device.py
rytilahti: Was this closed accidentally? :)

miio/device.py
starkillerOG: Yes mypy was complaining, will fix it another way.

miio/device.py
starkillerOG: I seperated settings/sensors/actions.

starkillerOG:
@rytilahti I discovered a bit of a chicken and egg problem during testing:
Apprently some devices like the airdehumidifier have the following problem:
When calling status() they call info() which calls _initialize_descriptors() which calls status() and you get into a loop...

Do you know how to solve this?
Besides obviously moving _initialize_descriptors out of info()?

rytilahti:
@starkillerOG looks like airdehumidifier is passing the info even when it is not used at all, so in this case it should be enough to remove that call, no?

I think this should be checked case-by-case basis, but one possibility would be requiring making an `info()` call prior trying to access any other device commands. This is tangentially related to one of the missing pieces for descriptors, namely how to handle settings whose choices or ranges depend on the model, which is relevant for fans and lights with differing features and characteristics.

rytilahti (commented):
Oops, I wrote some comments earlier but forgot to submit them, sorry for the delay.

miio/device.py
```diff
@@ -30,9 +31,10 @@ class UpdateState(Enum):
 class Device(metaclass=DeviceGroupMeta):
     """Base class for all device implementations.
 
-    This is the main class providing the basic protocol handling for devices using the
-    ``miIO`` protocol. This class should not be initialized directly but a device-
-    specific class inheriting it should be used instead of it.
+    This is the main class providing the basic protocol handling for
+    devices using the ``miIO`` protocol. This class should not be
+    initialized directly but a device- specific class inheriting it
+    should be used instead of it.
```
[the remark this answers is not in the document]
starkillerOG: Yes, sorry, docformatter did this, I removed it now.

miio/device.py
```diff
@@ -106,8 +110,8 @@ def send_handshake(self):
         click.argument("parameters", type=LiteralParamType(), required=False),
     )
     def raw_command(self, command, parameters):
-        """Send a raw command to the device. This is mostly useful when trying out
-        commands which are not implemented by a given device instance.
+        """Send a raw command to the device. This is mostly useful when trying
+        out commands which are not implemented by a given device instance.
```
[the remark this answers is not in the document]
starkillerOG: Yes, sorry, docformatter did this, I removed it now.

miio/device.py
```diff
@@ -136,6 +140,12 @@ def info(self, *, skip_cache=False) -> DeviceInfo:
 
         return self._fetch_info()
 
+    def initialize(self) -> None:
+        """This method implicitly populates the internal data structures needed
+        prior accessing `info()`, `sensors()`, `settings()`, or `actions()`."""
+        self.info()
```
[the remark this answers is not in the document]
starkillerOG: Alright, I removed the initialize method, and moved it back to the info() call.
I changed the other parts of the library to call _fetch_info() instead of info() to prevent these circular errors.

miio/device.py
```diff
@@ -252,41 +312,25 @@ def actions(self) -> Dict[str, ActionDescriptor]:
...
-
-                setting.setter = getattr(self, setting.setter_name)
-            if (
-                isinstance(setting, EnumSettingDescriptor)
-                and setting.choices_attribute is not None
-            ):
-                retrieve_choices_function = getattr(self, setting.choices_attribute)
-                setting.choices = retrieve_choices_function()  # This can do IO
+        if self._settings is None:
+            self._settings = {}
```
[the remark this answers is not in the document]
starkillerOG: Done, this was just a typing thing.

miio/device.py
```diff
@@ -252,41 +312,25 @@ def actions(self) -> Dict[str, ActionDescriptor]:
...
-        return settings
+        return self._settings
 
     def sensors(self) -> Dict[str, SensorDescriptor]:
         """Return device sensors."""
-        # TODO: the latest status should be cached and re-used by all meta information getters
-        sensors = self.status().sensors()
-        return sensors
+        if self._sensors is None:
+            self._sensors = {}
```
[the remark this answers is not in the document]
starkillerOG: done

starkillerOG:
@rytilahti I finnally got all tests working and fixed the looping issues with the info() call.
Could you have another look/merge?

miio/airdehumidifier.py
```diff
@@ -185,7 +185,7 @@ def status(self) -> AirDehumidifierStatus:
         values = self.get_properties(properties, max_properties=1)
 
         return AirDehumidifierStatus(
-            defaultdict(lambda: None, zip(properties, values)), self.info()
+            defaultdict(lambda: None, zip(properties, values)), self._fetch_info()
```
rytilahti, line 188: This is wrong, the `_fetch_info()` is device-class internals and should not be accessed by any other class.

starkillerOG: Well how do you then suggest to prevent the loop: status --> info --> _initialize_descriptors--> status --> etc.
I can just remove the info call, but I am not sure what will then break downstream.

rytilahti: I'm not sure what's the best approach right off the bat (especially as genericmiot constructs status containers dynamically), but for statically defined containers it would be enough to access the status container class without really initializing one, right?

```
In [5]: from miio.integrations.vacuum.roborock import VacuumStatus

In [6]: vars(VacuumStatus)
Out[6]: 
mappingproxy({'__module__': 'miio.integrations.vacuum.roborock.vacuumcontainers',
<snip>
              '_sensors': {'state_code': SensorDescriptor(id='VacuumStatus.state_code', type=<class 'int'>, name='State code', property='state_code', unit=None),
               'state': SensorDescriptor(id='VacuumStatus.state', type=<class 'str'>, name='State', property='state', unit=None),
<snip>
              '_settings': {'fanspeed': NumberSettingDescriptor(id='VacuumStatus.fanspeed', name='Fanspeed', property='fanspeed', unit='%', min_value=0, max_value=100, step=1, range_attribute=None, type=<SettingType.Number: 2>),
<snip>
```

Probably the simplest way to access the status container would be using the return type annotation like this:
```
_status_container = self_.status.__annotations__["return"]
```

Do you think that could work for the time being until we figure out a better solution for that?

starkillerOG: I just tried it, but that does not include the embeded sensors:

```
from miio import RoborockVacuum

r = RoborockVacuum(ip = "192.168.1.IP", token = "TokenTokenToken")
status = r.status.__annotations__["return"]
print(status.sensors(status))
```

That is missing the embeded sensors/settings....

rytilahti: Ouch, that's indeed annoying... The real solution would be doing the "embedding" of status container _classes_ outside of the status query (maybe inside `__init__`), and performing the update on the container prior to returning it in `status()`. This way:
1. we can introspect the available information without performing any I/O.
2. if some of the method calls required to embedded containers fail, we can remove those queries from future update cycles.

The question is how/where to do this sanely, I don't have an answer to that right now.

starkillerOG:
I will make some additional tests once the actual code is approved and we decided how to handle the info call.

miio/airdehumidifier.py
starkillerOG: I do see what you mean, and think it is a good path forward,
but I think at this point it gets a bit to complicated for me.

I think it would be more efficient if you would implement those changes.
Than we also do not need to have long wait times in between us working on it.

You could do one of the following:
- make the changes and push them to this branch.
- copy this code as a start, make your own branch with a new PR and close this one
- First merge this as a starting point and then make the changes later to get rid of the _fetch_info calls in device classes (can be added as a TO DO)

starkillerOG:
Closing due to merge conflicts, follow up PR is here: https://github.com/rytilahti/python-miio/pull/1701

PR #1596, opened 2022-11-14
Add multi map handling to roborock

starkillerOG:

Add basic multi map support to roborock vacuums
 - Get current map
 - Set current map

Later PR will add clean details per floor and floor information to last clean details.


starkillerOG:
@rytilahti ready for review/merge

starkillerOG:
@rytilahti thanks for the quick revieuws the last few days, really making great progress!
Can you also review/merge this, so I can continue with the last part of this (the per floor cleaning details).

starkillerOG:
@rytilahti could you have a look at this and merge?
This is the straight forward part of the multi map support.

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -131,6 +133,8 @@ def __init__(
     ):
         super().__init__(ip, token, start_id, debug, model=model)
         self.manual_seqnum = -1
+        self._multi_maps: Optional[MultiMapList] = None
+        self._multi_map_enum = None
```
rytilahti, line 137: Do we really need to keep so much state in the class?

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -94,6 +97,46 @@ def pretty_area(x: float) -> float:
 }
 
 
+class MultiMapList(DeviceStatus):
+    """Contains a information about the maps/floors of the vacuum."""
+
+    def __init__(self, data: Dict[str, Any]) -> None:
+        # {'max_multi_map': 4, 'max_bak_map': 1, 'multi_map_count': 3, 'map_info': [
```
rytilahti, line 104: Considering we are slowly converting to pydantic (e.g., for genericmiot), I'm wondering should we also do the same here to simplify the parsing and validation?

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -361,6 +365,36 @@ def map(self):
...
+
+    @command(click.argument("multi_map_id", type=int))
+    def load_multi_map(self, multi_map_id: int):
+        """Change the current map used."""
+        return self.send("load_multi_map", [multi_map_id])[0] == "ok"
+
+    @command()
+    def load_multi_map_by_enum(self, multi_map_enum):
+        """Change the current map used by enum."""
+        return self.load_multi_map(multi_map_enum.value)
```
rytilahti, line 396: Having two separate setters sounds conceptually really odd and broken API, but I don't have currently no answer what'd be the best approach here.

rytilahti (commented):
I have had these pending reviews for a while, but didn't want to send them before offering a solution for some of them. Considering I'm busy with other things at the moment, I thought I'll just send the feedback as-is. Please take a look @starkillerOG and let me know what you think.

Also, you could create a PR for the part 2 (or merge with this?), if you think it will make it simpler to review the concepts.

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -94,6 +97,46 @@ def pretty_area(x: float) -> float:
...
+        #    {'mapFlag': 0, 'add_time': 1664448893, 'length': 10, 'name': 'Downstairs', 'bak_maps': [{'mapFlag': 4, 'add_time': 1663577737}]},
+        #    {'mapFlag': 1, 'add_time': 1663580330, 'length': 8, 'name': 'Upstairs', 'bak_maps': [{'mapFlag': 5, 'add_time': 1663577752}]},
+        #    {'mapFlag': 2, 'add_time': 1663580384, 'length': 5, 'name': 'Attic', 'bak_maps': [{'mapFlag': 6, 'add_time': 1663577765}]}
+        #  ]}
+        self.data = data
+        if self.map_count != len(self.data["map_info"]):
+            _LOGGER.warning("Roborock multi_map_count does not equal amount of maps")
```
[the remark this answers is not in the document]
starkillerOG: Sure, I removed the warning, we will just trust Roborock that they implemented it sainly.
If we get any issues from users, we will look into it then.

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -94,6 +97,46 @@ def pretty_area(x: float) -> float:
...
+
+        self._map_name_dict = {}
+        for idx, map in enumerate(self.data["map_info"]):
+            self._map_name_dict[map["name"]] = map["mapFlag"]
+            if map["mapFlag"] != idx:
+                _LOGGER.warning("Roborock mapFlag does not equal map_info list index")
```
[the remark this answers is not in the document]
starkillerOG: see above

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -361,6 +365,36 @@ def map(self):
...
+        """Return list of multi maps."""
+        if self._multi_maps is not None and not skip_cache:
+            return self._multi_maps
+
+        self._multi_maps = MultiMapList(self.send("get_multi_maps_list")[0])
+        return self._multi_maps
+
+    @command()
+    def multi_map_enum(self, skip_cache=False) -> Optional[enum.Enum]:
+        """Enum of the available map names."""
```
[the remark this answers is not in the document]
starkillerOG: Yes that is nessesary, the get_multi_maps container is nicely fromatted and contains all information.
The multi_map_enum does not contain all information, but is nessesary for the setting decorator:
```
    @setting(
        "Multi map",
        choices_attribute="multi_map_enum",
        setter_name="load_multi_map_by_enum",
        icon="mdi:floor-plan",
    )
```
The decorator can only handle a function that directly returns the enum, it can not handle containers...

miio/integrations/vacuum/roborock/vacuum.py
starkillerOG: Well we need to cache it somewhere, in normal use cases the amount of floors of a house does not change and therefore the multi_map information does not change.
I find it a bit resource wastefull to do IO on each status call to retrieve the same map information over and over again while we kind of know it will never change.

starkillerOG: If someone actually renames a floor or builds a new floor to there house, they just need to restart homeassistant or reload the integration.

miio/integrations/vacuum/roborock/vacuumcontainers.py
starkillerOG: I don't know what pydantic is...
Maybe we can do that in a seperate PR?

miio/integrations/vacuum/roborock/vacuum.py
starkillerOG: I merged them to a single setter: load_multi_map with two optional arguments, ether the multi_map_id or the multi_map_enum needs to be supplied.

starkillerOG:
@rytilahti I finished processing all feedback, can you have another look?

starkillerOG:
@rytilahti All checks pass now, the error is unrelated.
Part 2 of the mult map support is a bit more difficult, it is basically what is left in https://github.com/rytilahti/python-miio/pull/1543.

But because part 2 builds upon part 1 I will get lots of merge conflicts when we change stuff in part 1.
I don't want to make one big PR because that will take a lot longer to review.

I think this part 1 is really straight forward and can be merged quite fast.
Lets get this merged now and then I will submit part 2.

starkillerOG:
@rytilahti could you look at my responses/merge?

miio/integrations/vacuum/roborock/vacuumcontainers.py
rytilahti: Okay, so pydantic would allow defining the structure using class annotations like this:
```
class MapInfo(BaseModel):
    flags: int = Field(alias="mapFlag")
    added: datetime = Field(alias="add_time")
    name: Optional[str]  # backup maps contain no name
    backup_maps: Optional[List["MapInfo"]]  # backup maps contain now backups

class MultiMapList(BaseModel):
    maps: List[MapInfo] = Field(alias="map_info")
```
and it would perform the parsing automatically.

The problem I see though is that the DeviceStatus bases itself on the defined properties, so this would need further adjustments... So let's keep it as it is for now.

miio/integrations/vacuum/roborock/vacuum.py
rytilahti: Okay, fair enough. I'm not that happy about having hidden state information stored in the device instance, but let's keep that for now.

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -365,6 +369,41 @@ def map(self):
...
+
+    @command(click.argument("multi_map_id", type=int))
+    def load_multi_map(
+        self,
+        multi_map_enum: Optional[enum.Enum] = None,
+        multi_map_id: Optional[int] = None,
+    ):
+        """Change the current map used."""
+        if multi_map_enum is None and multi_map_id is None:
+            _LOGGER.error("Either multi_map_enum or multi_map_id is required.")
```
rytilahti, line 400: Should this rather raise an exception if those are really required?

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -502,6 +555,11 @@ def area(self) -> float:
         """Total cleaned area."""
         return pretty_area(self.data["area"])
 
+    @property
+    def multi_map_id(self) -> int:
```
rytilahti, line 559:
```suggestion
    def current_map_id(self) -> int:
```
Maybe to make it clearer what it is?

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -365,6 +369,41 @@ def map(self):
         # returns ['retry'] without internet
         return self.send("get_map_v1")
 
+    @command()
+    def get_multi_maps(self) -> MultiMapList:
```
rytilahti, line 373:
```suggestion
    def get_maps(self) -> MapList:
```
What do you think about renaming these?

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -365,6 +369,41 @@ def map(self):
...
+        """Return list of multi maps."""
+        if self._multi_maps is not None:
+            return self._multi_maps
+
+        self._multi_maps = MultiMapList(self.send("get_multi_maps_list")[0])
+        return self._multi_maps
+
+    @command()
+    def multi_map_enum(self) -> Optional[enum.Enum]:
+        """Enum of the available map names."""
```
rytilahti, line 383: I thought I mentioned about this, but if this is just for internal use, removing `@command()` and prefixing the name with `_` should make it clear for anyone using the API.

miio/integrations/vacuum/roborock/vacuum.py
starkillerOG: Yes I changed it to a raise ValueError

miio/integrations/vacuum/roborock/vacuumcontainers.py
starkillerOG: Yes I agree, I changed it

miio/integrations/vacuum/roborock/vacuum.py
starkillerOG: I have renamed all "multi_map" to "map"

miio/integrations/vacuum/roborock/vacuum.py
starkillerOG: I renamed the function to _map_enum, the internal variable to _map_enum_cache.
I also removed the @command()

starkillerOG:
@rytilahti I have finished processing your feedback, thanks!
Do you have any additional comments?

rytilahti (approved):
This is good to go as soon as the CI passes, thanks @starkillerOG! :+1: edit: could you please add tests that `MapList` is parsed properly (and property accesses return what's expected)?

starkillerOG:
@rytilahti I added the test you suggested, how can I easily run the test locally?
(I think you have a poetry command for that right)?

starkillerOG:
Thanks, I will start working on the final part of the multi maps handeling

PR #1597, opened 2022-11-15
Use type instead of string for SensorDescriptor type

rytilahti:
The descriptor reports now the return type of the property instead of artificial "sensor" or "binary_sensor", leaving it to the downstream to handle it as they wish.


rytilahti:
Ping @starkillerOG just so that you are also in the loop, this fixes the case where binary sensors (that is, sensors returning bool) were constructed as regular sensors in the homeassistant fork.

PR #1601, opened 2022-11-22
Off fan speed for Roborock S7

rogelio-o:
The Roborock S7 MaxV includes the Off fan speed, but the S7 does not. That speed is also valid for the S7 (it can only use the mop too). Checked with `firmware 4.3.5_1578`.


rytilahti (approved):
Thanks @rogelio-o!

PR #1559, opened 2022-10-24
Improve viomi.vacuum.v8 (styj02ym) support

rytilahti:
**Breaking change: the naming of properties and methods have been adjusted to be more clear about their intention, enums with boolean values have been replaced with booleans.**

This PR improves support for viomi vacuums and aims to make it available for homeassistant without custom components (like https://github.com/KrzysztofHajdamowicz/home-assistant-vacuum-styj02ym). While this has been developed against "Mi Robot Vacuum-Mop P" (viomi.vacuum.v8, styj02ym), it should work on all other viomivacuum supported vacuums. Thanks to @KrzysztofHajdamowicz for providing access to a test device :-)

This is related to #1495 and requires using https://github.com/rytilahti/home-assistant/tree/xiaomi_miio/feat/entities_from_upstream for the time being. The screenshot below shows the current (WIP) state.

image

Closes #1161
Closes #750
Closes #1003

TODO:
* [x] Consolidate common vacuum API (e.g., states need to be standardized for homeassistant to make it compatible with the vacuum platform)
* [x] Sensor/setting/switch icons and other metadata
* [x] Add actions 
* [-] Add buttons (reset consumables, ..)

Some notes:
* The class previously requested properties unknown for this model, causing the mismatch in the number of elements requested vs. reported.
* The device seems to be rather unresponsive for requests in many cases, this needs some investigation.

PR #1603, opened 2022-11-23
Expose dnd status, add actions for viomivacuum

rytilahti:
Forgot to push this commit in #1559 ..


jeremysherriff:
Hiya @rytilahti are you planning to add this and #1559 to a release?

rytilahti:
Hi @jeremysherriff, those will be part of the 0.6 release when it gets done, but there are still some work to do until that happens. If there will be any new 0.5 releases, those are not going to be included within as they require too many breaking changes.

PR #1608, opened 2022-12-03
Add Roborock S7 MaxV Ultra station sensors

jpbede:
This PR adds two sensors of the auto-emptying and washing station of the Roborock S7 MaxV.

### DIFF
--- a/miio/integrations/vacuum/roborock/vacuumcontainers.py
+++ b/miio/integrations/vacuum/roborock/vacuumcontainers.py
@@ -714,6 +714,32 @@ class ConsumableStatus(DeviceStatus):
             return self.data["dust_collection_work_times"]
         return None
 
+    @property
+    @sensor(
+        "Strainer used",
+        icon="mdi:air-filter",
+        entity_category="diagnostic",
+        enabled_default=False,
+    )
+    def strainer_used(self) -> Optional[int]:
+        """Return ``strainer_work_times``"""
+        if "strainer_work_times" in self.data:
+            return self.data["strainer_work_times"]
+        return None
+
+    @property
+    @sensor(
+        "Cleaning brush used",
+        icon="mdi:brush",
+        entity_category="diagnostic",
+        enabled_default=False,
+    )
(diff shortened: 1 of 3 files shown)


starkillerOG:
Looks good to me

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -656,6 +656,32 @@ def dustbin_auto_empty_used(self) -> Optional[int]:
...
 
+    @property
+    @sensor(
+        "Strainer used",
+        icon="mdi:air-filter",
+        entity_category="diagnostic",
+        enabled_default=False,
+    )
+    def strainer_used(self) -> Optional[int]:
+        """Return ``strainer_work_times``"""
```
rytilahti, line 667:
```suggestion
    def strainer_cleaned_count(self) -> Optional[int]:
        """Return strained cleaned count."""
```
It's a bit unclear what does this mean? Is it a count of times the strainer has been cleaned? If yes, maybe change the wording.

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -656,6 +656,32 @@ def dustbin_auto_empty_used(self) -> Optional[int]:
...
+
+    @property
+    @sensor(
+        "Cleaning brush used",
+        icon="mdi:brush",
+        entity_category="diagnostic",
+        enabled_default=False,
+    )
+    def cleaning_brush_used(self) -> Optional[int]:
+        """Return ``cleaning_brush_work_times``"""
```
rytilahti, line 680: Same for this, is this also a count? Please use descriptive docstrings to describe what type of information is returned instead of just naming what is the internal name for the piece of information.

rytilahti (changes requested):
LGTM, but could you check the wording of the properties & add simple tests for these two?

jpbede:
@starkillerOG @rytilahti thanks for your review. I've implemented the suggestions and added two tests for both properties.

PR #1614, opened 2022-12-05
Add multi map handling to roborock part 2

starkillerOG:
Part 2 of adding multi maps handeling for Roborock vacuums.
- Clean details per floor
- better caching of clean details and history to minimize IO calls.
- current map name
- last clean details map name

### DIFF
--- a/miio/integrations/roborock/vacuum/vacuum.py
+++ b/miio/integrations/roborock/vacuum/vacuum.py
@@ -533,7 +533,7 @@ class RoborockVacuum(Device, VacuumInterface):
         """
         if self._clean_history is None or skip_cache:
             self.clean_history()
-        assert isinstance(self._clean_history, CleaningSummary) # nosec assert_used
+        assert isinstance(self._clean_history, CleaningSummary)  # nosec assert_used
         if not self._clean_history.ids:
             return None
 
@@ -552,7 +552,7 @@ class RoborockVacuum(Device, VacuumInterface):
         """
         if self._clean_history is None or skip_cache:
             self.clean_history()
-        assert isinstance(self._clean_history, CleaningSummary) # nosec assert_used
+        assert isinstance(self._clean_history, CleaningSummary)  # nosec assert_used
 
         map_ids = self.get_maps().map_id_list
 


starkillerOG:
@rytilahti this is the last piece of the Roborock s7 MaxV+ support.
Could you make a first review and profide some feedback?

I know you were not that happy at the FloorCleanDetails class beeing very technicall and difficult to read (because it is dynammically constructed), but I don't know how I could improve that withouth just hard coding the 4 floors that are currently supported as a max, but you initially also did not like that.

Also passing in the MapList class to other classes did not have your prefrence, It is needed for mapping the map_id to the user frendly map name. How would you suggest to do that diffrently?

starkillerOG:
@rytilahti could you review so we can get this merged?

starkillerOG:
@rytilahti would you have some time this week or next week to make an appointment on discord to revieuw this.
I would really like to get this finished.

Maybe we could also discuss releasing a new python-miio version?
It has been 5 months since the last release, I think we schould at least aim to have a monthly release of python-miio since there are so many (merged) PRs for it.
It is really starting to hold up development on the HomeAssistant side.

rytilahti:
Hi @starkillerOG and happy belated christmas! Sorry I haven't answered to you, the holiday season was (and remains) busy, but I'll try to find some time to look into this come next week. We can also have a chat on discord if you want, just give me a ping some time next week whenever it suits you.

starkillerOG:
@rytilahti that would be awesom!
Happy new year

starkillerOG:
@rytilahti I have updated this PR to use the new UpdateHelper, have simplified a few things and are not passing around the maplist anymore but use the _parent refrence.

Could you review this?

starkillerOG:
All tests (except for codecov) are now passing

starkillerOG:
@rytilahti could you please have a look?

miio/integrations/roborock/vacuum/vacuumcontainers.py
```diff
@@ -605,6 +627,57 @@ def complete(self) -> bool:
...
+    def __init__(self, data: Dict[int, Any]) -> None:
+        self.data = data
+
+        for map_id in self.data:
+            if self.data[map_id] is None:
+                setattr(self, f"CleanDetails_{map_id}", None)
+                setattr(self, f"start_{map_id}", None)
+                continue
+            setattr(self, f"CleanDetails_{map_id}", self.data[map_id])
+            setattr(self, f"start_{map_id}", self.data[map_id].start)
```
rytilahti, line 642: `setattr`, `getattr` etc. should be avoided wherever possible as that breaks static code analysis. It'd be better to avoid this, store the data as it's already done, and expose the necessary information through regular methods using that data.

miio/integrations/roborock/vacuum/vacuumcontainers.py
```diff
@@ -586,6 +597,17 @@ def map_id(self) -> int:
         """Map id used (multi map feature) during the cleaning run."""
         return self.data.get("map_flag", 0)
 
+    @property
+    @sensor("Last clean map name", icon="mdi:floor-plan", entity_category="diagnostic")
+    def map_name(self) -> str:
+        """The name of the map used (multi map feature) during the cleaning run."""
+        try:
+            map_list = self._parent.map_list.map_list
```
rytilahti, line 605: The need to access `map_list` twice like this feels like a code smell...

miio/integrations/roborock/vacuum/vacuum.py
```diff
@@ -515,20 +522,68 @@ def enable_lab_mode(self, enable):
...
+
+    @command()
+    def last_clean_all_floor(self, skip_cache=False) -> FloorCleanDetails:
+        """Return details from the last cleaning and for each floor.
+
+        Returns None if there has been no cleanups for that floor.
+        """
+        if self._clean_history is None or skip_cache:
+            self.clean_history()
+        assert isinstance(self._clean_history, CleaningSummary)  # nosec assert_used
```
rytilahti, line 555: Let's remove the caching from this PR and concentrate on getting the multi map handling implemented.

rytilahti (commented):
Ah, looks like commenting on this slipped through the cracks, sorry @starkillerOG and thanks for the reminder!

I added a couple of comments inline, but the general feeling that I have is that merging this as-is could introduce potentially hard to track down bugs, especially related to the use of `setattr` and the changes to caching mechanisms.

My proposal is following:
1. Let's remove, caching `set/getattr` tricks etc. to keep this PR as simple as possible. A single I/O request or a few is not a big deal in the big picture, and we can always come back to that after we have a clean, working solution.
2. We clean up the vacuumcontainers by moving the code into feature-specific files (i.e., by the command that is used to get the data). This will make it not only easier to maintain but easier to add new features.
3. We find a way to introduce caching, if it still makes sense.

As I'm not seeing any tests or payloads for the new information here, I have no idea about what would be the best way forward on this. When/if you find time, ping me on discord and let's have a synchronous chat/call to figure this all out :-)

PR #1621, opened 2022-12-07
Add mop dryer add-on of the S7 MaxV Ultra station

jpbede:
The S7 MaxV Ultra station (and AFAIK also the S7 Ultra station) has an optional mop dryer add-on. Based on reverse engineering of the Mi Home app plugin code, I added support for this.

Currently this code is partially untested as I am waiting for my add-on to be delivered. Delivery is expected around 12/19.

But I appreciate code review and feedback already :)

### DIFF
--- a/miio/integrations/vacuum/roborock/tests/test_vacuum.py
+++ b/miio/integrations/vacuum/roborock/tests/test_vacuum.py
@@ -41,70 +41,71 @@ class DummyVacuum(DummyDevice, RoborockVacuum):
         self._maps = None
         self._map_enum_cache = None
 
-        self.dummies = {}
-        self.dummies["consumables"] = [
-            {
-                "filter_work_time": 32454,
-                "sensor_dirty_time": 3798,
-                "side_brush_work_time": 32454,
-                "main_brush_work_time": 32454,
-                "strainer_work_times": 44,
-                "cleaning_brush_work_times": 44,
-            }
-        ]
-        self.dummies["clean_summary"] = [
-            174145,
-            2410150000,
-            82,
-            [
-                1488240000,
-                1488153600,
-                1488067200,
-                1487980800,
-                1487894400,
-                1487808000,
-                1487548800,
+        self.dummies = {
+            "consumables": [
+                {
+                    "filter_work_time": 32454,
+                    "sensor_dirty_time": 3798,
+                    "side_brush_work_time": 32454,
+                    "main_brush_work_time": 32454,
+                    "strainer_work_times": 44,
+                    "cleaning_brush_work_times": 44,
+                }
             ],
-        ]
-        self.dummies["dnd_timer"] = [
-            {
-                "enabled": 1,
-                "start_minute": 0,
-                "end_minute": 0,
-                "start_hour": 22,
-                "end_hour": 8,
-            }
-        ]
-        self.dummies["multi_maps"] = [
-            {
-                "max_multi_map": 4,
-                "max_bak_map": 1,
-                "multi_map_count": 3,
-                "map_info": [
-                    {
-                        "mapFlag": 0,
-                        "add_time": 1664448893,
-                        "length": 10,
-                        "name": "Downstairs",
-                        "bak_maps": [{"mapFlag": 4, "add_time": 1663577737}],
-                    },
-                    {
-                        "mapFlag": 1,
-                        "add_time": 1663580330,
-                        "length": 8,
-                        "name": "Upstairs",
-                        "bak_maps": [{"mapFlag": 5, "add_time": 1663577752}],
-                    },
-                    {
-                        "mapFlag": 2,
-                        "add_time": 1663580384,
-                        "length": 5,
-                        "name": "Attic",
-                        "bak_maps": [{"mapFlag": 6, "add_time": 1663577765}],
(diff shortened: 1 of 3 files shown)


miio/integrations/vacuum/roborock/tests/test_vacuum.py
```diff
@@ -443,10 +444,63 @@ def test_cleaning_brush_cleaned_count(self):
         """Test getting cleaning brush cleaned count."""
         assert self.device.consumable_status().cleaning_brush_cleaned_count == 44
 
+    def test_mop_dryer_model_check(self):
```
rytilahti, line 447: What do you think about simplifying this using pytest.mark.parametrize?
Something like this (untested):
```
@pytest.mark.parametrize("method_name,params", [("mop_dryer_settings", None), ("set_mop_drier_enabled", {"enabled": True}), ...])
def test_mop_drier_raises_unsupported(method_name, params):
    with pytest.raises(UnsupportedFeatureException):
        method = getattr(self.device, method_name)
        if params is not None:
            method(**params)
         else:
            method()
```

miio/integrations/vacuum/roborock/tests/test_vacuum.py
```diff
@@ -458,12 +512,36 @@ def dummyvacuums7(request):
...
+
+    def test_set_mop_dryer_dry_time_model_check(self):
+        """Test setting mop dryer dry time."""
+        assert self.device.set_mop_dryer_dry_time(dry_time=2)
+
+    def test_start_mop_drying_model_check(self):
+        """Test starting mop drying."""
+        assert self.device.start_mop_drying()
+
+    def test_stop_mop_drying_model_check(self):
```
rytilahti, line 545: The setters should assert that the device state changes accordingly.

miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -958,6 +959,52 @@ def set_child_lock(self, lock: bool) -> bool:
         """Set child lock setting."""
         return self.send("set_child_lock_status", {"lock_status": int(lock)})[0] == "ok"
 
+    def _verify_mop_dryer_supported(self) -> None:
+        """Checks if model supports mop dryer add-on."""
+        # dryer add-on is only supported by following models
+        if self.model not in [ROCKROBO_S7, ROCKROBO_S7_MAXV]:
```
rytilahti, line 965: Btw, could you run `get_fw_features` raw command and report the information you get at https://github.com/marcelrv/XiaomiRobotVacuumProtocol/blob/master/fw_features.md ? Maybe we can remove the hardcoded models and use the device-provided information in the future :-)

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -410,6 +410,32 @@ def got_error(self) -> bool:
...
+        return None
+
+    @property
+    @sensor(
+        "Dryer remaining seconds",
+        unit="s",
+        entity_category="diagnostic",
+        enabled_default=False,
+    )
+    def mop_dryer_remaining_seconds(self) -> Optional[int]:
```
rytilahti, line 433: Maybe this should return a timedelta instead?

miio/integrations/vacuum/roborock/vacuumcontainers.py
```diff
@@ -953,3 +979,40 @@ def current_high(self) -> int:
...
+        icon="mdi:fan",
+        unit="hour",
+        min_value=2,
+        max_value=4,
+        step=1,
+        entity_category="config",
+        enabled_default=False,
+    )
+    def dry_time(self) -> bool:
+        """Return mop dry time."""
```
rytilahti, line 1017: What is the meaning of dry_time? The setting for how long it will be drierd?

The return value should likely be timedelta here, too.

rytilahti (commented):
It would be great to get it also tested on a real device, but looks good on the surface to me, thanks for the PR @jpbede!

miio/integrations/vacuum/roborock/tests/test_vacuum.py
jpbede: AFAIK `@pytest.mark.parametrize` is not usable when using `TestCase` because the reference to `self` is missing.

miio/integrations/vacuum/roborock/vacuumcontainers.py
jpbede: Also changed to timedelta.

jpbede:
I've tested it now with a real device. Unfortunately the add-on is not reporting `mop_dryer_remaining_seconds` and `is_mop_drying`, so I need to remove the check in `_verify_mop_dryer_supported`.

Starting, stopping and setting the dry time is working.

miio/integrations/vacuum/roborock/vacuum.py
jpbede: Will do :)

rytilahti (approved):
LGTM, thanks for the PR @jpbede!

This does not currently modify the `status()` logic to allow exposing them directly to homeassistant, but that requires some internal changes on how the embedding is done so I think this is good to be merged now.

PR #1630, opened 2022-12-18
Add support for pet waterer mmgg.pet_waterer.wi11

Alex-ala:
The pet waterer mmgg.pet_waterer.wi11 has the "fault" and "on" ids switched.
https://home.miot-spec.com/spec?type=urn:miot-spec-v2:device:pet-drinking-fountain:0000A067:mmgg-wi11:1

The petwaterdispenser device was updated to support all 3 versions of water dispensers.


rytilahti (approved):
LGTM, thanks for the PR @Alex-ala! :+1:

PR #1641, opened 2023-01-01
Bump dependencies in poetry.lock

rytilahti:
Run poetry update to update all dependencies to correct dependabot-sreported security issues.

Closes #1637
Closes #1624
Closes #1616


rytilahti:
Looks like something has changed regarding how fixtures are handled, causing CI to fail. I cannot reproduce this locally, so some investigation is needed to get this fixed.

PR #1645, opened 2023-01-03
Enable auto-empty settings for roborock Q7 Max+

nijel:


rytilahti (approved):
Thanks! Would you also mind creating a PR to add the features of this model here: https://github.com/marcelrv/XiaomiRobotVacuumProtocol/blob/master/fw_features.md ? When enough features are decoded, we can start leveraging that information instead of hardcoding model information.

nijel:
> Thanks! Would you also mind creating a PR to add the features of this model here: https://github.com/marcelrv/XiaomiRobotVacuumProtocol/blob/master/fw_features.md ? When enough features are decoded, we can start leveraging that information instead of hardcoding model information.

Maybe it would be useful to have get_fw_features somehow exposed in `miiocli`? It took me a while to figure out that it's a protocol command… Anyway, it is in https://github.com/marcelrv/XiaomiRobotVacuumProtocol/pull/71.

nijel:
After looking a bit, I'm uncertain whether it's exposed in the features. In the end Q7 Max and Q7 Max+ seem to differ only in the dock. Maybe `dock_type` in `status` is the place to look for this kind of information.

rytilahti:
Ahh, that could very well be. Oh well, having the feature list available is still a step onward so thanks for doing that!

PR #1646, opened 2023-01-03
roborock: Fix waterflow setting for Q7 Max+

nijel:
It has additional Custom setting (207) and `distance_off` provides more fine-grained setting in this case what is not gracefully handled by this PR, as I will create an issue to discuss this.


rytilahti (changes requested):
Please add a test for this behavior so that we won't break it accidentally in the future.

nijel:
Good idea, I've added the tests.

rytilahti (approved):
LGTM, thanks again!

PR #1685, opened 2023-01-26
Improve roborock update handling

rytilahti:
Not all devices support all features, but we have currently no way of knowing what is supported. In order to allow embedding all supported information in the status container while avoiding making unnecessary I/O on subsequent queries, this introduces a small helper to do just that.

The initial status() call will call all defined devicestatus-returning methods to find out which information is acquired correctly, and skip the unsupported queries in the following update cycles.

This also embeds some more information (last clean details, mop dryer settings).

Ping @starkillerOG

### DIFF
--- a/miio/integrations/roborock/vacuum/tests/test_updatehelper.py
+++ b/miio/integrations/roborock/vacuum/tests/test_updatehelper.py
@@ -1,0 +1,28 @@
+from unittest.mock import MagicMock
+
+from miio import DeviceException
+
+from ..updatehelper import UpdateHelper
+
+
+def test_updatehelper():
+    """Test that update helper removes erroring methods from future updates."""
+    main_status = MagicMock()
+    second_status = MagicMock()
+    unsupported = MagicMock(side_effect=DeviceException("Broken"))
+    helper = UpdateHelper(main_status)
+    helper.add_update_method("working", second_status)
+    helper.add_update_method("broken", unsupported)
+
+    helper.status()
+
+    main_status.assert_called_once()
+    second_status.assert_called_once()
+    unsupported.assert_called_once()
+
+    # perform second update
+    helper.status()
+
+    assert main_status.call_count == 2
+    assert second_status.call_count == 2
+    assert unsupported.call_count == 1
--- a/miio/integrations/roborock/vacuum/tests/test_vacuum.py
+++ b/miio/integrations/roborock/vacuum/tests/test_vacuum.py
@@ -4,9 +4,10 @@ from unittest.mock import patch
 
 import pytest
 
-from miio import RoborockVacuum, UnsupportedFeatureException
-from miio.tests.dummies import DummyDevice
+from miio import DeviceError, RoborockVacuum, UnsupportedFeatureException
+from miio.tests.dummies import DummyDevice, DummyMiIOProtocol
 
+from ..updatehelper import UpdateHelper
 from ..vacuum import (
     ROCKROBO_Q7_MAX,
     ROCKROBO_S7,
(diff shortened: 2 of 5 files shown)


miio/integrations/vacuum/roborock/vacuum.py
```diff
@@ -141,6 +143,33 @@ def __init__(
...
         self._maps: Optional[MapList] = None
         self._map_enum_cache = None
+        self._status_helper = UpdateHelper(self.vacuum_status)
+        self._status_helper.add_update_method("consumables", self.consumable_status)
+        self._status_helper.add_update_method("dnd_status", self.dnd_status)
+        self._status_helper.add_update_method("clean_history", self.clean_history)
+        self._status_helper.add_update_method("last_clean", self.last_clean_details)
+        self._status_helper.add_update_method("mop_dryer", self.mop_dryer_settings)
+
+    def send(
```
starkillerOG, line 153: Why are we overwriting the send method?
We can just add the "unknown_mehtod" response check with raising the exception to the main send method used for all devices right?

starkillerOG (commented):
Looks very good to me, just a small remark about overwriting the send method.

Another point:
The order in which status calls are made could be important, so we schould make sure that the order in which we add the methods is the same order in which the calls are made.
Furthermore in some cases we might want to return multiple status containers from a single method call, see https://github.com/rytilahti/python-miio/pull/1614/files

miio/integrations/vacuum/roborock/vacuum.py
rytilahti: It's overwritten as that response is roborock-specific, the regular errors are reported inside `error` key of the response payload.

On the multiple status containers, we'll have to definitely think how to do that simpler than what the PR you linked does, it feels currently very complex and potentially error-prone.

starkillerOG: Alright, than I think this is good to go.

rytilahti: Great, I'll fix the tests later and get this merged then! I had an idea about potential solution for the multimaps issue(s). How about just adding a back-reference to the child container during embedding? This way the embedded container can define the descriptors while leveraging information from the parent container. Would that solve at least some of the issues you were solving with passing the containers around?

starkillerOG: Yes I think that will solve some of the issues.
But we schould make a method/attribute in order to acces the parent/a diffrent embeded container from a embeded container.

starkillerOG: > Design-wise it would be really clean when one status container corresponds to a response from one method, and while that might not be always doable, I think we should strive towards that where possible.

I can look into that, first I want the caching merged and then I will start working on the MultiMap PR.

starkillerOG: Are you going to add the back-reference to the child container in this PR, or schould I do that in a seperate PR?

starkillerOG: I removed the need to pass around the MapList container to the main status container in this commit: https://github.com/rytilahti/python-miio/pull/1614/commits/9161737b1aa07f49b6548ce02ab8cca4087a3215

If you implement the back-refrence such that a child container can acces the main container and through that also other containers I can also remove the need to pass MapList to the CleaningDetails

rytilahti: Feel free to create a PR to add the back reference. Otherwise I'll take a look into that later this week.

rytilahti: Created #1711

rytilahti:
Hi @starkillerOG, I'll merge this now to unblock you. I changed the signature for embed to require a name (#1712 ), so that accessing the main status and other containers can be done in a nicer manner. Let me know if you encounter any issues!

starkillerOG:
@rytilahti thank you very much, I have used this PR to update the roborock multi map PR here: https://github.com/rytilahti/python-miio/pull/1614

PR #1701, opened 2023-01-29
Cache descriptors on first access

starkillerOG:
The descriptors accessed through `sensors()`, `settings()`, and `actions()` are now cached on the first use to avoid unnecessary I/O.

follow up on PR https://github.com/rytilahti/python-miio/pull/1592 which got merge conflicts and is closed.


starkillerOG:
@rytilahti could you review/merge this?

miio/device.py
```diff
@@ -177,6 +179,63 @@ def _fetch_info(self) -> DeviceInfo:
...
+                raise Exception(
+                    f"Neither setter or setter_name was defined for {setting}"
+                )
+            setting = cast(EnumSettingDescriptor, setting)
+            if (
+                setting.type == SettingType.Enum
+                and setting.choices_attribute is not None
+            ):
+                retrieve_choices_function = getattr(self, setting.choices_attribute)
+                setting.choices = retrieve_choices_function()
```
rytilahti, line 200: Looks like trying to get the settings crashes on my vac, but this was already the case with current master so it can be fixed in a separate PR.
```
DEBUG:miio.miioprotocol:192.168x.x:54321 >>: {'id': 207, 'method': 'get_multi_maps_list', 'params': []}
DEBUG:miio.miioprotocol:192.168x.x:54321 (ts: 2023-01-30 13:09:12, id: 207) << {'id': 207, 'result': 'unknown_method'}
  File "<stdin>", line 1, in <module>
  File "/home/tpr/code/python-miio/miio/device.py", line 349, in sensors
    self._initialize_descriptors()
  File "/home/tpr/code/python-miio/miio/device.py", line 236, in _initialize_descriptors
    self._settings = self._setting_descriptors_from_status(status)
  File "/home/tpr/code/python-miio/miio/device.py", line 201, in _setting_descriptors_from_status
    setting.choices = retrieve_choices_function()
  File "/home/tpr/code/python-miio/miio/integrations/roborock/vacuum/vacuum.py", line 388, in _map_enum
    maps = self.get_maps()
  File "/home/tpr/code/python-miio/miio/click_common.py", line 185, in _wrap
    return func(self, *args, **kwargs)
  File "/home/tpr/code/python-miio/miio/integrations/roborock/vacuum/vacuum.py", line 380, in get_maps
    self._maps = MapList(self.send("get_multi_maps_list")[0])
  File "/home/tpr/code/python-miio/miio/integrations/roborock/vacuum/vacuumcontainers.py", line 112, in __init__
    for map in self.data["map_info"]:
TypeError: string indices must be integers
```

rytilahti (commented):
Looks fairly ready to go to me, please see the comments for some minor issues.

miio/device.py
rytilahti: Created #1702 but there's a another problem in this code path: the attribute access should do no I/O but the update cycle should handle the map list initialization, too.

miio/tests/test_devicestatus.py
```diff
@@ -140,6 +140,7 @@ def level(self) -> int:
             return 1
 
     mocker.patch("miio.Device.send")
+    mocker.patch("miio.Device.send_handshake")
```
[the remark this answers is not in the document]
starkillerOG: No it was needed because the tests otherwise fail, I can't exactly remember why but somehow the device_id is used somewhere, and when that is needed, the send_handshake was called.

Will test some more later tonight.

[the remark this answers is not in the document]
starkillerOG: You were right, the `mocker.patch("miio.Device.send_handshake")` is not needed and I have removed it.
I was missing one patch of the device_id.

miio/device.py
```diff
@@ -271,49 +330,26 @@ def status(self) -> DeviceStatus:
     def actions(self) -> Dict[str, ActionDescriptor]:
         """Return device actions."""
         if self._actions is None:
```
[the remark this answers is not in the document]
starkillerOG: Done, I added a test that checks it is only called once.
It actually executes the _initialize_descriptors call and only patches the underlying _sensor_descriptors_from_status, _setting_descriptors_from_status and _action_descriptors

miio/tests/test_devicestatus.py
```diff
@@ -247,6 +250,24 @@ def level(self) -> TestEnum:
...
+    getter = getattr(d, getter_name)
+    initialize_descriptors = mocker.patch(
+        "miio.Device._initialize_descriptors",
+        side_effect=Device._initialize_descriptors,
+        autospec=True,
+    )
+    mocker.patch("miio.Device.status")
+    mocker.patch("miio.Device._sensor_descriptors_from_status", return_value={})
+    mocker.patch("miio.Device._setting_descriptors_from_status", return_value={})
+    mocker.patch("miio.Device._action_descriptors", return_value={})
```
rytilahti, line 265: These should not be needed as `_initialize_descriptors` is mocked above, and that's the only call-site for these.

starkillerOG: No the mock of _initialize_descriptors has as side_effect itself which makes sure the code of _initialize_descriptors is actually executed. The mock is just used to count the number of calls, but the function is not overwritten.

We need _initialize_descriptors to be executed, otherwise the self._sensors etc. are not set and therefore nothing is cached...

starkillerOG:
@rytilahti the tests are actually succesfull, there is an error uploading the codecov....
That is unrelated to this PR.

miio/tests/test_devicestatus.py
rytilahti: Ah, I see. Instead of defining a side effect, it's better to use a spy there: https://pytest-mock.readthedocs.io/en/latest/usage.html#spy

rytilahti: I just noticed this is in the wrong file, the correct place would be `test_device.py`. Otherwise this looks good to go, I'll merge as soon as tests are passing.

starkillerOG: Thanks for the tip to use spy!
Learning each day :)

I have moved the test to the test_device.py file.

miio/tests/test_device.py
```diff
@@ -182,3 +182,17 @@ def test_supports_miot(mocker):
...
+@pytest.mark.parametrize("getter_name", ["actions", "settings", "sensors"])
+def test_cached_descriptors(getter_name, mocker):
+    d = Device("127.0.0.1", "68ffffffffffffffffffffffffffffff")
+    getter = getattr(d, getter_name)
+    initialize_descriptors = mocker.spy(d, "_initialize_descriptors")
+    mocker.patch("miio.Device.status")
+    mocker.patch("miio.Device._sensor_descriptors_from_status", return_value={})
+    mocker.patch("miio.Device._setting_descriptors_from_status", return_value={})
+    mocker.patch("miio.Device._action_descriptors", return_value={})
+    for _i in range(5):
```
rytilahti, line 196: Just for the future, the proper way to do this is to use plain `_` here. No need to fix that now though.

starkillerOG:
@rytilahti all tests have passed

PR #1711, opened 2023-02-03
Add parent reference to embedded containers

rytilahti:
Allows embedded containers to access data from other embeddeds or the main status.

### DIFF
--- a/miio/devicestatus.py
+++ b/miio/devicestatus.py
@@ -39,6 +39,7 @@ class _StatusMeta(type):
         cls._sensors: Dict[str, SensorDescriptor] = {}
         cls._settings: Dict[str, SettingDescriptor] = {}
 
+        cls._parent: Optional["DeviceStatus"] = None
         cls._embedded: Dict[str, "DeviceStatus"] = {}
 
         descriptor_map = {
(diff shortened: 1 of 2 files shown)


starkillerOG (commented):
Should we not make some changes to __getattr__ of DeviceStatus to make sure a embeded container can acces the master/other embeded properties?

rytilahti:
The embedded container can use `_parent` to access its parent, and then use `_embedded` to find other child containers. We could introduce a helper method to access embedded ones by name if that's what you are looking for?

starkillerOG:
@rytilahti I ment we could make it such that we can use `getattr(self, "MapList__map_name_dict")` for example from within another embeded status container.
And then maybe `getattr(self, "Main__battery")` for example to acces stuff from the main container from within a embeded container.

rytilahti:
You can already access the parent as well as other embedded containers like this:
```
battery = self._parent.battery  # it's better to use attribute access over getattr
consumables = self._parent.consumables

main_brush_used = self._parent.consumables.main_brush
main_brush_used_another_syntax = self._parent.consumables__main_brush  # the one above is preferred
```
Is that not enough for the use case at hand?

starkillerOG:
@rytilahti fair, I think this will surfice, we will do it as you illustrated.
I do think mypy is going to complain about self._parent.consumables not existing, but we will deal with that later.

starkillerOG:
I tested this and it is working. I implemented it in the Roborock multi map PR: https://github.com/rytilahti/python-miio/pull/1614

PR #1724, opened 2023-02-09
Allow defining id for descriptor decorators

rytilahti:
This is necessary to allow non-miot integrations to use custom naming for their descriptors. It is undecided how the standardized properties and actions are to be exposed to downstreams in the end, but one example would be allowing defining light:brightness for light brightness across the board.

### DIFF
--- a/miio/devicestatus.py
+++ b/miio/devicestatus.py
@@ -180,7 +180,9 @@ class DeviceStatus(metaclass=_StatusMeta):
         return getattr(self._embedded[embed], prop)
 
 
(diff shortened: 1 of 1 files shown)


rytilahti:
Ping @starkillerOG just that you are in the loop, too :-) This is related to allow defining standardized naming for common actions & properties used by "specialized" homeassistant platforms, e.g., `vacuum:start`, `vacuum:stop` (or whatever we decide to call them) where special naming is needed (https://developers.home-assistant.io/docs/core/entity/vacuum/) instead of having a generic button/sensor/binary_sensor.

PR #1730, opened 2023-02-10
Add roborock mop washing actions

starkillerOG:


rytilahti (approved):
LGTM, thanks @starkillerOG! The docs CI was broken by my earlier changes which I need to fix later.

PR #1740, opened 2023-02-19
Ignore parser import error for doc8

rytilahti:


rytilahti:
Already added in #1755.

PR #1753, opened 2023-03-05
Generalize settings and sensors into properties

rytilahti:
This removes the unnecessary division between sensors and settings in favor of just having "properties" that can have different access flags.

This will greatly simplify the API as all properties are alike, the difference is just in the access flags.

The earlier API to get settable properties (settings) and readable properties (sensors) is kept intact for the time being.

* SettingDescriptor is no more, all readable and writable properties are described using PropertyDescriptors:
	* SettingType is replaced with PropertyConstraint to allow signaling allowed ranges or choices.
	* EnumSettingDescriptor is now EnumDescriptor
	* NumberSettingDescriptor is now RangeDescriptor

* Add 'access' to Descriptor
	* This will also allow for generic `descriptors` interface in the future, if needed.

* Add 'property' to Descriptor
	* None for actions, allwos more generic interface for properties.

* Add 'properties' method to 'Device' to get all property descriptors. 'settings' and 'sensors' return a filtered dict based on the properties for backwards compat.

### DIFF
--- a/miio/descriptors.py
+++ b/miio/descriptors.py
@@ -2,15 +2,13 @@
 
 The descriptors contain information that can be used to provide generic, dynamic user-interfaces.
 
-If you are a downstream developer, use :func:`~miio.device.Device.sensors()`,
-:func:`~miio.device.Device.settings()`, and
+If you are a downstream developer, use :func:`~miio.device.Device.properties()`,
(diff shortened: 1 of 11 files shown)


rytilahti:
Windows builds are failing due to a dependency issue, which I'll ignore for now to unblock further development.

PR #1755, opened 2023-03-05
Update dependencies and pre-commit hooks

rytilahti:
* Update pre-commit hooks and run them against the whole codebase
* Relax development dependencies where there is no need for specific version
* Update locked dependencies


rytilahti:
Windows builds are broken due to a bug in poetry: https://github.com/python-poetry/poetry/issues/7505 https://github.com/python-zeroconf/python-zeroconf/issues/1125

Issue #1034, opened 2021-05-06
Add "cycles" option from Roborock S7 and older models


Dielee:
**Is your feature request related to a problem? Please describe.**
I have already tried with wireshark and bluestacks to sniff the new function of "cycles". Unfortunately no UDP packets arrive on the miio port 54321. Maybe someone can help ? 

**Describe the solution you'd like**
I would like to implement the "cycles" option form newer roborock vaccums into python-miio.I can't manage to capture the traffic of the miio protocol.

**Device information:**
If the enhancement is device-specific, please include also the following information.

  - Name(s) of the device: Roborock S7 (roborock.vacuum.a15) and maybe some more

rytilahti:
What does the cycles do? If you execute `miiocli vacuum --help` it will print you a list of already supported features (some of which may not be implemented in homeassistant, if you are using it).

I'm not sure why you are not seeing related packets using wireshark or bluestacks, so I cannot really help you there, but it seems to indicate that the communication is not done locally but over the cloud..

Dielee:
Cycles cleans the same a room for X times. This is very useful for mopping. 

This feature looks like this:
![Screenshot_20210507-095252__01.jpg](https://user-images.githubusercontent.com/45101649/117417148-55e4ca80-af1a-11eb-8c14-572eef4e5bd4.jpg)

Maybe communication ist working over cloud, but python-miio works local.
S7 works with python-miio, so there has to be a way to start this feature locally.
@fettlaus implemented the S7 integration, maybe he can help ? 

Thank you!

Dielee:
Looks like S7 is using the new miotspec protocol, which does not work locally, but supports the "old" miio protocol also.
Maybe I will take a look into the https connection with miotspec tomorrow.

fettlaus:
To be fair: all I did was some trial and error. I didn't even need to sniff network traffic.

fettlaus:
I currently don't want to start my vacuum, but please try something like this: `miiocli vacuum --ip 192.168.x.x --token xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx segment_clean '[{"segments":[18],"repeat":2}]`
If it works there is no code change necessary maybe only a hint in the docs. (got the code from marcelrv/XiaomiRobotVacuumProtocol/issues/44)

rytilahti:
[...]

PR #1771, opened 2023-05-03
Add `repeat` param to Roborock segment clean

MrBartusek:
Hey! This is actuall fix for #1034, for some reason, this issue was closed with sketchy workarounds.

I've added `repeat` prop for `segment_clean` function in `RoborockVacuum`. This represents the count of repeats that the vacuum does in specified segments:

image

This is done as specified in [XiaomiRobotVacuumProtocol](https://github.com/marcelrv/XiaomiRobotVacuumProtocol/blob/master/segment_clean.md). I've tested this with my Roborock S5 Max and it works well.

### DIFF
--- a/miio/integrations/roborock/vacuum/vacuum.py
+++ b/miio/integrations/roborock/vacuum/vacuum.py
@@ -891,12 +891,16 @@ class RoborockVacuum(Device):
         return self.send("resume_segment_clean")
 
     @command(click.argument("segments", type=LiteralParamType(), required=True))
-    def segment_clean(self, segments: List):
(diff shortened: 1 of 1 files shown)


rytilahti (approved):
Thanks for the PR, I think this is good to go (the linting checks are unrelated)!

The earlier format might be what earlier firmware versions used, but I think this is fine to merge. If someone is using this and we get bug reports about incorrect behavior, we can adjust the payload based on the version information.

MrBartusek:
Thanks for the merge! I will probably be also adding support for that in Home Assistant after the next release.

PR #1784, opened 2023-06-16
Add support for Dmaker.fan.p39

Landrash:
Adds support for fanmiot p39. Already supported by genericmiot but does not expose mode.
Compared to P33 there's a few changes.

### Changes
Added to README and to fan_miot.py

### Source 
https://miot-spec.org/miot-spec-v2/instance?type=urn:miot-spec-v2:device:fan:0000A005:dmaker-p39:1

### Function test
Tested each command and function with `miiocli`
```  miiocli fanmiot --ip 10.20.30.50 --token 000000000000000000000000 COMMAND ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ integration, this library supports also the following devices:
 * Xiaomi Philips Zhirui Bedroom Smart Lamp
 * Huayi Huizuo Lamps
 * Xiaomi Universal IR Remote Controller (Chuangmi IR)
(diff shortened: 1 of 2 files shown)


rytilahti:
Hi @Landrash and thanks for the PR! Which mode is not supported (siid and piid?) when using the genericmiot? Looking at the linked spec file, the mode seems to be there.

Landrash:
I'm incorrect.
Sorry about that.

I'll close this. Should I make another pr just to the Readme adding it being supported by the genericmiiot?

rytilahti:
Ah, thanks for letting me know! No need to add it to the readme as the genericmiot should cover all modern devices :-)

PR #1787, opened 2023-06-23
Set zhimi.fan.za4 countdown timer to minutes

### LINKED ISSUE #1786: Countdown timer for fan zhimi.fan.za4 should be in minutes
**Describe the bug**
As mentioned in home-assistant/core#91824 the countdown timer for `zhimi.fan.za4` is currently in seconds, but this should be in minutes.

**Version information (please complete the following information):**
 - OS: Docker (python:3 image)
 - python-miio: 0.5.12

**Device information:**
If the issue is specific to a device [Use `miiocli device --ip <ip address> --token <token> info`]:
  - Model: zhimi.fan.za4
  - Hardware version: esp32
  - Firmware version: 2.0.3

**To Reproduce**
1. Set the countdown timer
2. See that the countdown timer should be in minutes, but is in seconds

**Expected behavior**
The countdown timer should be set in minutes and retrieved in minutes

Maybe an exception should be made here?:

https://github.com/rytilahti/python-miio/blob/ecdabcd963d77526a84b0a3df4023595de22a657/miio/integrations/fan/zhimi/fan.py#L171-L174

alex3305:
This change fixes #1786.

There may be some minor cosmetic tweaks needed for the CLI. For instance, setting the timer still shows 'seconds':

```
❯ miiocli fan --ip 192.168.10.8 --token REDACTED delay_off 600
Setting delayed turn off to 600 seconds
['ok']
```

while actually it is in minutes 🤷. But apparently I'm not profound enough in Python to fix this. So any help for this is welcome!

I've added a (working) test and also tested it with my device and that seem to work great. So I was pretty content with myself 😄.

---

As you can see I've created a simple extension upon the base `FanStatus` class. I thought this was the most efficient way to implement this (very small) change for a specific device. I've not yet added the decorators [per your suggestion](https://github.com/rytilahti/python-miio/issues/1786#issuecomment-1603360862) as I wanted to get the basic stuff out of the way first. Maybe another extension is needed for those to be correct everywhere, but I want to keep the tasks / PR small, if you don't mind 😉.

Thanks again for the great library 🚀!


syssi:
Thanks for your contribution! AFAIK the issue affects a bunch of fan models. I will try to provide some more input as soon as possible (probably next week). May be we are able to provide a more generic solution to provide a consistent interface.

alex3305:
@syssi Thanks for your reply and input. Sounds good to me, although I hope this issue doesn't lose it's scope. The countdown timer is now unusable for this device (from HA) and a fix would be very welcome.

syssi:
Trust me. In best case we are able to reduce your diff to 2-3 lines of code. :-)

alex3305:
😢 Any updates on this? This functionality is broken for over 3 months now. Which terribly influenced my HA WAF. I hoped it could be done before summer ended, but now I would be happy to have implemented it at all.

_pinging @syssi @rytilahti_

alex3305:
Still broken 

@rytilahti

rytilahti (approved):
LGTM and thanks for the PR! I'll get this merged right now so that it will be part of the dev release I am currently preparing.

PR #1793, opened 2023-07-14
Fix flake8 SIM910 errors and add pin pydantic==^1

rytilahti:


rytilahti:
Looks like the builds were failing on some win & mac combinations with the same error as discussed in https://github.com/python-poetry/poetry/issues/7611 - also, the docs build was failing due as the code base is not pydantic v2 compatible at the moment, so this adds a pin to pydantic==^1.

PR #1809, opened 2023-08-20
Replace datetime.utcnow + datetime.utcfromtimestamp

cdce8p:
Starting with Python 3.12, `datetime.utcnow` and `datetime.utcfromtimestamp` will be deprecated.

https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcnow
https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcfromtimestamp

### DIFF
--- a/miio/miioprotocol.py
+++ b/miio/miioprotocol.py
@@ -7,7 +7,7 @@ import binascii
 import codecs
 import logging
 import socket
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from pprint import pformat as pf
 from typing import Any, Dict, List, Optional
 
@@ -48,7 +48,7 @@ class MiIOProtocol:
 
         self._discovered = False
         # these come from the device, but we initialize them here to make mypy happy
-        self._device_ts: datetime = datetime.utcnow()
+        self._device_ts: datetime = datetime.now(tz=timezone.utc).replace(tzinfo=None)
         self._device_id = b""
 
     def send_handshake(self, *, retry_count=3) -> Message:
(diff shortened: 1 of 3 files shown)


miio/miioprotocol.py
```diff
@@ -48,7 +48,7 @@ def __init__(
 
         self._discovered = False
         # these come from the device, but we initialize them here to make mypy happy
-        self._device_ts: datetime = datetime.utcnow()
+        self._device_ts: datetime = datetime.now(tz=timezone.utc).replace(tzinfo=None)
```
rytilahti, line 51: Thanks for the PR, @cdce8p! I'm wondering if it's necessary to have `replace()` calls at all? Alas, I'm not going to have access to any test devices until next month, so I cannot really test this myself.

The CI failures are irrelevant to this PR, so they are not blockers to get this merged.

cdce8p: I can't test it myself either, unfortunately.

However, I believe you might be correct here. The call in `_valid_cache` is covered by the test suite. As for the protocol: AFAICT the conversion happens inside the `TimeAdapter`. `obj.timetuple()` strips the timezone anyway and for the `decode` we might as well leave the `tz` in place.

cdce8p:
@rytilahti Gentle ping. Anything left todo here for me?

rytilahti:
@cdce8p thanks for the ping and sorry for the delay. I think it's good to go, but I haven't been still able to test it on any real devices. But well, we can always revert it if needed, so I just updated on top of the current master in hopes it fixes the CI issues and I'll merge this when the checks pass.

cdce8p:
> so I just updated on top of the current master in hopes it fixes the CI issues and I'll merge this when the checks pass.

FYI the windows error seems to be a poetry issue. Downgrading to `poetry<1.4.0` might help.
https://github.com/python-poetry/poetry/issues/7611

PR #1827, opened 2023-09-18
Added support for dreame d10 plus

TxMat:
Dreame D10 plus (https://home.miot-spec.com/spec/dreame.vacuum.r2205) seems to be using the same mapping as DREAME_TROUVER_FINDER. This allows for miio to use the robot properly by supporting it with dreamevacuum instead of genericmiot


rytilahti (approved):
LGTM, thanks @TxMat!

PR #1832, opened 2023-09-24
Don't log error message when decoding valid discovery packets

### LINKED ISSUE #1824: KeyError('token') in EncryptionAdapter
**Describe the bug**
Library prints `Unable to decrypt, returning raw bytes` message for Hello packets due to missing context. This scary text confuses people when they debug unrelated issues. (yes, the issue title is a clickbait)

**Details**
* In `MiIOProtocol::discover` at https://github.com/rytilahti/python-miio/blob/f7b554f49fac028d858e6f6e573328c153d1b687/miio/miioprotocol.py#L121 we parse the Hello `Message` without giving context `token`
* In `EncryptionAdapter::_decode` at https://github.com/rytilahti/python-miio/blob/f7b554f49fac028d858e6f6e573328c153d1b687/miio/protocol.py#L165 `token` is accessed from the context. It doesn't exists in the context and this throws `KeyError('token')`
* The error handling thinks this is a decoding error and prints a scary warning at https://github.com/rytilahti/python-miio/blob/f7b554f49fac028d858e6f6e573328c153d1b687/miio/protocol.py#L168

**Proposed fix**
Since the hello message is parsed differently in any case (see `Utils.is_hello`), we could separate the message types as:
```
Message = Struct(
    # for building we need data before anything else.
    "data" / Pointer(32, RawCopy(EncryptionAdapter(GreedyBytes))),
    "header"
    / RawCopy(
        Struct(
            Const(0x2131, Int16ub),
            "length" / Rebuild(Int16ub, Utils.get_length),
            "unknown" / Default(Int32ub, 0x00000000),
            "device_id" / Hex(Bytes(4)),
            "ts"
            / TimeAdapter(
                Default(
                    Int32ub,
                    datetime.datetime.now(tz=datetime.timezone.utc),
                )
            ),
        )
    ),
    "checksum" / Checksum(Bytes(16), Utils.md5, Utils.checksum_field_bytes),
)

DiscoverMessage = Struct(
    "header"
    / RawCopy(
        Struct(
            Const(0x2131, Int16ub),
            "length" / Const(0x20, Int16ub),
            "unknown" / Default(Int32ub, 0x00000000),
            "device_id" / Hex(Bytes(4)),
            "ts"
            / TimeAdapter(
                Default(
                    Int32ub,
                    datetime.datetime.now(tz=datetime.timezone.utc),
                )
            ),
        )
    ),
    "token" / Bytes(16),
)
```

Additionally.
* In `_discover` , we'd then use  `DiscoverMessage` instad of `Message`
* We remove `Utils.is_hello`
* We replace `m.checksum` when accessing the token with `m.token` in `miioprotocol.py`

**Other**
I can also do a PR if that's prefereable. It'll take me a few moments to set up the environment first though.

gunjambi:
Fix `Unable to decrypt, returning raw bytes` warning message when decoding Discovery messages. Fixes #1824

I opted for a quick surgical fix here rather than rework the whole message parsing. While separating the Discovery messages from normal data messages is quite easy in the context of miioprotocol, there's a quite bit of uses of the `Message` in other places too. I didn't feel confident in changing those.

Additionally, it doesn't even seem desirable to separate Discovery-Ack and Data messages. It seems useful in some cases to be able to parse any message without knowing ahead of time what kind of data it might contain. (I doesn't really matter right now as the token, i.e. decryption key, must be supplied ahead of time, but it is conceivable that a parser might first match the ID of the reply to the request and then parse the contents. )

Fixes #1824

### DIFF
--- a/miio/protocol.py
+++ b/miio/protocol.py
@@ -161,6 +161,11 @@ class EncryptionAdapter(Adapter):
 
     def _decode(self, obj, context, path) -> Union[Dict, bytes]:
         """Decrypts the payload using the token stored in the context."""
+        # if there is no payload, decode to 0 bytes. Missing payload is expected for discovery messages.
+        # note that we don't have "token" in the context for discovery replies so we couldn't decode it
+        # anyway.
+        if obj == b"":
+            return b""
         try:
             decrypted = Utils.decrypt(obj, context["_"]["token"])
             decrypted = decrypted.rstrip(b"\x00")
--- a/miio/protocol.py
+++ b/miio/protocol.py
@@ -161,11 +161,9 @@ class EncryptionAdapter(Adapter):
 
     def _decode(self, obj, context, path) -> Union[Dict, bytes]:
         """Decrypts the payload using the token stored in the context."""
-        # if there is no payload, decode to 0 bytes. Missing payload is expected for discovery messages.
-        # note that we don't have "token" in the context for discovery replies so we couldn't decode it
-        # anyway.
-        if obj == b"":
-            return b""
+        # Missing payload is expected for discovery messages.
+        if not obj:
+            return obj
         try:
             decrypted = Utils.decrypt(obj, context["_"]["token"])
             decrypted = decrypted.rstrip(b"\x00")


miio/protocol.py
```diff
@@ -161,6 +161,11 @@ def _encode(self, obj, context, path):
 
     def _decode(self, obj, context, path) -> Union[Dict, bytes]:
         """Decrypts the payload using the token stored in the context."""
+        # if there is no payload, decode to 0 bytes. Missing payload is expected for discovery messages.
+        # note that we don't have "token" in the context for discovery replies so we couldn't decode it
+        # anyway.
```
rytilahti, line 166:
```suggestion
        # Missing payload is expected for discovery messages.
```
let's keep it simple here, also, there are some (very) old devices that actually respond with a valid token :-)

rytilahti (changes requested):
A couple of simple changes and this is good to go, thanks! :+1:

gunjambi:
Proposed changes applied.

rytilahti (approved):
Great, thanks again! :+1:

PR #1845, opened 2023-10-20
Introduce common interfaces based on device descriptors

### LINKED ISSUE #1692: genericmiot: handle duplicate types
It is possible that two siids/services have the same "name" component in the urn, for example, the schema for `yeelink.switch.sw1` defines these two sharing the same urn:
```
            "iid": 2,
            "type": "urn:miot-spec-v2:service:switch:0000780C:yeelink-sw1:1:0000C809",
            "description": "First Switch Service",
```
and 
```
            "iid": 3,
            "type": "urn:miot-spec-v2:service:switch:0000780C:yeelink-sw1:1:0000C809",
            "description": "Second Switch Service",
```
that have same properties. As the `id` for properties is currently the combination of `<name>` (here `switch`), the latter overwrites the former.

This is similar to the actions (e.g., reset filter on some vacuums) which should also be handled somehow, maybe just adding a suffix?

### LINKED ISSUE #1814: Implement common API for sensors, settings and actions
Having a stable API for `Device` to obtain sensors, settings, and actions.
The cli tool should use the descriptors where available to provide the device status.

rytilahti:
This completes the common descriptor-based API for all device-inherited classes:
- status() -> DeviceStatus: returns device status
- descriptors() -> DescriptorCollection[Descriptor]: returns all defined descriptors
- actions() -> DescriptorCollection[ActionDescriptor]: returns all defined actions
- settings() -> DescriptorCollection[PropertyDescriptor]: returns all settable descriptors
- sensors() -> DescriptorCollection[PropertyDescriptor]: returns all read-only descriptors
- call_action(name, params): to call action using its name
- change_setting(name, params): to change a setting using its name

These functionalities are also provided as cli commands for all devices:
- status
- descriptors
- actions
- settings
- sensors
- call (call_action)
- set (change_setting)

TODO:
- [x] Add tests for DescriptorCollection
- [ ] Add docs

Fixes #1814
Fixes #1692


rytilahti:
The docstrings shall do fine for now so I'll just merge this to allow people test it, proper docs for library usage can be added in a separate PR.

PR #1853, opened 2023-10-21
Add support for dmaker.fan.p45

saxel:


rytilahti (approved):
LGTM, btw, the genericmiot should be able to control this device too. Have you tried it out in the current git master?

saxel:
Yes, the genericmiot seems to be able to control the fan just fine on master branch.

rytilahti:
Great, the thing with genericmiot is that while it offers an universal support, it requires online access so having something functioning as-is can be useful. There may also be existing users of fanimiot, so extending the support for those sounds fair, so let's get this merged for the upcoming release. Thanks for the PR! :+1:

PR #1868, opened 2023-12-03
Added support for Xiaomi Smart Space Heater 1S (zhimi.heater.mc2a)

paranerd:
This PR adds support for Xiaomi Smart Space Heater 1S (zhimi.heater.mc2a).
Changes are minimal as the mc2a is basically identical to the mc2.

I own the device myself and can confirm that everything reads and performs as expected.

I would appreciate a quick merge as this is only the first step on my journey to natively integrate both devices into the main "Xiaomi" integration for Home Assistant.

Thanks a lot for the amazing foundation!

### DIFF
--- a/miio/integrations/dmaker/fan/fan_miot.py
+++ b/miio/integrations/dmaker/fan/fan_miot.py
@@ -24,6 +24,7 @@ MODEL_FAN_P11 = "dmaker.fan.p11"
 MODEL_FAN_P15 = "dmaker.fan.p15"
 MODEL_FAN_P18 = "dmaker.fan.p18"
 MODEL_FAN_P33 = "dmaker.fan.p33"
+MODEL_FAN_P39 = "dmaker.fan.p39"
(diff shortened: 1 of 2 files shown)


paranerd:
Getting a linting error "`miio/integrations/dmaker/fan/fan_miot.py:314: error: Incompatible types in assignment (expression has type "dict[str, object]", base class "MiotDevice" defined the type as "dict[str, dict[str, dict[str, Any]]]")  [assignment]`"

The error itself makes sense to me, because the types indeed seem incompatible. However I didn't edit that specific line, so not sure how to fix it.

rytilahti (changes requested):
Ohh, please split the PR into two, one per each device.

paranerd:
I removed the p39 code and merged the recent changes into my branch but for some reason now `macos-latest` and `windows-latest` CI checks seem to be randomly failing - any ideas?

rytilahti (approved):
Those are known to be flaky, no worries on that :-)

PR #1877, opened 2023-12-05
Added support for Xiaomi Tower Fan (dmaker.fan.p39)

paranerd:
This PR adds support for the Xiaomi Tower Fan (dmaker.fan.p39).
Changes are minimal as the p39 is quite similar to p33.

I own the device myself and can confirm that everything reads and performs as expected.

I would appreciate a quick merge as this is only the first step on my journey to natively integrate both devices into the main "Xiaomi" integration for Home Assistant.

Thanks a lot for the amazing foundation!


rytilahti (approved):
LGTM, thanks! :+1:

PR #1885, opened 2024-01-09
Added initial zhimi.humidifier.ca6 support

### LINKED ISSUE #1884: Add support for Smartmi Evaporative Humidifier 3 (zhimi.humidifier.ca6) support to xiaomi_miio
**Is your feature request related to a problem? Please describe.**

python-miio and HA currently support Smartmi Evaporative Humidifier 2 (```zhimi.humidifer.ca4```).

Please add support for [Smartmi Evaporative Humidifier 3](https://eu.smartmiglobal.com/pages/smartmi-evaporative-humidifier-3) (```zhimi.humidifier.ca6```). It is similar to ca4. It even works if one manually specifies the ca4 model in the configation. But there substantial differences:

* Modes are different, i.e. there are three modes: Fav (aka Max), Auto, Sleep.
* Water level meaning is different, i.e. 0 - min/empty, 1 - normal, 2 - max/full.
* Target humidity max level is 60%, instead of 80%.

**Describe the solution you'd like**

Integration similar to ```zhimi.humidifier.ca4```

**Device information:**
If the enhancement is device-specific, please include also the following information.

  - Name(s) of the device: Smartmi Evaporative Humidifier 3
  - Link: https://eu.smartmiglobal.com/pages/smartmi-evaporative-humidifier-3
  - Model: zhimi.humidifier.ca4
  - Hardware version: esp32
  - Firmware version: 2.2.6_0043

smmoroz:
Adds support for zhimi.humidifier.ca6 and status container descriptors for airhumidifier_miot.
Closes #1884.


smmoroz:
Dear @rytilahti and @syssi , first of all, I have to say that I'm not a professional software developer for a long time. Thus please excuse me for my way of working with GitHub (branching, PR, etc.) 

In addition to this PR I have implemented initial support of  ```zhimi.humidifier.ca6``` in home-assistant/core in this [branch](https://github.com/smmoroz/core/tree/xiaomi_miio_zhimi_humidifier_ca6). All ```zhimi.humidifier.ca6``` features work well. Below are a couple of screenshots.

But I can't make PR to home-assistant/core because of the [```manifest.json```](https://github.com/smmoroz/core/blob/xiaomi_miio_zhimi_humidifier_ca6/homeassistant/components/xiaomi_miio/manifest.json). I have to use custom requirements (see below), which I can't commit to home-assistant/core repo:

```
  "requirements": ["construct==2.10.68", "micloud==0.5", "git+https://github.com/smmoroz/python-miio.git#python-miio==0.6.0"],
```

So, I don't know how to proceed. Maybe you can backport my changes and release ```0.5.13``` of python-miio?

Here are a few screenshots from HA:

CleanShot 2024-01-13 at 11 12 11@2x
CleanShot 2024-01-13 at 11 13 46@2x

rytilahti (commented):
Sorry for the late reply, but instead of creating a new file to maintain, it would be better to either use `genericmiot` or adapt the `airhumidifer_miot` as necessary.

smmoroz:
> Sorry for the late reply, but instead of creating a new file to maintain, it would be better to either use `genericmiot` or adapt the `airhumidifer_miot` as necessary.

I moved CA6 support to `airhumidifer_miot` file. Please review. 

I also noticed that in order to support `status` command of `miiocli,` I need to implement `@sensor` and `@setting`. This is not yet done. BTW, `status` doesn't work even for CA4 devices.

smmoroz:
I added `@sensor` and `@setting` to `AirHumidifierMiot` and its subclasses. But I'm not sure I did it in the most elegant way, because I had to use `embed` method in subclasses.

tomaszzmuda:
+1

TB-SE:
+1

miio/integrations/zhimi/humidifier/airhumidifier_miot.py
```diff
@@ -6,14 +6,16 @@
...
 
 _LOGGER = logging.getLogger(__name__)
 
 
 SMARTMI_EVAPORATIVE_HUMIDIFIER_2 = "zhimi.humidifier.ca4"
+SMARTMI_EVAPORATIVE_HUMIDIFIER_3 = "zhimi.humidifier.ca6"
 
 
-_MAPPINGS = {
+_MAPPINGS_CA4 = {
```
rytilahti, line 18: It would be better to have the new mappings contained in the very same `_MAPPINGS` as then it will be automatically used based on the model information. Could you change that?

miio/integrations/zhimi/humidifier/airhumidifier_miot.py
```diff
@@ -381,3 +540,238 @@ def set_dry(self, dry: bool):
...
+    @sensor(
+        "Pump Cnt",
+        entity_category="diagnostic",
+    )
+    def pump_cnt(self) -> int:
+        """Return pump-cnt."""
+        return self.data["pump_cnt"]
+
+
+class AirHumidifierMiotCA6(MiotDevice):
```
rytilahti, line 660: Are the differences so large between the models, that the implementation for CA6 couldn't fit in the same implementation class with if-guarding based on the model?

rytilahti:
Looks much better, thanks. I'm wondering, are these devices so different that they wouldn't fit in the same implementation class?

miio/integrations/zhimi/humidifier/airhumidifier_miot.py
smmoroz: I tried to have all mapping in one `_MAPPING`, but in this case I didn't manage to make the whole thing work. I still don't understand why. I'm not a professional software developer anymore ;)

miio/integrations/zhimi/humidifier/airhumidifier_miot.py
smmoroz: I think, that the differences are big. CA6 introduced new properties, like `pump_cnt`, and changed behaviour of some other props. I already introduced some common thing in classes like `AirHumidifierMiotCommonStatus`. Moreover, I'm pretty sure that the manufacturer will not stop added/changing things in new models. Thus, it's better to be prepared. If-guarding in this case would not be convenient.

miio/integrations/zhimi/humidifier/airhumidifier_miot.py
rytilahti: Okay, great, let's get this merged then! :-)

rytilahti (approved):
LGTM, this can be revisited and polished if need to be, thanks @smmoroz! :+1:

PR #1914, opened 2024-03-17
Roborock: various improvements to exposed sensors

SLaks:
* Add `clean_percent`
* Expose `options`, `device_class`, `suggested_display_precision` on specific sensors for better homeassistant integration
* Also, add q revo to the list of devices that have a mop

### DIFF
--- a/miio/integrations/roborock/vacuum/vacuum.py
+++ b/miio/integrations/roborock/vacuum/vacuum.py
@@ -127,6 +127,8 @@ AUTO_EMPTY_MODELS = [
     ROCKROBO_Q_REVO,
 ]
 
+MODELS_WITH_MOP = [ROCKROBO_S7, ROCKROBO_S7_MAXV, ROCKROBO_Q_REVO]
+
 
 class RoborockVacuum(Device):
     """Main class for roborock vacuums (roborock.vacuum.*)."""
@@ -981,7 +983,7 @@ class RoborockVacuum(Device):
     @command()
     def mop_intensity(self) -> MopIntensity:
         """Get mop scrub intensity setting."""
-        if self.model not in [ROCKROBO_S7, ROCKROBO_S7_MAXV]:
+        if self.model not in MODELS_WITH_MOP:
             raise UnsupportedFeatureException(
                 "Mop scrub intensity not supported by %s", self.model
             )
@@ -991,7 +993,7 @@ class RoborockVacuum(Device):
     @command(click.argument("mop_intensity", type=EnumType(MopIntensity)))
     def set_mop_intensity(self, mop_intensity: MopIntensity):
         """Set mop scrub intensity setting."""
-        if self.model not in [ROCKROBO_S7, ROCKROBO_S7_MAXV]:
+        if self.model not in MODELS_WITH_MOP:
             raise UnsupportedFeatureException(
                 "Mop scrub intensity not supported by %s", self.model
             )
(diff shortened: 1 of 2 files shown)


rytilahti (commented):
LGTM, thanks for the PR, @SLaks! Did you mark this as a draft on purpose, or should this be merged right away?

SLaks:
Update: I verified that the map dropdown works correctly.  However, it resets to the first map in the list (which, incidentally, is the map the robot is physically in) after 10 seconds or so.

miio/integrations/roborock/vacuum/vacuumcontainers.py
```diff
@@ -185,7 +186,12 @@ def state_code(self) -> int:
         return int(self.data["state"])
 
     @property
-    @sensor("State", entity_category="diagnostic")
+    @sensor(
+        "State",
+        device_class="enum",
+        entity_category="diagnostic",
+        options=list(STATE_CODE_TO_STRING.values()),
```
rytilahti, line 193: Cool, didn't know about this! :-)

rytilahti (approved):
Looks good to me, tested on gen1 vac. We can either merge this and continue on subsequent PRs, or leave this open for the time being, your call.

SLaks:
I pushed a WiFi strength sensor.

This is everything I wanted to add at the moment; I may send followups if I discover more issues

miio/device.py
```diff
@@ -119,7 +119,7 @@ def raw_command(self, command, parameters):
     @command(
         skip_autodetect=True,
     )
-    def info(self, *, skip_cache=False) -> DeviceInfo:
+    def info(self, skip_cache=False) -> DeviceInfo:
```
rytilahti, line 122:
```suggestion
    def info(self, *, skip_cache=False) -> DeviceInfo:
```
The kwarg-only parameter definition is intentional, as it forces one to type `info(skip_cache=True)` to make it unambiguous what is really wanted.

miio/integrations/roborock/vacuum/vacuum.py
```diff
@@ -153,6 +156,7 @@ def __init__(
         self._map_enum_cache = None
         self._status_helper = UpdateHelper(self.vacuum_status)
         self._status_helper.add_update_method("consumables", self.consumable_status)
+        self._status_helper.add_update_method("connectivity", self.connectivity_status)
```
rytilahti, line 159: Please move this into its own PR, as that will require some more thought and the rest of the changes are good to go.

One way would be to move the "helper" and exposing common sensors like wifi signal in the base class. I'm a bit hesitant on adding more I/O to the main update cycle for two reasons: (1) not all devices support this (none of I have does, but they are _old_ so maybe all newer ones do) and (2) devices that are disconnected from the cloud are known to fail to respond to info properly.

It could be that all of this is irrelevant with modern devices and firmware versions, but I'd like to avoid potentially breaking working setups, including my own, and have this in a separate PR to allow merging this one.

miio/device.py
SLaks: TIL; thanks!

Done in separate PR

miio/integrations/roborock/vacuum/vacuum.py
SLaks: Done

rytilahti:
Thanks! I updated the title & description to be a bit more descriptive for the git logs, I hope that's fine with you.

SLaks:
Of course

Thanks!

PR #1915, opened 2024-03-18
Linking poetry tool on getting started section of documentation

diosmiodio:
For folks who aren't super familiar with python, might save them some time down the road. Saw it after I checked out the contribution guidelines, but I figure most people will start with getting started :)


rytilahti (approved):
Thanks @diosmiodio! :+1:

PR #1918, opened 2024-03-20
Roborock: Add WiFi signal strength sensor

SLaks:
Re: https://github.com/rytilahti/python-miio/pull/1914#discussion_r1531240944

All I have is Q Revo and S7 (both cloud connected); I don't know about other devices.

Options include:

 - Store prior `info()` failures and skip this if it fails
 - Only call `add_update_method(connectivity_status)` if the WiFi strength sensor is enabled in HA (requires a more complex refactor)
 - Move this sensor to a separate feature in the base class, independent of status classes and everything else (would decouple async work, and may make it easier to implement the previous bullet)

### DIFF
--- a/miio/integrations/roborock/vacuum/vacuum.py
+++ b/miio/integrations/roborock/vacuum/vacuum.py
@@ -47,6 +47,7 @@ from .vacuumcontainers import (
     CarpetModeStatus,
     CleaningDetails,
     CleaningSummary,
+    ConnectivityStatus,
     ConsumableStatus,
     DNDStatus,
     MapList,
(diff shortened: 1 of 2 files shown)


SLaks:
Update: As I ran the vacuum, this sensor never changed.

I'm not sure whether that means my implementation is incorrectly caching, or whether the vacuum reports stale data.

SLaks:
I tried using https://github.com/marcelrv/XiaomiRobotVacuumProtocol/blob/master/network_info.md instead, and watched it print the same value in the terminal as the vacuum moved halfway across the house.

Either my WiFi coverage is literally perfect (which seems very unlikely) or the vacuum doesn't actually make this possible.

Confusingly, my S7 reports 0, even after rebooting somewhere else.

PR #1922, opened 2024-03-31
fix json decode quirk for xiaomi e10

matan-h:
Version:
Firmware version: `2.2.4_0050`
Hardware version: `esp32`

some commands (e.g. `genericmiot call map:rename-map '[0,"string"]'`) generate json-like string like this:
```json
{"id":2,"result":,"exe_time":0}
```
and this produce the error:
```log
ERROR:miio.protocol:Unable to parse json 'b'{"id":2,"result":,"exe_time":0}'': Expecting value: line 1 column 18 (char 17)
ERROR:miio.click_common:Exception: Unable to parse message payload
```
I fix that by removing the nonsense `result":,`.


rytilahti:
Is the command still working, but just returning invalid json? Otherwise looks fine for me, thanks for the PR!

matan-h:
@rytilahti this is an interesting story. e10 is a cheap model that don't support maps. However, it still has the maps commands, just in a way no one can access them: there is only one map called "hello" with index 0, and according to the spec, map indexes start from one. So this corrupted json is the robot confused and returning empty result. (probably something like `f"results:get_results(),exec_time:get_time()"`)

rytilahti:
Heh, yeah, some firmware behave weirdly on error states, likely for the reason you mention. Would you mind adding a unit test similar as https://github.com/rytilahti/python-miio/blob/master/miio/tests/test_protocol.py#L129 for this quirk?

rytilahti (approved):
Thanks for the PR, @matan-h! I added a quick test and will merge this after the CI passes.

PR #1924, opened 2024-04-02
Update dreamevacuum_miot.py with Xaiomi X10

MaciekIzdebski:
Added support for Xaiomi/dreame Vaccum X10/X10+


rytilahti (commented):
Hi and thanks for the PR @MaciekIzdebski! A couple of quick comments, btw, if you didn't already notice, most of the miot devices like this are also controllable directly using the `genericmiot` integration, i.e., `miiocli genericmiot --ip <x> --token <x> status` etc.

PR #1929, opened 2024-04-15
zhimi-rma2 support and descriptors for air purifier settings and sensors

DawidPietrykowski:


rytilahti (approved):
LGTM, thanks @DawidPietrykowski! Looks like codecov has some issues, but I'll merge this when I get CI fixed.

PR #1970, opened 2024-09-22
replace deprecated appdirs with platfromdirs fork

a-detiste:
python3-appdirs is dead upstream[1] and its Debian maintainer has indicated
that it should not be included in trixie[2]. A recommended replacement is
python3-platformdirs[3], which is a fork of appdirs with a very similar API.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1067997


rytilahti (approved):
LGTM, thanks @a-detiste! The CI errors seem unrelated (even when caused by some lib upgrades) so I'll just merge this as is.

PR #1979, opened 2024-10-26
Drop python3.8 support

rytilahti:


rytilahti:
The py3.13 tests fail due to a regression (https://github.com/python/cpython/issues/125316) that is going to be fixed in a newer 3.13 release, so ignoring this for now.

PR #1984, opened 2024-11-09
Add unique_identifier property to miot properties, actions, and events

rytilahti:
This allows descriptors to have device-unique identifiers, the format is '<normalized_name>_<siid>_<id>'.

This also changes 'id' of the descriptors to use this identifier in-place of a plain name from the description.


rytilahti:
The CI fails are irrelevant for this change.

PR #1993, opened 2024-12-14
Fix partial in enum for Python 3.13

cdce8p:
Fix Deprecation warning emitted by Python 3.13.1.
```
functools.partial will be a method descriptor in future Python versions;
wrap it in enum.member() if you want to preserve the old behavior
```

Ref: https://github.com/python/cpython/issues/125316

### DIFF
--- a/miio/miot_device.py
+++ b/miio/miot_device.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 from enum import Enum
 from functools import partial
 from typing import Any, Optional, Union
@@ -8,6 +9,9 @@ import click
 from .click_common import EnumType, LiteralParamType, command
 from .device import Device, DeviceStatus  # noqa: F401
 from .exceptions import DeviceException
+
+if sys.version_info >= (3, 11):
+    from enum import member
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


rytilahti (approved):
Thanks @cdce8p, looks rather straight-forward! As this is only affecting the cli tool, I'm ready to merge it even though I haven't been able to test it against a real device.

Before doing that, should the check be against >=3.13 instead of 3.11, though?

cdce8p:
> As this is only affecting the cli tool, I'm ready to merge it even though I haven't been able to test it against a real device.

Not sure about that. It has been a while but AFAIR I first saw the DeprecationWarning in the Home Assistant test suite.

> Before doing that, should the check be against >=3.13 instead of 3.11, though?

Enum.member is available since 3.11 and from what I've read on the cpython issue, it's the "correct" way to do it anyway.

rytilahti:
Fair enough, IIRC, this was just a hack to make the cli nicer to use, so if it doesn't work will need to revisit it (e.g., by converting away from the enum to simply use a dict for mapping).

garry0garry:
Is there any news on the topic "member()"?

garry0garry:
https://github.com/home-assistant/core/issues/132902#issuecomment-3288719698

Issue #1387, opened 2022-04-07
Revert change to AQI value polling from air purifier


mouth4war:
**Describe the bug**
A previous issue  (https://github.com/rytilahti/python-miio/issues/1281) changed the way AQI value is polled from the air purifier by enabling AQI hearbeat to remove manufacturer's smoothening and filters on the hardware-reported values. This enabled real-time raw AQI values to be read at the cost of the filters.

I prefer the filtered AQI data instead as it is much less noisy.

**Version information (please complete the following information):**
 - OS: [e.g.Linux]
 - python-miio: [Use `miiocli --version` or `pip show python-miio`]

**Device information:**
If the issue is specific to a device [Use `miiocli device --ip <ip address> --token <token> info`]:
  - Model: Air Purifier 3H
  - Hardware version:
  - Firmware version:

mak3:
> manufacturer's smoothening and filters

Where did you read that about it?
Nobody noticed any smoothening, all readings were just flat for 30 minutes. 

Only readings above 40 PM2 could previosly trigger proper reporting. 
Maybe in China it's fine, but in Europie it's 250% of allowed norm, so it isn't acceptable for automation to react only after 40 PM2.

If you want smoothened results, you should use moving average (ca. 1-5 minutes) and not introduce fake reporitng.

https://www.home-assistant.io/integrations/filter/ -> time_simple_moving_average

mouth4war:
https://github.com/home-assistant/core/issues/67528

rytilahti:
So I'm not sure what's the best option here so that's why I haven't responded... I  have no device to test, so I'm not sure, for example, what would happen if the value would be adjusted to be `30` or `60`? Would this change the update interval to be that? Should it also be done only once during the initialization? 

In principle I agree with @mak3 that it would make more sense to give out the reading as reported by the device, and let the downstream smoothen (or filter) the reported values as needed..

Ping @Razer0123 who reported the delayed updates in the first place.

PR #2040, opened 2025-08-29
Permit setting aqi_realtime_update_duration and do not override the value of aqi_realtime_update_duration if it is set for the Mi Air Purifier 3/3H (zhimi.airpurifier.mb3)

d1b:
Tested against a 3H.

### DIFF
--- a/miio/integrations/zhimi/airpurifier/tests/test_airpurifier_miot.py
+++ b/miio/integrations/zhimi/airpurifier/tests/test_airpurifier_miot.py
@@ -31,7 +31,9 @@ _INITIAL_STATE = {
     "filter_rfid_product_id": "0:0:41:30",
     "filter_rfid_tag": "10:20:30:40:50:60:7",
     "button_pressed": "power",
+    "aqi_realtime_update_duration": 0,
 }
+
 
 _INITIAL_STATE_MB4 = {
     "power": True,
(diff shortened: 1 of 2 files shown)


miio/integrations/zhimi/airpurifier/tests/test_airpurifier_miot.py
```diff
@@ -31,8 +31,10 @@
     "filter_rfid_product_id": "0:0:41:30",
     "filter_rfid_tag": "10:20:30:40:50:60:7",
     "button_pressed": "power",
+    "aqi_realtime_update_duration": 0,
```
d1b, line 34: The default value for my 3H for this setting is 0. I can move this change to device specific init state if that is desired.

miio/integrations/zhimi/airpurifier/tests/test_airpurifier_miot.py
```diff
@@ -236,6 +241,44 @@ def test_set_anion(self):
...
+        # Set a specific value
+
+        self.device.set_aqi_realtime_update_duration(specific_value)
+
+        # Check that we do not override a specific set value
+        status = self.device.status()
+
+        assert status.aqi_realtime_update_duration == specific_value
+
+    def test_set_aqi_realtime_update_duration_negative_value(self):
```
d1b, line 273: I am not sure what other models have or do not have the ability to set and or provide information on `aqi_realtime_update_duration_negative_value`  are there other changes that I should do in relation to this change and the related tests ?

d1b:
I would like to make it possible to set a custom AQI update interval (re https://github.com/rytilahti/python-miio/issues/1387). While this seems to work - it seems like the device doesn't want to "stick" to a specific update duration after you set it. The value it provides back is often 0 after setting the value to a non-zero value ...

PR #1618, opened 2022-12-05
Add metadata support for genericmiot

rytilahti:
This PR is continuation for #1581 and improves genericmiot by adding support for defining metadata (e.g., description, icon) for miotspec entities.

The idea behind this PR is to have consolidated descriptions, icons, etc. for standard properties, actions, and settings no matter what is defined in the the miotspec file.

At the moment the following metadata can be defined:
* description (overrides the spec file given one)
* icon
* device_class

The structure for metadata files is:
```
<namespace>:
  services:
    <service_name>:
       action:
         <action_name>:
           # metadata
       property:
         <property_name>:
           # metadata
```
Example definition (namespace defined one level higher and thus not shown here):
```
description: Metadata miot-spec-v2 namespace
services:
  battery:
    description: Battery service
    icon: mdi:battery
    action:
      start-charge:
        description: Go charging
    property:
      charging-state:
        description: Charging status
      battery-level:
        description: Battery level
        icon: mdi:battery
```

The screenshot below shows what a simulated `dreame.vacuum.p2009` looks like in homeassistant (compare to the one in #1581).

image

`miiocli status` output showing also items in Chinese that have no metadata defined:
```
Status (vacuum:status): Idle (value: 2) (from: Sweeping (1), Idle (2), Paused (3), Error (4), Go Charging (5), Charging (6), Mopping (7))
Error (vacuum:fault): 33  (min: 0, max: 100, step: 1)
[S] Fan speed (vacuum:mode): Full Speed (value: 3) (from: Silent (0), Basic (1), Strong (2), Full Speed (3))
Battery level (battery:battery-level): 8 % (min: 0, max: 100, step: 1)
Charging status (battery:charging-state): Charging (value: 1) (from: Charging (1), Not Charging (2), Go Charging (5))
Brush time left (brush-cleaner:brush-left-time): 4 days, 9:00:00 (min: 0, max: 200, step: 1)
Brush life level (brush-cleaner:brush-life-level): 85 % (min: 0, max: 100, step: 1)
Filter life level (filter:filter-life-level): 79 % (min: 0, max: 100, step: 1)
Filter time left (filter:filter-left-time): 2 days, 1:00:00 (min: 0, max: 150, step: 1)
Work mode (vacuum-extend:work-mode): 50  (min: 0, max: 50, step: 1)
Cleaned time (vacuum-extend:cleaning-time): 13:02:00 (min: 0, max: 32767, step: 1)
[...]

PR #2062, opened 2026-03-26
Add YAML metadata support for genericmiot

Acrobot:
Implements the metadata approach from #1618, adapted to the current codebase where descriptor creation has moved into the miot models.

Adds a YAML-based metadata system that enriches genericmiot descriptors with human-readable descriptions, replacing the vendor-provided, often-Chinese defaults from miotspec files:

- Metadata is organized into namespace YAML files mirroring the MIoT URN namespace structure
- A fallback chain allows dreame-specific services to fall back to miot-spec-v2, and all namespaces (including unknown ones) to fall back to `common`
- A `__ANY__` wildcard service in `common` provides cross-service property descriptions (temperature, humidity, CO2, PM2.5/PM10, voltage, etc.) without per-service repetition
- Descriptor enrichment is non-mutating: the original descriptor is preserved in `extras['original']` for callers that need the raw device-given name

## Design

- `meta.py`: `Metadata` class loads YAML namespace files and looks up metadata by URN namespace, service, type, and entity name
- `metadata/base.yaml`: index mapping namespace names to YAML files
- `metadata/miotspec.yaml`: definitions for `miot-spec-v2` (battery, vacuum, filter, brush, light, identify services)
- `metadata/dreamespec.yaml`: definitions for `dreame-spec` (vacuum-extend, DnD, audio, clean-logs, time services); falls back to `miot-spec-v2`
- `metadata/common.yaml`: cross-namespace definitions under a `__ANY__` wildcard service (MAC address, voltage, temperature, humidity, CO2, PM2.5/PM10, brightness, cleaning stats); implicit fallback for all namespaces including unknown ones
- `genericmiot.py`: `_enrich_with_metadata()` returns a new descriptor via `attr.evolve` with the metadata name; the original descriptor is preserved in `extras['original']` for callers that need the raw device-given name

## Namespace fallback chain

```
dreame-spec → miot-spec-v2 → common
<any unknown namespace> → common
```

Namespaces declare their fallback with a `fallback:` key. `common` is the implicit fallback for all namespaces. Within each namespace, the `__ANY__` wildcard service matches properties regardless of which service they belong to.

Model-specific namespaces (keyed by `urn.model`) are a natural future extension: add a model YAML file with a `fallback:` pointing to the relevant namespace.


rytilahti (approved):
Thanks a lot for this PR, @Acrobot! Now that I have a real device to test against, I did some rewiring based on how I felt it should behave.

I did quite a few cleanups to keep it simple (and not bound to homeassistant specifics like device_classes and icons), but this might be extended later. I am planning to do some more improvements to stress test the design & to add proper support for my airmonitor (`urn:miot-spec-v2:device:air-monitor:0000A008:cgllc-cgd1st:1`), but I will save the improvements into separate PRs.

Issue #1750, opened 2023-03-01
Bug TimeOut Dreame F9, works intermittently


Cityloue15:
**Describe the bug**
The device responds intermittently, following a disconnection or reconnection, it responds 1 or 2 times, then nothing

**Version information (please complete the following information):**
 - OS: [e.g.Linux] Linux 5.10.103-v8+ Debian GNU/Linux 10 (buster) 64bits (aarch64)
 - python-miio: version 0.6.0.dev0

**Device information:**
  - Model: dreame.vacuum.p2008
  - Hardware version: Linux
  - Firmware version: 3.5.8_1050

![image](https://user-images.githubusercontent.com/73710807/222079129-8ef00115-42a5-484d-ad65-36bdaee0b40b.png)

**To Reproduce**
I would like the robot to respond

sudo miiocli genericmiot --ip 192.168.10.32 --token MYTOKEN info
or
sudo miiocli dreamevacuum --ip 192.168.10.32 --token MYTOKEN info

ERROR:miio.miioprotocol:Got error when receiving: timed out
ERROR:miio.click_common:Exception: No response from the device
...

**Expected behavior**

![image](https://user-images.githubusercontent.com/73710807/222078380-e757a419-108c-47bb-bb95-a46c9047e505.png)

**Console output**
miiocli discover 
```
INFO:miio.miioprotocol:Sending discovery to <broadcast> with timeout of 5s..
INFO:miio.miioprotocol:  IP 192.168.10.32 (ID: 18c2a073) - token: b'ffffffffffffffffffffffffffffffff'
INFO:miio.miioprotocol:  IP 192.168.10.6 (ID: 175d8e5c) - token: b'ffffffffffffffffffffffffffffffff'
INFO:miio.miioprotocol:Discovery done
INFO:miio.discovery:Discovering devices with mDNS for 5 seconds...
INFO:miio.discovery:Found a supported 'DreameVacuum' at 192.168.10.32 - token: ffffffffffffffffffffffffffffffff
```
sudo miiocli --debug genericmiot --ip 192.168.10.32 --token MYTOKEN info
```
DEBUG:miio.click_common:Unknown model, trying autodetection
DEBUG:miio.protocol:Unable to decrypt, returning raw bytes: b''
DEBUG:miio.miioprotocol:Got a response: Container:
    data = Container:
        data = b'' (total 0)
        value = b'' (total 0)
        offset1 = 32
        offset2 = 32
        length = 0
    header = Container:
[...]

PR #2063, opened 2026-03-26
Cache miIO protocol sequence IDs across CLI invocations

### LINKED ISSUE #1751: Modify miiocli to save the "options" object much like mirobo does for sequence ids. This would fix the long-standing problems some users are having caused by the id reuse.

Acrobot:
Devices track message sequence IDs and ignore duplicates. Without persisting the counter, restarting the CLI resets it to 0, causing devices to silently drop messages and time out (#1750).

This generalizes the approach already used by the roborock vacuum CLI (`vacuum_cli.py`) into the shared `DeviceGroup` infrastructure, so all device types benefit automatically.

- `device_cache.py`: Per-device cache using hashed IPs (safe for IPv6) under `platformdirs.user_cache_dir`
- `click_common.py`: `DeviceGroup.group_callback` loads cached `start_id` on init, saves `raw_id` via `ctx.call_on_close`
- 12 tests covering path generation, read/write, corruption handling

Full test suite: 1298 passed (same as master), no new failures. Ruff clean.

Closes #1751

### DIFF
--- a/miio/click_common.py
+++ b/miio/click_common.py
@@ -14,6 +14,7 @@ from typing import Any, ClassVar
 
 import click
 
+from .device_cache import read_cache, write_cache
 from .exceptions import DeviceError
 
 try:
@@ -266,7 +267,20 @@ class DeviceGroup(click.MultiCommand):
         gco = ctx.find_object(GlobalContextObject)
         if gco:
             kwargs["debug"] = gco.debug
+
+        ip = kwargs.get("ip")
+        if ip:
+            cached = read_cache(ip)
+            kwargs.setdefault("start_id", cached["seq"])
+
         ctx.obj = self.device_class(*args, **kwargs)
+
+        if ip:
+
+            def _save_cache() -> None:
+                write_cache(ip, {"seq": ctx.obj.raw_id})
+
+            ctx.call_on_close(_save_cache)
 
     def command_callback(self, miio_command, miio_device, *args, **kwargs):
         return miio_command.call(miio_device, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


miio/tests/test_device_cache.py
```diff
@@ -0,0 +1,83 @@
...
+
+@pytest.fixture()
+def cache_dir(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Path:
+    """Override CACHE_DIR to use a temporary directory."""
+    monkeypatch.setattr("miio.device_cache.CACHE_DIR", tmp_path)
+    return tmp_path
+
+
+class TestCachePath:
+    def test_ipv4_produces_valid_path(self, cache_dir: Path) -> None:
```
rytilahti: I know claude wants to often pack python tests inside classes, but this is IMO not a good thing. Better to keep it flat :-)

miio/click_common.py
```diff
@@ -265,8 +266,21 @@ def group_callback(self, ctx, *args, **kwargs):
...
+
+        ip = kwargs.get("ip")
+        if ip:
+            cached = read_cache(ip)
+            kwargs.setdefault("start_id", cached["seq"])
+
         ctx.obj = self.device_class(*args, **kwargs)
 
+        if ip:
+
```
rytilahti, line 279:
```suggestion
```

miio/click_common.py
```diff
@@ -265,8 +266,21 @@ def group_callback(self, ctx, *args, **kwargs):
         gco = ctx.find_object(GlobalContextObject)
         if gco:
             kwargs["debug"] = gco.debug
+
+        ip = kwargs.get("ip")
+        if ip:
+            cached = read_cache(ip)
+            kwargs.setdefault("start_id", cached["seq"])
```
rytilahti, line 274: Nice solution! :-)

rytilahti (approved):
We could store some more information in the device cache, but this is a great starting point & is immediately useful for any cli users.

PR #2064, opened 2026-03-26
Add configurable handshake timeout

### LINKED ISSUE #1683: Replace lazy discovery in favor of configurable time-since-last-handshake

Acrobot:
The current `lazy_discover` flag only offers two extremes: handshake once on first request (`True`), or before every request (`False`). Devices that need a recent handshake but not on *every* request (referenced in HA issues #59215, #92774) have no middle ground.

This adds a `handshake_timeout` parameter (in seconds) to `MiIOProtocol` and `Device`. It tracks when the last handshake occurred and only re-handshakes when enough time has passed. Setting it to `0` matches `lazy_discover=False`, and leaving it unset preserves existing behavior.

The `lazy_discover` parameter is kept for backward compatibility - it's translated into the timeout internally.

Changes:
- `miioprotocol.py`: `_needs_handshake()` replaces the inline boolean check, `_last_handshake` timestamp recorded on each handshake
- `device.py`: threads `handshake_timeout` kwarg through to `MiIOProtocol`
- 8 new tests covering all timeout/lazy_discover combinations

Full test suite: 1294 passed (same as master), no new failures. Ruff clean.

Addresses #1683
Fixes #1683

### DIFF
--- a/miio/miioprotocol.py
+++ b/miio/miioprotocol.py
@@ -35,7 +35,7 @@ class MiIOProtocol:
         lazy_discover: bool = True,
         timeout: int = 5,
         *,
-        handshake_timeout: Optional[int] = None,
+        handshake_timeout: int | None = None,
     ) -> None:
         """Create a :class:`Device` instance.
 
@@ -60,7 +60,7 @@ class MiIOProtocol:
         self.__id = start_id
 
         if handshake_timeout is not None:
-            self._handshake_timeout: Optional[timedelta] = timedelta(
+            self._handshake_timeout: timedelta | None = timedelta(
                 seconds=handshake_timeout
             )
         elif lazy_discover:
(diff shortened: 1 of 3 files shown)


miio/miioprotocol.py
```diff
@@ -96,6 +114,27 @@ def send_handshake(self, *, retry_count=3) -> Message:
 
         return m
 
+    def _needs_handshake(self) -> bool:
+        """Return True if a handshake is needed before sending.
+
+        The decision is based on _handshake_timeout, which is set in __init__
+        either from the explicit handshake_timeout parameter, or derived from
+        the legacy lazy_discover flag:
```
rytilahti: It is enough to describe the logic briefly, no need to name the exact variable names etc. in this docstring.

Maybe the main docstring for the class could describe _why_ you want to use these variables (lazy_discover was never really properly documented), but it is also good as it is.

rytilahti (approved):
Really nice to clean up the logic behind the handshaking, thanks @Acrobot!

Btw, "addresses #x" does not trigger the issue linking, I modified the description to make this PR to close the issue.

Issue #1156, opened 2021-10-09
Consolidate device class constructors


rytilahti:
Different device implementations take currently different parameters, making it cumbersome to change common settings like timeouts.

The API should be consolidated among all integrations and unit tests should be added to keep the API consistent in the future.

PR #2065, opened 2026-03-27
Make model keyword-only in all Device subclass constructors

Acrobot:
The base `Device.__init__` defines `model` as keyword-only (after `*`), but six subclasses had it as positional: FanP5, Fan1C, Huizuo, AirConditioningCompanion, AirConditioningCompanionMcn02, and Yeelight. This breaks Liskov substitution - code that works with the base class signature can't safely substitute these subclasses.

Fixes all six to use `*, model: Optional[str] = ...` matching the base class, and adds `test_model_is_keyword_only` that inspects every Device subclass signature to prevent regressions.

No positional `model` usage was found anywhere in the codebase, so this is a safe change.

Full test suite: 1345 passed (1286 on master + 59 new parametrized `test_model_is_keyword_only` cases). Same pre-existing failures only. Ruff clean.

Addresses #1156

### DIFF
--- a/miio/integrations/dmaker/fan/fan.py
+++ b/miio/integrations/dmaker/fan/fan.py
@@ -112,7 +112,8 @@ class FanP5(Device):
         debug: int = 0,
         lazy_discover: bool = True,
         timeout: Optional[int] = None,
-        model: str = MODEL_FAN_P5,
+        *,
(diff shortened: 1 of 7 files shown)


rytilahti (approved):
Looks good and straightforward to me, thank you @Acrobot!

PR #2069, opened 2026-03-30
Add Xiaomi Smart Pet Fountain 2 support

JanOstrowka:
## Summary
- add first-class `python-miio` support for Xiaomi Smart Pet Fountain 2 (`xiaomi.pet_waterer.70m2`)
- expose a dedicated MIOT device class, typed status enums, and control methods
- include tests for the main controls and missing-value handling

## Why
Home Assistant core review requested that the pet fountain implementation live in the library used by the integration instead of being carried as a local helper in `xiaomi_miio`.

## Included
- `XiaomiPetFountain` device class
- status model for fountain status, water mode, charging state, DND times, battery, filter metrics, and fault state
- control methods for mode, water interval, child lock, do not disturb, DND start/end, and filter reset
- export from `miio.__init__`
- README supported-device entry

## Related
- home-assistant/core#166815

## Validation
- `python3 -m py_compile` on the new pet fountain package and tests
- the equivalent entity model and control flow were exercised against a real device from the Home Assistant side while iterating on the integration support

I was not able to run the full `pytest` suite in this environment because `pytest` is not installed locally.


JanOstrowka:
I split the Read the Docs cleanup into a dedicated follow-up PR: #2070.

That PR fixes the warning sources causing the RTD build to fail on Python 3.9 / Sphinx 7.4.7, and I verified it locally with:

`/tmp/python-miio-rtd-rtd/bin/sphinx-build -b html -W docs /tmp/python-miio-rtd-rtd74-html`

So the remaining RTD failure on this feature PR should be addressable independently of the pet fountain implementation.

PR #2073, opened 2026-06-26
Drop PyPy from the CI test matrix

syssi:
## Problem

The `pypy3.9` CI job fails because `cryptography>=49` uses `pyo3-ffi v0.29.0` which requires PyPy >= 3.11. There is no version of PyPy that satisfies all dependency constraints simultaneously:

- `cryptography` has no prebuilt wheels for PyPy 3.11+ and source builds require pyo3 >= 0.23
- `pydantic-core` dropped prebuilt PyPy 3.10 wheels in recent versions
- Older `cryptography` versions (43.x) cap out at PyPy 3.10 via pyo3 0.22.2

## Fix

Remove PyPy from the CI matrix. The Rust-based dependency ecosystem (cryptography, pydantic-core) no longer provides a stable PyPy target that works across all supported versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13", "pypy3.9"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:


syssi:
@rytilahti Could you remove the job (`Python pypy3.9 on ubuntu-latest`) from the branch protection rules? Thanks in advance!

rytilahti:
@syssi done! thanks for working on getting things cleaned up! Python 3.9 support should be dropped completely as it is end-of-life already (https://devguide.python.org/versions/), I would say that (at least) 3.10 can also go as its support ends later this year.

We can take a look at the usage data to see which versions are used by the people downloading the current release (looks like it is still being actively downloaded :o), and maybe consider modernizing our target to be even newer than that.

syssi:
Let's drop the end-of-life python versions in a dedicated pull request after the official 0.6.0 release.

rytilahti (approved):

PR #2075, opened 2026-06-27
Pin CI actions to commit hashes

syssi:
## Summary

Pins all GitHub Actions in `.github/workflows/ci.yml` to exact commit hashes instead of mutable version tags to prevent supply-chain attacks. Version numbers are retained as inline comments for readability.

| Action | Before | After |
|---|---|---|
| `actions/checkout` | v4 | v4.3.1 |
| `actions/setup-python` | v5 | v5.6.0 |
| `codecov/codecov-action` | v4 | v5.5.5 |

## Test plan

- [x] CI passes with pinned action hashes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,8 @@ jobs:
         python-version: ["3.12"]
 
     steps:
-      - uses: "actions/checkout@v4"
-      - uses: "actions/setup-python@v5"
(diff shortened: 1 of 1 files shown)


syssi:
Let me know if you don't like commit hashs here.

rytilahti (approved):

rytilahti:
I think it's fine even when the actions are from GitHub. They are going to improve the CI security later this year, so we may need to revise this after that.

